PR #12, opened 2015-06-14
Add a command to list the plugins already installed

kdisneur:
This command could be useful when, for example, you have a setup script and want to know if you need to install the plugin or if the plugin is already installed on the machine.


HashNuke:
Thank you @kdisneur ~!

kdisneur:
Thanks for the merge :)

PR #13, opened 2015-06-15
Skip installation if the version is already installed

kdisneur:


HashNuke:
@kdisneur default exit code is zero. This change isn't required. Just removing the "exit 1" was sufficient.

HashNuke:
Sorry just read the code. It looks like this exit is required. I'll re-read it when I get back to desk again (in bed now)

HashNuke:
@kdisneur Thank you. Merged ~!

kdisneur:
:+1: thanks - with this version my CI build passes and a rebuild passes also :)

PR #20, opened 2015-08-11
Fix path: syntax in .tools-version

MSch:


HashNuke:
@MSch Thank you. I missed this and just noticed.

MSch:
@HashNuke No worries, thanks for merging! I fear though that there's still a few places where `exec` is missing, I sometimes see the shim processes in Activity Monitor.

HashNuke:
@MSch Thanks for looking into it. This needs a few tests so that changes can be made confidently. I'm still not sure how to have tests that cover the plugins too.

PR #22, opened 2015-11-05
Add check to ensure current "file" in loop is actually and executable file

Stratus3D:
This works most of the time, but when the directory does not contain any files `$install_path/$bin_path/*` doesn't expand (e.g. it might look like `"~/.asdf/install/lua/lua-empty-bin-dir/*"` for a Lua install). As a result what is passed to `write_shim_script` is passed a non-existent file and this causes the function to fail.  With this check, we ensure the file exists and is an executable. This ensures we are only creating shims for real executable files.


Stratus3D:
Also removed the TODO comment

PR #23, opened 2015-11-17
Add more details to the 'asdf install' command help and the README

Stratus3D:
I wasn't aware of how the `asdf install` command worked when invoked with no arguments. Hopefully these additions will make it clear that it can install packages listed in the .tool-versions file.


HashNuke:
Thank you @Stratus3D ~! Merged.

PR #24, opened 2015-11-28
Add basic support for legacy version files (e.g. .ruby-version)

Stratus3D:
Add basic support for legacy version files (e.g. .ruby-version) via an optional `get_version_from_legacy_file` script in each plugin.

Questions:
* Is it safe to equate plugin name with tool name? Or could there be variation in the tool name? (I make this assumption in the `get_tool_version_from_legacy_file` function).
* Should `get_version_from_legacy_file` be named `get-version-from-legacy-file`? It seems the other "callback" scripts use dashes. I'll be happy to change this. I only noticed this after I finished.

I have manually tested this PR with the updated asdf-ruby plugin on Ubuntu 14.04 and OSX 10.10.5.

### DIFF
--- a/docs/creating-plugins.md
+++ b/docs/creating-plugins.md
@@ -50,6 +50,10 @@ Setup the env to run the binaries in the package.
 
 Uninstalls a specific version of a tool.
 
+#### bin/get_version_from_legacy_file
+
+Prints the version of the tool to use if a legacy version file is found in the current directory (e.g. rbenv's `.ruby-version`). Current directory is passed as the only argument to this script.
+
 ### Custom shim templates
 
 **PLEASE use this feature only if absolutely required**
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -135,8 +135,8 @@ get_tool_version_from_legacy_file() {
   local plugin_path=$(get_plugin_path $plugin_name)
   check_if_plugin_exists $plugin_path
 
-  if [ -f ${plugin_path}/bin/get_version_from_legacy_file ]; then
-    local legacy_tool_version=$(bash ${plugin_path}/bin/get_version_from_legacy_file $(pwd))
+  if [ -f ${plugin_path}/bin/get-version-from-legacy-file ]; then
+    local legacy_tool_version=$(bash ${plugin_path}/bin/get-version-from-legacy-file $(pwd))
   fi
 
   # Should return the version/tag/commit/branch/path
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -137,7 +137,7 @@ get_tool_version_from_legacy_file() {
   local legacy_version_script="${plugin_path}/bin/get-version-from-legacy-file"
   check_if_plugin_exists $plugin_path
 
-  if [ -f $legacy_tool_version ]; then
+  if [ -f $legacy_version_script ]; then
     local legacy_tool_version=$(bash $legacy_version_script $directory)
   fi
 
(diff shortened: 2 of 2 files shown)


Stratus3D:
This does not add support for legacy version files to the install command. I wasn't sure if that would be a good idea or not.

HashNuke:
@Stratus3D Thank you. I'll check this out and merge once I write tests for this. Need a week to finish up other work. I also have to add a config option to enable/disable this feature.

HashNuke:
@Stratus3D Yes please. dasherized names. Sorry about that I forgot I use dashes everywhere I can avoid underscores.

Yes the assumption is plugin name is tool name. I think if you add the ruby plugin repo as "foobar" plugin, then your .tool-version files will be "foobar 2.1.2" instead of "ruby 2.1.2". Hence the convention for names of repositories. "asdf-foobar" github repo name means the user is assumed to add it as a foobar plugin.

Stratus3D:
@HashNuke I updated the PR so it expects the legacy version script to use dashes. I also renamed the file in the asdf-ruby plugin: https://github.com/HashNuke/asdf-ruby/pull/4

Stratus3D:
Should I add legacy version support to the install command? I could add it in a separate PR if you wanted.

HashNuke:
ah no worries. This amount of detail is enough. We just want to make the
transition/co-operation between other version managers easier

HashNuke:
I promise to finish up my work this week.

TODO:
* [ ] setup tests
* [ ] support config file

Stratus3D:
Please do not merge this. I believe I discovered a bug while using my forked version containing these changes. I will update the PR when I figure out the issue.

HashNuke:
@Stratus3D No worries. Thanks for updating the issue :)

colinrymer:
Any update on this? Anything I can do to help? I'm very interested in using this, but support for `.ruby-version` and `.node-version` files is a prerequisite.

HashNuke:
@colinrymer I'll try to get this in this weekend. @Stratus3D needs to confirm too.

colinrymer:
I think I see the bug @Stratus3D was talking about (and if not, then identified another): the presence of a `.ruby-version`, etc. file should be checked in parent directories if a `.tool-version` file can't be found higher up as well.

Stratus3D:
@colinrymer I believe that's by design. I we decided to only check for the legacy version file in the current directory. If it's not found in the current directory we resort to searching for the `.tool-versions` file in parent directories.

I think the bug I found was due to what happens when a `.ruby-version` file isn't present. It wasn't kicking back over to what's defined in the `.tool-versions` file. Not sure what was causing that though. Hopefully I'll get it figured out this weekend.

HashNuke:
@Stratus3D Oh yes I remember that conversation now. Thanks for clarifying Trevor ~!

Stratus3D:
I just pushed a change that fixes the issue I was experiencing. @HashNuke @colinrymer do you guys want to test it out on your machines? You'll also need asdf-ruby, since that's the only plugin that uses this functionality right now.

Stratus3D:
Made another mistake in the get_tool_version_from_legacy_file function. Just pushed the fix.

HashNuke:
@Stratus3D Thanks Trevor. I'll test it out in a few hours.

Stratus3D:
@HashNuke let me know if there is anything else I can do to help with this PR. If you need tests written let me know.

HashNuke:
@Stratus3D Sorry abt the delay in reply. PR is fine. This needs config file support before it gets merged - https://github.com/asdf-vm/asdf/issues/41

Created an issue and marked it as "help wanted".

tuvistavie:
@Stratus3D config file support just go merged, so I think this should be ready to go too!

I think you would just need to add the following condition when getting the version from the legacy file.

`$(get_asdf_config_value "legacy_version_file") = "yes"`

HashNuke:
@tuvistavie @Stratus3D Resolved conflicts in PR #47 and added the check for config.

Stratus3D:
Thanks @HashNuke !

PR #27, opened 2015-12-23
Fixes #26

### LINKED ISSUE #26: Bins break when called from within a directory whose name has a space
Steps to reproduce:

1. `mkdir My\ Stuff && cd My\ Stuff`
2. Try to run `irb`, `ruby`, `mix`, or `iex`. On my system they all just sit on a blinking cursor.

I can run any of those commands from any path that doesn't have a space, as far as I can tell. Furthermore, running the bin directly seems to work. (i.e. `~/.asdf/installs/ruby/2.2.3/bin/irb`)

Of course, my favorite option would be to avoid naming my directories with spaces, but I don't always have full say in the matter.

*Update I was originally trying this in Bash on Fedora 22. When I try the same thing from OS X, the screen will spam "usage: dirname path".

P.S. I've been quietly using asdf for about six months now, and love it! I don't have much experience with Bash programming, but if the cause of this doesn't seem immediately obvious to you I would be willing to try to track this down.

Jwashton:
This fixes the case where `dirname` fails if `pwd` contains a space.

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -62,7 +62,7 @@ get_asdf_versions_file_path() {
       asdf_tool_versions_path="$search_path/.tool-versions"
       break
     fi
-    search_path=$(dirname $search_path)
+    search_path=$(dirname "$search_path")
   done
 
   echo $asdf_tool_versions_path


HashNuke:
@Jwashton Thank you ~!

PR #31, opened 2016-02-29
Add bash completion

kennyp:


HashNuke:
Thank you again @kennyp ~!
Merging right away.

PR #40, opened 2016-04-07
Updated README.md to include new plugins

smashedtoatoms:
I wrote some plugins for openresty, postgres, redis, and riak and added them to the list.


HashNuke:
Holy butterflies and dosa batter ~!
@smashedtoatoms Thank you. That's an incredible is list of plugins.

PR #42, opened 2016-04-09
Add Fish Shell support

RobLoach:
http://fishshell.com


HashNuke:
@RobLoach Thank you very much for the PR.
Just saw that @tuvistavie's PR is more detailed. I'm in favour of merging that. Thank you very much for taking the time to send a PR.

PR #44, opened 2016-04-18
Add fish support.

tuvistavie:
I added support for [fish shell](http://fishshell.com/), with nice completions :tada: 

out


tuvistavie:
I just realized that #42 already existed, sorry for that :sweat: 
However this PR also adds a good set of completions, so it is a superset of #42.

PR #45, opened 2016-04-24
Add support for config file.

### LINKED ISSUE #41: .asdfrc config file
The expected path to the config file is `$HOME/.asdfrc`. It needs to be in the following format:

```
key1=value1
key2=value2
```

The only config option that is required to be supported right now is `legacy_version_file`. The possible values for this option should be `yes` or `no`. The default value should be `no`.

The config file is optional. If the file is not present, then all options will use the default values.

tuvistavie:
This closes #41 

The file should be located at `$HOME/.asdfrc` but the path can be overriden using the `ASDF_CONFIG_FILE` environment variable. If it does not exists, the defaults are used.
The default config is expected to be at `asdf_dir/defaults`, does that seem ok?

I am quite lenient on the config file format, but I think it should be ok for this use case.

I also added a few tests and a `.travis.yml` to go with, so it would be nice if you could enable travis :smiley:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: c
+script: bats test
+before_script:
+  - git clone https://github.com/sstephenson/bats.git /tmp/bats
+  - export PATH=/tmp/bats/bin:$PATH
+os:
+  - linux
+  - osx
--- a/defaults
+++ b/defaults
@@ -1,0 +1,2 @@
+# enables the use of .ruby-version like files used by other version managers
+legacy_version_file = no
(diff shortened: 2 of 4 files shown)


tuvistavie:
Sorry, I just realized I indented everything with 4 spaces while all the project was indented with 2, I'll fix that! :sweat:

tuvistavie:
Oh, and thanks for Travis :smile:

HashNuke:
@tuvistavie yeah. that was a long time todo. Thanks for writing those tests.

PR #48, opened 2016-04-27
Added D(DMD) and Scala plugin

sylph01:
I made plugins for the following languages and added to the list:
- D Programming Language : https://dlang.org
  - To be specific, the reference compiler, DMD
- Scala : http://www.scala-lang.org

### DIFF
--- a/docs/creating-plugins.md
+++ b/docs/creating-plugins.md
@@ -23,6 +23,10 @@ Must print a string with a space-seperated list of versions. Example output woul
 1.0.1 1.0.2 1.3.0 1.4
 ```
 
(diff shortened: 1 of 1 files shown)


tuvistavie:
@sylph01 

Hi,sorry for the delay!
We just added a command to test plugins, could you run the tests (preferably on Travis, or something that can run tests on both Linux and OSX) so that we make sure everything is OK before adding them to the doc?
Take a look at the docs for more info.
https://github.com/asdf-vm/asdf/blob/master/docs/creating-plugins.md#testing-plugins

Thank you :smile:

smashedtoatoms:
I updated these in https://github.com/asdf-vm/asdf/pull/116

sylph01:
Sorry, I closed this PR in order to sync my fork with the main repository (When I did so it automatically closed).
I will make another pull request with the D plugin only (since there is another PR with a Scala plugin, which seems to cover the problems I had with my plugin).

PR #50, opened 2016-05-02
Fix local command to support legacy files.

tuvistavie:
I'm not very happy about the implementation, but at least it will make `local` command consistent
with the execution behavior.
If you have any idea to make this cleaner I am all ears.


HashNuke:
@tuvistavie this is good enough bash scripting. Thank you ~!

Issue #37, opened 2016-03-24
Plugin: Python


RobLoach:
https://www.python.org/

tuvistavie:
I just created a Python plugin.
https://github.com/tuvistavie/asdf-python
It only handles cpython, not other flavors.

The handling of the `python` executable is a little hacky, any idea if there is a better way to do this?
https://github.com/tuvistavie/asdf-python/blob/master/shims/python
The issue here is that Python3 does not have a `python` executable and so `python` gives no such command, which I do not think is desirable.

Also, for Python it is a little bit of an issue not to be allowed to set multiple versions of a language.
For example, pyenv allows to set

```
3.5.1
2.7.11
```

if the command exists in 3.5.11, it will use it, otherwise fallback to 2.7.11.

There are a lot of scripts in Python which uses `#!/usr/bin/env python2` or `#!/usr/bin/env python3`, and these would break if the currently set version does not match.

@HashNuke Do you think we could modify `asdf` to be able to write something such as

```
python 3.5.1 2.7.11
```

in `.tool-versions`, to use the second version as a fallback for the command?

HashNuke:
@tuvistavie Ah. That's a pretty huge edge case. If that's what it takes to support python, yes.

tuvistavie:
@HashNuke Yeah, Python is really problematic for these things. I'm going to give it a try, I want to stop using pyenv :smile:

HashNuke:
@tuvistavie anything I can do?

tuvistavie:
@HashNuke It should simply be a matter of changing `asdf-exec` a little, so it should not take too long, thank you :smiley:

tuvistavie:
Python plugin should work fine, closing :smile:

PR #51, opened 2016-05-03
Make it possible to use fallback versions.

tuvistavie:
I added the possibility to add fallback versions for a plugin.
As discussed in #37, this is a must for Python as (sadly) there are a lot
of scripts which work only with a single version of python, and we therefore
need to be able to call the python2 or python3 command.
A typical usage would be

```
asdf global python 3.5.1 2.7.11
pip2 install ansible
ansible # this uses python 2, but we can use it even though 3.5 is the default
python --version # 3.5.1
```

### DIFF
--- a/bin/private/asdf-exec
+++ b/bin/private/asdf-exec
@@ -15,50 +15,53 @@ if [ "$full_version" == "" ]; then
   exit -1
 fi
 
+IFS=' ' read -a versions <<< "$full_version"
 
-IFS=':' read -a version_info <<< "$full_version"
-if [ "${version_info[0]}" = "ref" ]; then
-  install_type="${version_info[0]}"
-  version="${version_info[1]}"
-  install_path=$(get_install_path $plugin_name $install_type $version)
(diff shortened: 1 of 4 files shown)


HashNuke:
@tuvistavie Just remembered abt this PR. We can go ahead with this if there's no other solution in sight for Python.

tuvistavie:
@HashNuke Yes, unfortunately I think it's the only way to have Python working correctly with asdf.

HashNuke:
@tuvistavie merged. Let's roll with this then.

HashNuke:
Someday we should support "system" version as a way to use OS-installed stuff. Very useful for Python and other stuff.

tuvistavie:
Agreed! I'm not exactly sure what would be the best way to handle this though,
what do you think we could do to know which executable to add in shims?

HashNuke:
In general when "system" we could execute the actual system path of the program OR just get out of the way (not sure abt how to do this; just an idea).

PR #53, opened 2016-05-10
Fix #52: Send any `cd` output to /dev/null

### LINKED ISSUE #52: `asdf.sh` can misconfigure `$PATH` on `zsh`
`zsh` provides a configurable hook that runs whenever the current directory changes.  If the hook command produces output, that output is captured [in the `asdf_dir` variable in `asdf.sh`](https://github.com/asdf-vm/asdf/blob/7476c02f0a42851843cc51b3ef9d0fb08cfcabb8/asdf.sh#L9).  This results in a misconfigured `$PATH` variable, which prevents lookup of the `asdf` binary even after running `asdf.sh`.

ctreatma:
Users of `zsh` can define hook functions that execute when
the current directory changes.  If a user has their `zsh`
configured with a hook function that writes to stdout, the
`asdf` setup script will capture that output in addition to
the output of `pwd`, which causes a misconfiguration of the
`PATH` environment variable.

By redirecting the output, if any, of the `cd` command to
`/dev/null`, this ensures that the path to `bin` in `asdf`
is captured correctly regardless of `zsh` configuration.

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -6,7 +6,7 @@ else
   current_script_path=$0
 fi
 
-asdf_dir=$(cd $(dirname $current_script_path) > /dev/null; echo $(pwd))
+asdf_dir=$(cd $(dirname $current_script_path) &> /dev/null; echo $(pwd))
 export PATH="${asdf_dir}/bin:${asdf_dir}/shims:$PATH"
 
 if [ -n "$ZSH_VERSION" ]; then
--- a/asdf.sh
+++ b/asdf.sh
@@ -6,7 +6,7 @@ else
   current_script_path=$0
 fi
 
-asdf_dir=$(cd $(dirname $current_script_path); echo $(pwd))
+asdf_dir=$(cd $(dirname $current_script_path) > /dev/null; echo $(pwd))
 export PATH="${asdf_dir}/bin:${asdf_dir}/shims:$PATH"
 
 if [ -n "$ZSH_VERSION" ]; then


HashNuke:
@ctreatma We don't want to add hooks to any shell. Can you add some steps to reproduce the $PATH bug and we'll try to fix that instead?

ctreatma:
@HashNuke I'm not proposing to add any hooks to `asdf`.  The `chpwd` hook is a built-in [`zsh` hook function](http://zsh.sourceforge.net/Doc/Release/Functions.html#Hook-Functions).

ctreatma:
@HashNuke I'm not proposing to add any shell-specific hooks to `asdf`.  The `chpwd` hook is a built-in [`zsh` hook function](http://zsh.sourceforge.net/Doc/Release/Functions.html#Hook-Functions).

The below is not necessarily a canonical example, but here is an example of defining `chpwd` to execute `ls` whenever the current directory changes:

    $  /usr/bin/env zsh
    $  function chpwd() {
          ls
        }
    $ cd ~
    
If you execute those commands, you will see the output of `ls` after running `cd ~`.  A user with an equivalent `zsh` configuration would end up with a misconfigured path upon executing `asdf.sh`, because this line

    asdf_dir=$(cd $(dirname $current_script_path); echo $(pwd))

will capture the output of `cd` as well as the output of `echo $(pwd)` in the `asdf_dir` variable, which is then included in the `$PATH`.

To avoid this, I am proposing to redirect the output of `cd $(dirname $current_script_path)` so that it is not captured in the `asdf_dir` variable.

HashNuke:
@tuvistavie this seems necessary? But since you've used zsh I'll leave it to you to decide.

@ctreatma thank you very much for clarifying ~!

Stratus3D:
@HashNuke I regularly use asdf with zsh, and I haven't experience this issue, but then again I don't use `chpwd`. I'm fine with merging it though.

tuvistavie:
While I know some users users have hooks or overriden version of `cd`s,
I never saw any hook writing to stdout.

@ctreatma Would you have any real-world sample hook that writes something to stdout so I can understand in what situations this is used?

I'm also fine merging this, but in this case maybe we could also redirect stderr to stdout?

ctreatma:
I don't use `zsh`, so I'll pull in my colleague who ran into this issue.

@nhashmi Could you link to your `cd` hook that you have configured for `zsh` and also mention where the hook came from if you didn't write it yourself?

nhashmi:
Here's the `chpwd` hook I have in my `.zsh/configs/post/`:
```sh
# Show the entries in a directory whenever you cd in
function chpwd {
  ls
}
```
This hook came from [@thoughbot/dotfiles](https://github.com/thoughtbot/dotfiles#zsh-configurations).

tuvistavie:
I see, thank you!.
Ok, let's merge this then!

PR #54, opened 2016-05-13
Add command and instructions to test plugin.

tuvistavie:
@HashNuke I believe this will help keeping good quality plugins.
It helped me find out my Python plugin did not work on OSX :sweat: 
I would also like to merge #48 so it will help to check if everything works.


HashNuke:
This was a long-time TODO. Merging. Thank you ~!

Issue #33, opened 2016-03-10
Order of `list-all`


NobbZ:
While setting up my environment, I realised that the `erlang` and `elixir` plugin do list the available versions from newest to oldest, while `nodejs` does it from oldest to newest.

There should be either a "filter" in asdf which does sorts all the version into a specific order, or a requirement to the plugins to list the versions in a particular order.

I'd prefer to see the most recent version as the last item, such that one can see the most recent version of something even in unscrollable environments.

Stratus3D:
Right now I believe list-all calls the `list-all` script in the plugin's bin directory. I suppose asdf could capture that output and sort it, but I think this would be likely to cause other problems as some plugins may need to handle odd cases. For example, the Erlang versioning scheme changed several versions back, from the old R16B02 format to the new semantic versioning (18.2.1). `sort` would not be able to properly sort all the Erlang versions. I think it's best to leave sorting up to the plugins, since they can handle the odd scenarios in the version numbers. 

We can't really enforce a "requirement" since anyone can create plugins and publish them on GitHub for others to use. It would probably be a good idea to have a convention for version ordering. @HashNuke what do you think? Perhaps we should come up with a convention for version sorting and mention it on the plugin creation page?

NobbZ:
I am aware of the fact that sorting using `sort` might be hard, especially because of Erlang version numbering scheme ;)

And by saying "require plugin maintainers to sort in a certain order" I did mean a soft requirement, since I do know it might be hard to enforce it.

HashNuke:
@NobbZ We'll change the nodejs plugin to list the latest first. The idea is to be able to get "see" the latest version quickly.

NobbZ:
That's why I do prefer latest last, I don't have to search the start of the
output, but can read the line right above the prompt. Also when I know the
list will be long, I have to use head to avoid scrolling, while I can just
let it go when latest is last.

Akash Manohar <notifications@github.com> schrieb am Sa., 12. März 2016
16:53:

HashNuke:
@NobbZ Ah yes makes sense. Will make changes.

HashNuke:
Changes made ~!

HashNuke:
@Stratus3D yes. We should mention the version ordering on the plugin creation page as oldest-first.

pvinis:
I guess related to this, is that `nodejs` gives me this:
```
nodejs
  10.15.1
  11.15.0
  12.17.0
  14.13.1
  15.3.0
  8.17.0
[...]

PR #58, opened 2016-06-12
Install the latest version when running tests

tuvistavie:
I had issues installing Python latest version, so I think running the CI on the latest
version will help.
I will eventually ask Travis to enable cron jobs for each language repositories.

https://docs.travis-ci.com/user/cron-jobs/

### DIFF
--- a/lib/commands/plugin-test.sh
+++ b/lib/commands/plugin-test.sh
@@ -33,7 +33,8 @@ plugin_test_command() {
         fail_test "list-all did not return any version"
     fi
 
-    (asdf install $plugin_name ${versions[0]})
+    latest_version=${versions[${#versions[@]} - 1]}
+    (asdf install $plugin_name $latest_version)
 
     if [ $? -ne 0 ]; then
         fail_test "install exited with an error"


tuvistavie:
Oh, after checking Elixir, the latest version is the first in the versions array.
I think the order should be consistent for all plugins, so maybe we could sort the array directly
in asdf rather than in the plugin, what do you think?

HashNuke:
@tuvistavie 👍
Yes we could move the sorting to asdf itself instead of doing it in the plugins. Will bring some uniformity.

tuvistavie:
Ok, so let's close this PR and sort the versions from the latest to the oldest! :+1:

Stratus3D:
@tuvistavie The problem is this doesn't handle language-specific quirks or semantic versioning properly. Here is an example (this code isn't the same as what's in the PR, but it's using `sort`):

    array=("1.20.1" "1.5.1" "1.1.1")
    sorted=$(sort <<<$array)
    # Output
    1.1.1
    1.20.1
    1.5.1

Clearly these versions are not sorted properly. Since this is semantic versioning 1.20.1 is the most recent version, but it's in the middle of the list, between 1.1.1 and 1.5.1, which are both older.

There are other versioning quirks, like versioning method changes (Erlang was using it's own versioning system until version 17, when they switched to semantic versioning). I think having plugins sort there own versions is the safest method, and it allows us to handle inconsistencies when we need to.

Stratus3D:
Also, this was discussed here: https://github.com/asdf-vm/asdf/issues/33

HashNuke:
@Stratus3D OTP-like versions is one thing that bothers me. Since those versions are now past, any other software not following semantic versioning?

I'm now split between choices.

B/w I just found out that `sort -V` will do semantic version sorting. But the `-V` option isn't available on the OSX version of sort.

tuvistavie:
@Stratus3D @HashNuke Thanks for the feedback.

I agree it is maybe better to handle this in each plugin.
For my Python plugin, I thing I am going to use

```
    local sort_cmd
    if sort --help | grep -q -- '-V'; then
        sort_cmd='sort -r -V'
    else
        sort_cmd='sort -r'
    fi
```

for now.
Maybe we can implement something to sort semantic versions later.

Stratus3D:
@tuvistavie I'll update my Lua plugin as well.

PR #59, opened 2016-06-12
Add travis badges to README

tuvistavie:
I added the Travis badge for each plugin, I'll probably send a PR to plugins which are not using Travis yet.
You can check how it looks like [here](https://github.com/tuvistavie/asdf/tree/add-travis-badges)
I also added my Python plugin to the list.


HashNuke:
@tuvistavie lovely ~!

PR #60, opened 2016-06-12
Sort versions in list-all

tuvistavie:
As discussed in #58, I sorted the versions when using list-all from the latest to the oldest.


tuvistavie:
Closing as discussed in #58.

PR #61, opened 2016-06-16
Improve plugin tests

tuvistavie:
My Python plugin [should never have passed the tests](https://github.com/tuvistavie/asdf-python/pull/1/files) but it did, so I improved the `plugin-test` command a little.

### DIFF
--- a/lib/commands/plugin-test.sh
+++ b/lib/commands/plugin-test.sh
@@ -6,9 +6,11 @@ fail_test() {
 
 plugin_test_command() {
     export ASDF_DIR=$(mktemp -dt asdf.XXXX)
+    git clone https://github.com/asdf-vm/asdf.git $ASDF_DIR
 
     local plugin_name=$1
     local plugin_url=$2
+    local plugin_command="${@:3}"
 
     if [ -z "$plugin_name" -o -z "$plugin_url" ]; then
         fail_test "please provide a plugin name and url"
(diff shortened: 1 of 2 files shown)


Stratus3D:
Should the most recent version be in the first line of output or the last? I'm fine with either, but some people might like the most recent version last. It would be closer to the prompt and more likely to be visible.

tuvistavie:
@Stratus3D I have no strong preference, but I also think the most recent being the last seems better in a CLI.
@HashNuke wdyt?

lib/commands/plugin-test.sh
```diff
@@ -33,11 +35,32 @@ plugin_test_command() {
         fail_test "list-all did not return any version"
     fi
 
-    (asdf install $plugin_name ${versions[0]})
+    latest_version=${versions[0]}
```
cmckni3: This won't be the latest version but the first version.

I think [line 28](https://github.com/asdf-vm/asdf/pull/61/files#diff-84d75d3e627bec07409965f77e402042R28) should read the following instead:

```sh
read -a versions <<< $(asdf list-all $plugin_name | tail -1)
```

cmckni3:
I think you should be able to pass in the version to test.

HashNuke:
Most recent first is fine. I used to think most-recent last just because it's closer to the next prompt. But later on as versions grow I think we'll have to limit the list we display (and then give out an option to list everything). Something like that.

HashNuke:
@tuvistavie LGTM 🖖

HashNuke:
@tuvistavie @Stratus3D actually, I prefer whatever is better for long outputs (most-recent last ofcourse). Showing most recent first for Ruby and then I have to scroll all the way back.

B/w I have no strong opinions abt version listing. I swing back and forth on this. So majority wins? 🙃

`/me facepalms at my previous comment`

lib/commands/plugin-test.sh
tuvistavie: Yep, thanks, we were discussing  about the order and I think we are now settled :smile:

tuvistavie:
@cmckni3 install should work properly with any version. I am planning on asking travis to use their cron feature, so that we can detect failures for new versions early enough, so I would rather use the latest all the time.

tuvistavie:
I updated to use the last version (which should be the most recent) I will update elixir and ruby plugins to stick with this order.
Thank you all for the feedback!
/cc @Stratus3D @HashNuke @cmckni3

Issue #68, opened 2016-06-21
command not found: complete


bigardone:
Hi there!
I've just installed `asdf` for Mac, and every time I start a new zsh it prints the following error:

```
/Users/user/.asdf/completions/asdf.bash:43: command not found: complete
```

What can be going wrong?
Thanks in advance!

jswny:
I'm getting the same error using ZSH on CentOS 7.

dbernheisel:
This pull request on a different project seems to have a solution:
https://github.com/eddiezane/lunchy/pull/59

HashNuke:
Can anyone using zsh confirm which solution is better?

\\cc: @asdf-vm/maintainers

tuvistavie:
It's been a while since I haven't used zsh, but #69 should be good!

dbernheisel:
#69 will work for zsh. When this merges, the readme should also be updated to say that zsh users don't need to source `./completions/asdf.bash` into their profiles since it's auto-loaded via `./asdf.sh`

tuvistavie:
@dbernheisel I think there are two choices:

* Autoload `completions/_asdf`, and avoid sourcing `completions/asdf.bash`, which is what we were doing before
* Forget about `_asdf` and use zsh bash compatibility to load `completions/asdf.bash`

I just merged #69, but I believe we still need to source the file as only `_asdf` would get autoloaded,
and I would rather avoid maintaining bash and zsh completion separately.
The bash completion currently being superior to the zsh one, I don't think it is worth trying to replicate the functionality under `_asdf`.

So, if you think it is ok (not being a zsh user myself, I want to be sure this fits everyone need),
I think that we can safely remove `_asdf` from `completions` and avoid appending the `fpath` as
we are not using autoloading here.

dbernheisel:
Great direction. I like the 2nd option as well. A quick test on my system proves it should work without issue.

tuvistavie:
This should now be fixed :smile:

dbernheisel:
If you're here because of Google, this thread is super old now and the docs have updated. As of 2020-05-05, do this:

```sh
# append completions to fpath 
fpath=(${ASDF_DIR}/completions $fpath) 

# initialise completions with ZSH's compinit 
autoload -Uz compinit && compinit
```

bernardo-martinez:
thanks @dbernheisel , but where should we put that?

jthegedus:
[...]

PR #25, opened 2015-12-05
Adds (very) basic zsh autocompletion

justin-calleja:
Only top level commands are supported i.e. `asdf <TAB>`

and you get a list of asdf commands. That's it.

PR #69, opened 2016-06-27
Use bashcompinit instead of compinit for ZSH

tejanium:
Related to https://github.com/asdf-vm/asdf/issues/68, an improvement for https://github.com/asdf-vm/asdf/pull/25


tuvistavie:
Thank you! :smile:

PR #75, opened 2016-07-01
Set ASDF_CONCURRENCY when installing

### LINKED ISSUE #57: Compiling in parallel
Many of the plugins compile some files through make, that's why I'm opening the issue here, in the parent repository.

Should asdf detect if it can compile in parallel and set proper `-jn` flags for make? Is this a job for asdf, or should it assume user has proper `MAKEFLAGS` env variable set?
This can potentially significantly speed up compilation.

tuvistavie:
This set and exports the `ASDF_CONCURRENCY` variable to the install script.
Each plugin should use this variable when appropriate.
I just updated my python plugin to give it a try:

https://github.com/tuvistavie/asdf-python/commit/1538892959436dd32928edba4ed5758e873fd9a2  

This closes #57

### DIFF
--- a/lib/commands/install.sh
+++ b/lib/commands/install.sh
@@ -9,6 +9,17 @@ install_command() {
   fi
 }
 
+get_concurrency() {
+  if which nproc > /dev/null 2>&1; then
+    echo $(nproc)
+  elif which nproc > /dev/null 2>&1; then
+    echo $(sysctl -n hw.ncpu)
+  elif [ -f /proc/cpuinfo ]; then
+    echo $(grep -c processor /proc/cpuinfo)
+  else
+    echo "1"
+  fi
+}
 
 install_local_tool_versions() {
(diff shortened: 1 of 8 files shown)


HashNuke:
❤️🖖

lib/commands/install.sh
```diff
@@ -9,6 +9,17 @@ install_command() {
   fi
 }
 
+get_concurrency() {
+  if which nproc > /dev/null 2>&1; then
+    echo $(nproc)
+  elif which nproc > /dev/null 2>&1; then
```
michalmuskala: Should this be the same test twice?

tuvistavie: Thanks, that should be `which sysctl`!!

tuvistavie: I just pushed a fix to master, thank you! :smile:

Issue #72, opened 2016-06-30
[Proposal] Command to get direct path of the binary


HashNuke:
We need a way to expose the path to a binary. For example to debug with `gdb`, we need the direct path of the `ruby` binary and not the shim.

Something like

```
$ asdf bin ruby
/Users/HashNuke/.asdf/installs/ruby/2.3.0/bin/ruby
```
#### As a tangential discussion

Do we rename `asdf which` command to `asdf version` OR `asdf current` instead?

That way this can be in like with the unix `which` command. So we can use `asdf which ruby`

```
# this is currently what the asdf-which command does
# proposing to change it to asdf-current instead
$ asdf current ruby
2.3.0

# and use asdf-which to return the full path of the binary
$ asdf which ruby
/Users/HashNuke/.asdf/installs/ruby/2.3.0/bin/ruby
```

rockwood:
I just switched asdf, and the `which` command through me off at first. I think the most idiomatic api would have `which` mimic the unix command and have a separate command for listing the current version. Just my two cents.

HashNuke:
@rockwood Same opinion. Will make the change.

tuvistavie:
I also agree on renaming the current `which` to `version` and making `which` behave like its UNIX counterpart.

rockwood:
I wonder if it's also worth having the new command print where the current version is being read from. [Rbenv does that](https://github.com/rbenv/rbenv#rbenv-version) and proved helpful on a few occasions. 

Something like:

```
> asdf current elixir
1.3.1 (set by /Users/rockwood/.tool-versions)
```

tuvistavie:
I agree this would also be a good idea.
I personally prefer `asdf version elixir` to  `asdf current elixir`,
what do you guys think?

rockwood:
My only concern with `asdf version ...` is the potential confusion with `asdf --version`. Although I don't feel that strongly either way.

PR #79, opened 2016-07-12
Rename which to current

rockwood:
Renames `asdf which` to `asdf current` and adds `(set by TOOL_VERSIONS_PATH)`

This is the first step of issue #72. 

Example:

    > asdf current erlang
    > 17.3 (set by /Users/kim/.tool-versions)

I'm by no means a qualified shell scripter, so I would appreciate some feedback on this.

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -67,6 +67,21 @@ display_error() {
   echo >&2 $1
 }
 
+get_version_file_path_for() {
+  local tool_name=$1
+  local legacy_version_file_support=$(get_asdf_config_value "legacy_version_file")
+
+  if [ ! -f "$(pwd)/.tool-versions" ] && [ "$legacy_version_file_support" = "yes" ]; then
+    # Check for legacy version and return "" if it exists
+    local legacy_version=$(get_tool_version_from_legacy_file $tool_name $(pwd))
+    if [ "$legacy_version" != "" ]; then
+      echo ""
+      return 1
+    fi
+  fi
+
+  echo $(get_asdf_versions_file_path)
+}
 
 get_asdf_versions_file_path() {
   local asdf_tool_versions_path=""
--- a/README.md
+++ b/README.md
@@ -105,8 +105,8 @@ asdf plugin-update <name>
 asdf install <name> <version>
 # asdf install erlang 17.3
 
-asdf which <name>
-# asdf which erlang
+asdf current <name>
+# asdf current erlang
 # 17.3
 
 asdf uninstall <name> <version>
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ asdf install <name> <version>
 
 asdf current <name>
 # asdf current erlang
-# 17.3
+# 17.3 (set by /Users/kim/.tool-versions)
 
 asdf uninstall <name> <version>
 # asdf uninstall erlang 17.3
--- a/bin/asdf
+++ b/bin/asdf
@@ -5,7 +5,7 @@ source $(dirname $(dirname $0))/lib/utils.sh
 source $(dirname $(dirname $0))/lib/commands/help.sh
 source $(dirname $(dirname $0))/lib/commands/install.sh
 source $(dirname $(dirname $0))/lib/commands/uninstall.sh
-source $(dirname $(dirname $0))/lib/commands/which.sh
+source $(dirname $(dirname $0))/lib/commands/current.sh
 source $(dirname $(dirname $0))/lib/commands/where.sh
 source $(dirname $(dirname $0))/lib/commands/version_commands.sh
 source $(dirname $(dirname $0))/lib/commands/list.sh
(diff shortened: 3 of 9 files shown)


tuvistavie:
This looks great, thank you very much! :smile: 

@HashNuke @Stratus3D If everyone is ok, I think we can merge this!

rockwood:
I just realized that the "set by" feature bipasses support for legacy tool files. Is that something we should support here?

tuvistavie:
Oh, yes, I think we should support it or it may be surprising to the user.

rockwood:
@tuvistavie It turns out it's a little bit more difficult that I anticipated.

Plugins that support legacy files only return the version number for a given directory, not the path to the tool file it found. Here are some possible options I see in order of easy to hard:

## Option 0

Revert it

This is obviously easy to do, but we'd loose the feature.

## Option 1

Hide the `set by FILE` message if the user has `legacy_version_file = yes`. 

This would pretty easy to implement, but users would loose the feature the even if they're only using `.tool-versions`.

## Option 2

Hide the `set by FILE` message if the version came from a legacy file. 

This would require some internal refactoring of how versions are determined, but would show the message for any version derived from .tool-versions.

## Option 3

Add an additional plugin hook like `get_legacy_file_path` that would take a directory and return a path like `/Users/kim/my_project/.ruby-version`. We could hide the `Set by FILE` message if the plugin didn't implement it. 

## Option 4

Change the existing plugin hooks.

We could have plugins register the filename(s) that they support, and asdf could search for those and `.tool-versions` simultaneously. It might make things more efficient, but would add breaking changes to the plugin api.

HashNuke:
Option #2, #3 and #4 seem fine.

Stratus3D:
For now option 2 seems like the best route.

Option 4 seems interesting though. We could have asdf find the right version file and then pass the filename to the plugin's get-version-from-legacy-file to parse the version.

rockwood:
Right, I think option 2 is the best thing to do for the time being. 

I'd like to get the rest of #72 finished and then revisit the idea of making plugins more dynamic.

rockwood:
Ok, Option 2 is now the behavior. If the version was derived from a legacy file, then only the version will be output by `asdf current`. If it was derived from .tool_versions, then it outputs the "set by TOOL_VERSIONS_PATH"

I think there are some improvements we can make to the legacy file api to make things easier and more efficient when determining versions. I'll put together a proposal issue with my ideas.

lib/utils.sh
```diff
@@ -67,6 +67,21 @@ display_error() {
...
+get_version_file_path_for() {
+  local tool_name=$1
+  local legacy_version_file_support=$(get_asdf_config_value "legacy_version_file")
+
+  if [ ! -f "$(pwd)/.tool-versions" ] && [ "$legacy_version_file_support" = "yes" ]; then
+    # Check for legacy version and return "" if it exists
+    local legacy_version=$(get_tool_version_from_legacy_file $tool_name $(pwd))
+    if [ "$legacy_version" != "" ]; then
+      echo ""
+      return 0
```
tuvistavie: Not very important, but I would rather return `1`, to make it clear the result should not be used.

tuvistavie:
Thanks, this look great! I also agree with option 2 for now, but we should definitely
think about how we can improve the legacy file API.
We now have three ways of checking the version `local`, `global` and `current`,
we should also think about simplifying this, but let's have this discussion somewhere else! :smile:

tuvistavie:
I opened another issue to continue the discussion about getters!
I think we can merge this, let's just wait for another :+1: 
@rockwood Thank you very much :smile:

rockwood:
@tuvistavie Thanks for taking a look. I've got another PR I'd like to open, but it's dependent on this work. Any way we could merge this first?

tuvistavie:
@rockwood Sorry for the delay and thank you for the great work!

Stratus3D:
@tuvistavie should we bump the version since these changes remove support for the `which` command? I'm thinking there are probably scripts out there that depend on `current`.

tuvistavie:
@Stratus3D I agree we should bump the version, but maybe we could wait for https://github.com/asdf-vm/asdf/pull/83 to be merged as it is also a breaking change. WDYT?

Stratus3D:
@tuvistavie Good idea. I'll wait to tag.

Stratus3D:
What if I retroactively tag the last commit before #79 and #83 as `v0.1.0`?

HashNuke:
@Stratus3D yes. Makes sense to have a tag before breaking changes. So we can ask people who still need the old version to use the first tag.

Stratus3D:
@HashNuke great! I'll push the tag.

Stratus3D:
It's tagged :) The next tag should be v0.2.0, since it will include these PRs that change the API. I'll need to update the README with the correct clone command. Should have a PR out this evening.

tuvistavie:
Great, thanks!!

PR #83, opened 2016-07-25
Unify Version Getter

rockwood:
This makes `asdf local` and `asdf global` setters only. `asdf current` will be the command to get the current version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,18 +132,10 @@ asdf list-all <name>
 #### View current version
 
 ```bash
-asdf local [name]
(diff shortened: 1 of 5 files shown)


tuvistavie:
This looks much better, thanks a lot @rockwood !
@Stratus3D @HashNuke It makes quite a lot of breaking changes, so I think we should tag a release!

HashNuke:
Thank you @rockwood & @tuvistavie ~!

PR #85, opened 2016-07-29
Setup command in README should use most recent tag

Stratus3D:
I also went ahead and incremented the version number to 0.2.0. Of course we can wait to tag if there other things that need to make it into 0.2.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Supported languages include Ruby, Node.js, Elixir and more. Supporting a new lan
 Copy-paste the following into command line:
 
 ```bash
-git clone https://github.com/asdf-vm/asdf.git ~/.asdf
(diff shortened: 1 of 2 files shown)


tuvistavie:
As for #86, I suggest we append `-dev` to the current version in development.
Otherwise, I think it's good, as long as we do not forget to update the README :smile:

Stratus3D:
@tuvistavie yeah, I'm not a fan of having to remember multiple things when incrementing versions. What do you think about having a bash script to do this for us? I'd check it in somewhere in repo, and it would take the version and update the readme with the tag and then tag the release.

Stratus3D:
Updated the current version with the `-dev` suffix.

PR #86, opened 2016-07-30
Add CHANGELOG.md

Stratus3D:
I'm a fan of flat files for this sort of thing, but this may be overkill. @HashNuke @tuvistavie what do you guys thing? Version 0.2.0 hasn't been tagged yet, but I went ahead and added a section for it, so that when tagging happens no changes will need to be made to CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+#Changelog
+
+##0.2.0-dev
+
+Features
+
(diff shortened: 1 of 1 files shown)


tuvistavie:
I think it's a very good idea to keep a proper changelog file, thanks! :+1: 
I suggest we postfix `-dev` or something if the version is not yet released,
so that people do not get confused.

Stratus3D:
Good idea @tuvistavie. I will update the file.

Issue #81, opened 2016-07-14
[Proposal] Update Legacy File Plugin API


rockwood:
The plugin api allows specifying alternative files to `.tool-versions` (called legacy files) which is very useful for interoperating with other version managers that use files like `.ruby-version` or `.node-version`.

The current implementation works great, but has a couple of drawbacks:

  1. Plugins must perform all the work of searching for legacy files and parsing their results, making it possible to plugins to have different behavior (one plugin might look only in the current directory, while another might look in all parent directories)

  2. Plugins don't expose the file(s) their looking for so `asdf` can't tell a user what file derived the current tool version if it came from a legacy file.

## Proposal:

Update the plugins api to make legacy file settings more deterministic. This would involve two new apis:

  1. `list-legacy-filenames` (required - for plugins that want to implement legacy files)

    Like `list-all` and `list-bin-paths`, this would take 0 arguments and return a space separated list of legacy filenames that are supported by this plugin. 

  Example output: `.node-version .nvmrc`

  2. `parse-legacy-file`(optional)

    Allows plugins to support alternative syntaxes beyond the basic version number. It would take 1 argument: the path to the legacy file, and return the version. (See issue: https://github.com/asdf-vm/asdf-ruby/issues/14)

    By default, asdf will assume the file just lists the version number alone and will simply `cat` it if `parse-legacy-file` is not implemented.

  Example output: `0.1.0`

## Benefits of this change

ASDF will be able to search for version setting files in a more deterministic (and efficient) manner.  It can recursively search parent directories for both `.tool-version` files and legacy files simultaneously. 

The plugins currently supported by asdf simply `cat` the contents of their legacy files so they would only need to implement `list-legacy-filenames` (relying on the default implementation of `parse-legacy-file`

## Potential Issues

If we go forward with these changes, do we attempt to maintain the existing `get-version-from-legacy-file`? Removing it obviously is easier, but raises compatibility issues. Since legacy files are turned off by default, maybe we can assume that users who have turned it on are comfortable with upgrading their plugins.

Plugins that want to support both new and old versions of asdf would simply leave their existing implementations of  `get-version-from-legacy-file`

Stratus3D:
I added the legacy file support and I agree with everything here. The less work the plugins have to do the better. I might be able to create a pull request sometime this week.

Do you happen to know which plugins include `get-version-from-legacy-file`? I think I only added it to the Ruby plugin. If it's not widely used we might be able to get away with just dropping support for it.

[...]

PR #87, opened 2016-07-31
Legacy File Redux

rockwood:
This implements the proposal in #81.

Now, plugins register the legacy filenames they support. ASDF searches for both legacy and .tool-version files simultaneously.

Users with plugins that only support the old legacy file api will see a warning message with instructions to update their plugin.

### DIFF
--- a/docs/creating-plugins.md
+++ b/docs/creating-plugins.md
@@ -50,9 +50,19 @@ Setup the env to run the binaries in the package.
 
 Uninstalls a specific version of a tool.
 
-#### bin/get_version_from_legacy_file
+#### bin/list-legacy-filenames
 
-Prints the version of the tool to use if a legacy version file is found in the current directory (e.g. rbenv's `.ruby-version`). Current directory is passed as the only argument to this script.
+Register additional setter files for this plugin. Must print a string with a space-seperated list of filenames.
+
+```
+.ruby-version .rvmrc
+```
+
+Note: This will only apply for users who have enabled the `legacy_version_file` option in their `~/.asdfrc`.
+
+#### bin/parse-legacy-file
+
+This can be used to further parse the legacy file found by asdf. If `parse-legacy-file` isn't implemented, asdf will simply cat the file to determine the version.
 
 ### Custom shim templates
 
--- a/docs/creating-plugins.md
+++ b/docs/creating-plugins.md
@@ -62,7 +62,7 @@ Note: This will only apply for users who have enabled the `legacy_version_file` 
 
 #### bin/parse-legacy-file
 
-This can be used to further parse the legacy file found by asdf. If `parse-legacy-file` isn't implemented, asdf will simply cat the file to determine the version.
+This can be used to further parse the legacy file found by asdf. If `parse-legacy-file` isn't implemented, asdf will simply cat the file to determine the version. The script will be passed the file path as its first argument.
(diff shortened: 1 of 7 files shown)


test/utils.bats
```diff
@@ -38,8 +41,100 @@ teardown() {
...
+  run check_if_plugin_exists "dummy"
+  [ "$status" -eq 0 ]
+  [ "$output" = "" ]
+}
+
+@test "check_if_plugin_exists should be noop if plugin exists" {
+  run check_if_plugin_exists "dummy"
+  [ "$status" -eq 0 ]
+  [ "$output" = "" ]
+}
```
Stratus3D: Duplicate of the test above.

Stratus3D:
This is awesome! I've had this PR open in a tab for the last week and just got around to reading the code. I think this is exactly what we need. 

We might want to implement the new legacy version scripts in the plugins that contain the get_version_from_legacy_file script before merging this though, since the updated plugins will continue to work with the current version of asdf, but once this is merged it won't work with old plugins (other than the warning).

rockwood:
@Stratus3D Good catch. I removed the duplicate test and rebased master. 

I just opened PRs for the two plugins that currently support the legacy file api.

https://github.com/asdf-vm/asdf-ruby/pull/17
https://github.com/tuvistavie/asdf-python/pull/4

Once those get merged, we should be good to merge this.

rockwood:
I'm closing this for the time being due to a bug with conflicting local and global versions. I'll reopen when I get that fixed

rockwood:
Ok, all fixed.

Stratus3D:
Great! I just merged the PR for the Ruby plugin, and it looks like the Python one has already been merged. 

I would like to have one more maintainer look at this before merging. @asdf-vm/maintainers what do you all think about these improvements to the legacy version file feature?

docs/creating-plugins.md
```diff
@@ -50,9 +50,19 @@ Setup the env to run the binaries in the package.
...
+
+```
+.ruby-version .rvmrc
+```
+
+Note: This will only apply for users who have enabled the `legacy_version_file` option in their `~/.asdfrc`.
+
+#### bin/parse-legacy-file
+
+This can be used to further parse the legacy file found by asdf. If `parse-legacy-file` isn't implemented, asdf will simply cat the file to determine the version.
```
tuvistavie: Maybe we could add

> The script will be passed the file path as its first argument.

or something like this to make it clear that we are passing the file path, and not its content.
I was not totally sure which one was passed before checking the implementation.

tuvistavie:
It looks great, thank you very much! :smile:
I just left a little comment about the README, let me know what you think!

docs/creating-plugins.md
```diff
@@ -50,9 +50,19 @@ Setup the env to run the binaries in the package.
...
+
+```
+.ruby-version .rvmrc
+```
+
+Note: This will only apply for users who have enabled the `legacy_version_file` option in their `~/.asdfrc`.
+
+#### bin/parse-legacy-file
+
+This can be used to further parse the legacy file found by asdf. If `parse-legacy-file` isn't implemented, asdf will simply cat the file to determine the version. The script will be passed the file path as its first argument.
```
rockwood: @tuvistavie Good call.

PR #88, opened 2016-08-13
Add section to readme on the .asdfrc file.

Stratus3D:
I noticed we didn't have the .asdfrc file documented anywhere so I figured I would add a section to the readme.


tuvistavie:
Looks great!

PR #90, opened 2016-08-16
Local whitespace fix

### LINKED ISSUE #89: Having issues with folders with spaces in them on OS X El Capitan.
`asdf local <name> <version>`
seems to fail if i have a folder with spaces in the name. Ive tried using the asdf plugin for nodejs

Stratus3D:
This should fix #89. This is my first time writing bats tests and also my first time working with the code for the local/global setters, so I'd like someone look this over and provide some feedback. 

- Are there other things I should test? 
- Are there other places in the code that might contain similar bugs?

### DIFF
--- a/lib/commands/version_commands.sh
+++ b/lib/commands/version_commands.sh
@@ -18,10 +18,10 @@ version_command() {
   check_if_plugin_exists $plugin
   check_if_version_exists $plugin $version
 
-  if [ -f $file ] && grep $plugin $file > /dev/null; then
-    sed -i -e "s/$plugin .*/$plugin $version/" $file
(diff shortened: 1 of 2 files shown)


tuvistavie:
Great job, thanks! It seems I forgot about spaces when adding this command :smile:  
I looked a little around, but could not find any other place where we forgot quotations,
so I think this should be enough.
Thank you very much!

Stratus3D:
@tuvistavie at first glance it also looks like this may be an issue in v0.1.0, should I branch off of v0.1.0 and a PR for it as well? We could tag it as v0.1.1. That way those that want to stay on 0.1.* could get the fix if they need to. Alternatively we could tag v0.2.0 now.

PR #94, opened 2016-09-18
Versions prefixed with "path:" no longer seem to work

bsedat:
"path:" prefixed versions no longer seem to work and instead error with "No such command in path of $plugin_name". This new behavior seems related to d505c0ee7127ae4a9516f6f3f8046e5e4865700e.

The first commit adds a failing test case for `get_preset_version_for` that shows it losing information when parsing a version prefixed with "path:". I made an attempt at fixing this behavior  in the second commit by switching the return value of `get_preset_version_for` to be space-delimited and then parsing that for `current_command`.

I'm happy to take suggestions for other ways to fix this (my bash is pretty rudimentary), or feel free to edit my branch if you have other ideas.

### DIFF
--- a/lib/commands/current.sh
+++ b/lib/commands/current.sh
@@ -4,9 +4,12 @@ current_command() {
   check_if_plugin_exists $plugin_name
 
   local search_path=$(pwd)
-  local version_and_path=$(find_version "$plugin_name" "$search_path")
-  local version=$(cut -d ':' -f 1 <<< "$version_and_path");
-  local version_file_path=$(cut -d ':' -f 2  <<< "$version_and_path");
+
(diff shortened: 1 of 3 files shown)


bsedat:
Are we moving away from space-delimiters to better support paths with spaces? Would love to fix this bug but don't want to break any of that behavior.

tuvistavie:
Hi, thank you very much for your PR, and sorry for the delay.
The approach is good but using spaces as a delimiter could break things,
so I think we should go with #112.

bsedat:
Sounds good! I mainly went for convention with the other splitting I saw, but I agree that another delimiter is probably better.

PR #96, opened 2016-09-21
Fix getting spaces to work in directories.

hoangelos:


HashNuke:
Hello @hoangelos,
Just looked at the PR. I think there are some changes that revert stuff in master. Like the version `0.2.0-dev`. Can you please try the actually changes with current master?

hoangelos:
I did make the change based off the current master.  I did a pull and a rebase and it said there was nothing else to do.  Is there something different you are asking?

Stratus3D:
@hoangelos It looks to me like your changes are based off of the v0.1.0 tag. Are you sure you branched off of master? I see things like this in the diff:

```
-  echo "0.2.0-dev"
+  echo "0.1"
```

This leads me to believe this was based on version 0.1. This change, and many of the others changes that revert back to old code shouldn't be merged into master. 

Thanks for taking the time to create a PR though. I'm aware of this bug and we do need to fix it.

Stratus3D:
This has been fixed in PR #104

PR #103, opened 2016-10-25
Add MongoDB plugin to README Usage list

brianvanburken:
Found this plugin when looking for a way to switch MongoDB versions. Since it is not listed I decided to add it and create this PR.

A possible improvement is to split up these plugins in categories. Currently the first column reads `Language` while the list contains a few non-language plugins like Postgres, Redis, Riak and MongoDB. What do you think?


Stratus3D:
@brianvanburken good find! I wasn't aware that plugin existed. 

It looks like they don't have a travis build setup for the plugin, so I can't verify that it passes the `asdf plugin-test` test. It must pass the test before being listed in the README. I've opened an issue for this: https://github.com/sylph01/asdf-mongodb/issues/1 Once this is addressed I will merge this PR.

PR #104, opened 2016-10-25
More whitespace fixes

Stratus3D:
Also added a test for this issue so I could verify that it was in fact broken.


tuvistavie:
LGTM! Thanks for taking care of #96!

PR #105, opened 2016-10-27
file_path may have some spaces.

namjae:


Stratus3D:
Was there a specific issue that this fixes? I'm curious because we have similar unquoted file_path variables in other functions so I suspect there could be other bugs present.

namjae:
i got a msg like this this
`/home/dhpark/.asdf/lib/utils.sh: line 97: [: /home/dhpark/study/elixir/learning: binary operator expected`
actual CWD was `/home/dhpark/study/elixir/learning elixir/hello_world`

Stratus3D:
Great! I'm going to add some tests to this just to make sure there isn't anything else we've missed. Then I'll merge. Thanks!

HashNuke:
@Stratus3D after @namjae merge the latest master branch, the change in this PR is down to 1-line. So merging it right away.

Issue #91, opened 2016-09-03
Proper way to set the versions outside of $HOME?


Stratus3D:
Today I fired up prax (https://github.com/ysbaddaden/prax) only to find that it wasn't working and was printing "No version set for ruby". After fiddling with the .praxconfig file (https://github.com/ysbaddaden/prax/wiki/Ruby-Version-Managers) and doing some debugging I found that Prax invokes `ruby` from `/opt/prax/`. I finally just ran `sudo cp ~/.tool-versions /opt/prax/` to make the settings in my tool version file available to prax.

Do we have a recommended way of handling scenarios like this? To me this seems kind of ugly, and I'll probably delete the `.tool-versions` file and replace it with a symlink to the one in my home directory so I can change the versions.

I'll be adding a section on asdf to the prax wiki based on what is decided here.

HashNuke:
@Stratus3D I think I wrote `path:/foo/bar` support for the elixir plugin because a lot of people I came across were using elixir master which was available somewhere in their local.

Stratus3D:
I forgot about that method. Thanks!

PR #106, opened 2016-10-29
Determine version from global file when project is outside home.

vic:
Without this patch, asdf was not able to determine the tool version
for a project located outside the user's HOME directory.

```
/work/project/
/home/me/.tool-versions
```

This changeset lets asdf find the global version stored at
$HOME/.tool-versions when the directory traversal from
the project dir was not able to find a suitable version.

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -56,6 +56,28 @@ display_error() {
   echo >&2 $1
 }
 
+get_version_in_dir() {
+  local plugin_name=$1
+  local search_path=$2
+  local legacy_filenames=$3
+
+  local asdf_version=$(parse_asdf_version_file "$search_path/.tool-versions" $plugin_name)
+
+  if [ -n "$asdf_version" ]; then
+    echo "$asdf_version:$search_path/.tool-versions"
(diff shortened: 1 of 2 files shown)


tuvistavie:
Hi @vic, thanks for the PR!
I think this is a good change, I often run into this issue when building something under `/tmp`.
@asdf-vm/maintainers What do you think?

Stratus3D:
I actually came across this issue recently as well. I was trying to set the Ruby version in a `/opt/` directory #91. I guess I don't have an opinion either way and the PR looks good. I would like to have @HashNuke's input before merging though. It seems like it might just better to write a `.tool-versions` file directly to `/` for system stuff, that way it's harder for asdf to interfere with existing tools.

HashNuke:
@asdf-vm/maintainers I've always assumed that the project dir or atleast it's parent dir(`/projects`, etc) would have the .tool-versions file. We can merge if you guys are fine with it.

@vic Thanks for the PR ~!

vic:
You're welcome and thanks for merging this :)

PR #110, opened 2016-11-10
Get version from env if ASDF_${TOOL}_VERSION is defined.

### LINKED ISSUE #49: [improvement] shell command
there are local and global commands that give you ability to set local or global version however sometimes we need to temporary set version for currently running shell session. 
So what I propose is addition of shell command

vic:
For example if asdf finds that the `ASDF_FOO_VERSION` is
defined in the current environment it will override the
version of `foo` from the `.tool-version` file.

Closes #49

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -82,6 +82,12 @@ find_version() {
   local plugin_name=$1
   local search_path=$2
 
+  local version=$(get_version_from_env "$plugin_name")
+  if [ -n "$version" ]; then
+      echo "$version"
+      return 0
+  fi
+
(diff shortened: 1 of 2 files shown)


HashNuke:
@vic Thanks for the fix. I'll try this out tonight and merge it.

tuvistavie:
I just noticed this PR was still opened.
@asdf-vm/maintainers I think we can merge, wdyt?

vic:
@tuvistavie rebased this PR on current master.

vic:
merging this :)

Issue #95, opened 2016-09-18
Error when setting "path:"-prefixed versions via `asdf global` or `asdf local`


bsedat:
```
$ asdf global ruby path:/usr
version path:/usr is not installed for ruby

$ asdf local ruby path:/usr
version path:/usr is not installed for ruby
```

PR #112, opened 2016-11-12
Bugfix: `ref:` and `path:` versions are not read correctly.

vic:
Before this patch, with a `.tool-versions` file like:

```
lfe ref:master
```

`get_preset_version_for` would return `ref` instead of `ref:master`.

Same was happening for `path:` versions. Actually there was PR #94 
on which I based my changes but instead of using space as delimiter
I went for using `|` which would be a lot more weird if present as
part of a file path, this also allows to specify paths which have
spaces which are much more frequent.

Closes #94 #95

### DIFF
--- a/lib/commands/current.sh
+++ b/lib/commands/current.sh
@@ -5,8 +5,8 @@ current_command() {
 
   local search_path=$(pwd)
   local version_and_path=$(find_version "$plugin_name" "$search_path")
-  local version=$(cut -d ':' -f 1 <<< "$version_and_path");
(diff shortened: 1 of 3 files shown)


tuvistavie:
Thank you very much!
I agree using `|` as a delimiter seems more robust than using spaces.
LGTM, I will wait someone else to review and I think we can merge this!

vic:
@HashNuke or @Stratus3D could any of you review and see if we are ok merging this?

vic:
@tuvistavie rebased changes on current master.

tuvistavie:
Thanks!

PR #113, opened 2016-11-23
Add asdf-elm plugin.

vic:
[vic/asdf-elm](http://github.com/vic/asdf-elm) plugin for [Elm](http://elm-lang.org)


Stratus3D:
This is awesome! Thanks for the PR!

Couple things I noticed when looking at the code:

* You might want to specify OSX and Linux in your .travis.yml file so you can verify it works on both platforms:
```
os:
  - linux
  - osx
```
* Why is the exec-env file needed? It doesn't seem to be used, in which case it should be removed.
* Nice trick with sort!

vic:
@Stratus3D hey thanks for merging, just fixed the travis file to test on both platforms and removed the unused exec-env file, thanks for the heads up.

PR #114, opened 2016-11-26
add julia lang plugin

rkyleg:


Stratus3D:
Nice! Thanks for the PR!

Couple things I noticed when looking at the code:

* The list-all-test and list-bin-paths should be removed from the plugin root. They aren't used by asdf.
* Since the list-all script uses python, you might want to say so in the README so that users are aware of that.
* Along with that, you might also want to flesh out the README with the basic details on the plugin, how to install it, and anything special it does when installing Julia (I saw it's symlinking to a directory outside of ~/.asdf).

Thanks for creating the plugin!

rkyleg:
Thanks for the suggestions, I will make the updates!

PR #115, opened 2016-11-29
Add Idris plugin

vic:
[plugin](http://github.com/vic/asdf-idris) for installing [Idris](http://idris-lang.org)


Stratus3D:
Another plugin! Yay!

Stratus3D:
Actually, it looks like two of the build jobs are failing and only one is succeeding. It looks like the error is due to the `idris` command not being found.

vic:
@Stratus3D  Yeah, i'm working on it, the one succeeding is just installing the OSX binary, the other two are building Idris with cabal (once on linux and other on osx), and it's taking lots of time, so I'm trying to get those tests green.

vic:
@Stratus3D  Yay, we are finally green !  

but look at how much time it takes to build from source:

https://travis-ci.org/vic/asdf-idris/builds/179768882

Stratus3D:
Wow. I'm not surprised though. At least it's automated :)

PR #116, opened 2016-12-02
Added LuaJIT

smashedtoatoms:


smashedtoatoms:
This doesn't have tests for luarocks because it is dependent on luaJIT being installed.  I can't seem to get the dependency to behave.  Any recommendations would be welcome.

smashedtoatoms:
nm, they're working now.

Stratus3D:
@smashedtoatoms will your luarocks plugin work with plain old Lua? My lua plugin installs luarocks along with Lua, but maybe it doesn't need to if we have a luarocks plugin.

smashedtoatoms:
It won't.  It's actually kind of a hack job to get it to find the luaJIT install.  If your regular Lua installer installs luarocks, maybe my luaJIT one should as well.  I can't think of a great reason not to.  What do you think?

Stratus3D:
I think it's probably better to keep them separate, because there are multiple versions of luarocks too. My script just grabs the latest stable luarocks version and tries to install it, but it might be nice to be able to specify which luarocks version should be used.

smashedtoatoms:
I'm playing around with specifying it in the version with a `::` delimiter.  You could do something like `2.0.3::2.4.2` to specify that you want luaJIT 2.0.3 with luarocks 2.4.2.

Stratus3D:
Sorry for taken so long to merge this. I was actually thinking of trying to split luarocks into a separate plugin, but after looking at the installation instructions it looks like that isn't possible, not even with some hacking.

I'll see if I can update the version syntax for my Lua plugin so it behaves like yours in regard to luarocks version. I might also see if I can have mine install a default version of luarocks if no version is specified, might be useful for those that don't care about the version but still need it installed.

PR #117, opened 2016-12-05
Ensure install commands always add asdf loading code on a new line in bashrc.

### LINKED ISSUE #98: -bash: asdf: command not found
I made all steps - installed all packages for Mac OS and updated `.bash_profile`, but have:

>  -bash: asdf: command not found

Stratus3D:
This could fix #98. @HashNuke what do you think? Good solution or ugly hack?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,12 @@ git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.2.0
 Depending on your OS, run the following
 ```bash
 # For Ubuntu or other linux distros
-echo '. $HOME/.asdf/asdf.sh' >> ~/.bashrc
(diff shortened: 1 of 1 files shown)


BenMorganIO:
I think those quotes have to be changed to double quotes. This broke on an Ubuntu machine during a copy and paste of those lines.

Stratus3D:
@BenMorganIO Good catch, it looks like the `-e` flag is required. Double quotes don't seem to have any effect (at least on CentOS and OSX). I'll test on Ubuntu tonight.

Stratus3D:
@BenMorganIO Just merged that PR. Want to give it a try on Ubuntu?

BenMorganIO:
@Stratus3D works like a charm! :D

PR #121, opened 2016-12-10
Add asdf-clojure

### LINKED ISSUE #101: Plugin wanted: Clojure
First thanks for asdf - setting up to use it as my version manager for all the things supported (tm).

Doing that I noticed that there is no [Clojure](http://clojure.org/) plugin yet. I didn't find any information on where to request plugins so I'm trying that here and now: A Clojure plugin would be lovely. 

Thanks! :heart: 
Tobi

vic:
[asdf-clojure](http://github.com/vic/asdf-clojure) is a plugin for installing [Clojure](http://clojure.org)

Closes #101

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Plugins are how asdf understands how to handle different packages. Below is a li
 
 | Language  | Repository  | CI Status
 |-----------|-------------|----------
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks @vic !

PragTob:
Thank you! :)

Issue #67, opened 2016-06-21
Plugin uninstall does not remove that plugin's shims


duijf:
**Steps to reproduce**

```
$ asdf plugin-add ruby https://github.com/asdf-vm/asdf-ruby.git
$ asdf install ruby 2.3.1
$ asdf global ruby 2.3.1
$ gem install bundler
$ asdf plugin-remove ruby
$ bundle
No such plugin
```

**Observed behavior**

Removing a plugin, leaves it's shims in the asdf shims directory, which leaves them polluting my PATH.

**Expected behavior**

When removing a plugin, all it's shims are removed as well. Calling bundle should give a "command not found" error.

HashNuke:
@duijf true. We should fix this ~!

duijf:
So I thought about this a bit, and I think the simplest way to go about things is to namespace the directory shims are stored per plugin. So instead of this:

```
❯ ls ~/.asdf/shims/ 
bundle*    elixirc*  escript*   listen*     rake*     sass-convert*
bundler*   epmd*     gem*       middleman*  rdoc*     scss*
ct_run*    erb*      haml*      mix*        ri*       thor*
dialyzer*  erl*      iex*       node*       ruby*     tilt*
dotenv*    erlc*     irb*       npm*        run_erl*  to_erl*
elixir*    erubis*   kramdown*  rackup*     sass*     typer*
```

Doing something like this:

```
❯ ls ~/.asdf/shims/ 
ruby/   elixir/    nodejs/    erlang/
```

And having `asdf plugin-remove` do a `rm -r asdf/shims/<plugin-name>`

I'll see if I can help with a PR, if this works for you.

HashNuke:
@duijf I went with this because it makes the importing of shims easier (only one dir to add to `PATH`). But I have this feeling that I might have missed something. Any suggestions to make adding to `PATH` easier but keep the per-plugin shim dir like you mentioned?

\\cc: @asdf-vm/maintainers

duijf:
Yeah, the downside is that you need to loop over that directory and add everything to the `PATH`. A huge PATH might actually be a dealbreaker.

Another option is to add a comment to each shim with the metadata about which plugin it belongs to.

```bash
echo """#!/usr/bin/env bash
# asdf-plugin: ${plugin_name}
exec $(asdf_dir)/bin/private/asdf-exec ${plugin_name} ${executable_path} \"\$@\"
""" > $shim_path
```

[...]

PR #122, opened 2016-12-10
Add metadata to shims

vic:
When a shim is created, add plugin metadata so we can later know which shims belong to which plugins, this will help aid with removing unused shims on uninstall. See #67

```
# asdf-plugin: ${plugin_name}”
```

Thanks to @duijf for the metadata proposal.


tuvistavie:
Thanks for working on this, LGTM!

Stratus3D:
Looks good to me!

vic:
@Stratus3D thanks for merging this 👍

PR #123, opened 2016-12-10
Shims remove on latest version uninstall (depends on #122)

### LINKED ISSUE #67: Plugin uninstall does not remove that plugin's shims

vic:
When the latest version of a tool is uninstalled,
Remove the plugin shims (marked with metadata at #122)

Also found lots of missing tests and added them.

Closes #67

### DIFF
--- a/lib/commands/list.sh
+++ b/lib/commands/list.sh
@@ -1,16 +1,15 @@
 list_command() {
   local plugin_name=$1
-  local plugin_path=$(get_plugin_path $plugin_name)
   check_if_plugin_exists $plugin_name
 
-  local plugin_installs_path=$(asdf_dir)/installs/${plugin_name}
+  local versions=$(list_installed_versions $plugin_name)
 
-  if [ -d $plugin_installs_path ]; then
-    #TODO check if dir is empty and show no-installed-versions msg
-    for install in ${plugin_installs_path}/*/; do
-      echo "$(basename $install)"
+  if [ -n "${versions}" ]; then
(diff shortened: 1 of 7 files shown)


tuvistavie:
@vic Thanks for this feature!
I think we should also remove the shims on `plugin-remove`, wdyt?

vic:
@tuvistavie yes we should. I'll rebase with master and implement on `plugin-remove`

vic:
@tuvistavie done, PR is ready.

tuvistavie:
Great, thank you, I think we can merge this!
@asdf-vm/maintainers Seems good?

Stratus3D:
Looks good to me! I'm always happy when a PR includes more tests 😄 

It will probably be a good idea to tag the 0.2.1 release after this is merged, since we've had quite a few fixes recently.

Stratus3D:
I think this is a pretty safe PR, so I'm going to merge now.

PR #128, opened 2016-12-13
Test produced shims by actually running them.

### LINKED ISSUE #127: Shims were broken and not caught by unit tests
I'd like to have some unit tests that actually test some dummy shims so we can be sure they work. The test would be something along the lines of

* Create a plugin
* reshim it
* Test that the shims actually work without returning non-zero exit codes

vic:
Test using a dummy executable.

Closes #127


Stratus3D:
Wow! I was figuring I would write this myself tonight. Thanks!

PR #129, opened 2016-12-14
Add a couple lines to the creating-plugins guide on version ordering.

Stratus3D:
This was something we talked about on #33 but never added to the plugin guide.

### DIFF
--- a/docs/creating-plugins.md
+++ b/docs/creating-plugins.md
@@ -23,6 +23,10 @@ Must print a string with a space-seperated list of versions. Example output woul
 1.0.1 1.0.2 1.3.0 1.4
 ```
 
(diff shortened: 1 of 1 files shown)


vic:
@Stratus3D what about suggesting a *better sort* for OSX environments. Turns out the `sort` command on OSX does not support the `-V` flag (version sort), thus some plugins like [asdf-elixir simply use `sort`](https://github.com/asdf-vm/asdf-elixir/blob/master/bin/list-all#L12) which wont actually be the best option. There's an alternative [sort function](https://github.com/vic/asdf-idris/blob/master/bin/list-all#L6) I used on `asdf-idris` and seemed to work much better even for osx.

Stratus3D:
Good idea @vic. Added that to the file and linked to your sort function.

PR #130, opened 2016-12-15
Add D(DMD) plugin

sylph01:
Updates: Synced the repository from #48, and removed Scala plugin.
(Scala plugin from #126 seems to handle version updates automatically, which I haven't been able to do)


tuvistavie:
Thank you! :smile:

PR #135, opened 2016-12-18
Add a helpful error message to the install command

### LINKED ISSUE #132: install command does not check for a version
If I run `asdf install <plugin_name>` the command won't fail, but claim lua is already installed. It's clear  with two spaces where the version should be displayed.

```
$ asdf install lua
lua  is already installed
```

I can't think of any reason this behavior would be desirable. All versions must of consist of something more than whitespace.

Stratus3D:
Add a helpful error message to the install command to warn users when they don't specify a version to install. I came across this a couple days ago, and was surprised when asdf behaved as if nothing had gone wrong. Adding a simple arity check fixes this.

This fixes #132.

### DIFF
--- a/lib/commands/install.sh
+++ b/lib/commands/install.sh
@@ -4,6 +4,9 @@ install_command() {
 
   if [ "$plugin_name" = "" ] && [ "$full_version" = "" ]; then
     install_local_tool_versions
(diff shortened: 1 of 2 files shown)


tuvistavie:
LGTM!

PR #136, opened 2016-12-18
Prepare for 0.2.1 release.

Stratus3D:
We've had a lot of bugfixes over the last month, so I figured it's about time to tag another patch version. I don't see anything here that would break the existing 0.2.0 API, so I think it's safe to just increment the patch version. With some of the upcoming changes there will likely be some API differences, so the next release after this one will probably be 0.3.0.

I'm going to wait until #135 is merged before merging this. Once this is merge I'll tag v0.2.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,18 @@
 #Changelog
 
-##0.2.0-dev
+##0.2.1-dev
+
+##0.2.1
 
 Features
 
 * Determine global tool version even when used outside of home directory (#106)
+
+Fixed Bugs
+
+* Correct reading of `ref:` and `path:` versions (#112)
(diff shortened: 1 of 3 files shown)


Stratus3D:
Updated the README and utils.sh with the new version too.

tuvistavie:
Thanks! I think it is indeed a good idea to release a new version.
If we are sure that #139 does not affect a significant number of users, I think we can merge this!

Stratus3D:
I'm going to go ahead and merge this. I don't think #139 affects many users. If it does turn out to be a common issue we can add a warning to asdf and tag 0.2.2.

tuvistavie:
Great, thanks! :+1:

Issue #124, opened 2016-12-10
Updating plugins fails with "You have unstaged changes"


akeating:
```
asdf --version
0.2.0-dev
```
```
asdf plugin-update --all
Updating elixir...
Current branch master is up to date.
Updating erlang...
Current branch master is up to date.
Updating mongodb...
Current branch master is up to date.
Updating nodejs...
Current branch master is up to date.
Updating postgres...
error: Cannot pull with rebase: You have unstaged changes.
Updating python...
error: Cannot pull with rebase: You have unstaged changes.
Updating ruby...
error: Cannot pull with rebase: You have unstaged changes.
```
and looking into `postgres`
```
postgres adam$ git diff
diff --git a/bin/exec-env b/bin/exec-env
old mode 100644
new mode 100755
diff --git a/bin/install b/bin/install
old mode 100644
new mode 100755
diff --git a/bin/list-all b/bin/list-all
old mode 100644
new mode 100755
```

Stratus3D:
Hi @akeating, just checked and that's definitely a bug. It looks like we correct the permission of plugin scripts after we clone down the plugin repository (lib/commands/plugins-add.sh line 19). In the meantime you can go to the plugin directory `cd ~/.asdf/plugins/postgres` and checkout the changes `git checkout .`, then after updating the plugins run `chmod +x bin/*` to fix the permissions.

There are a couple ways we could solve this issue:
* Remove the code adds the execution permission and force all plugins to correct the permission issues. We would also need to add a check for correct permissions to the plugin-test command so plugin maintainer's builds would fail when permissions are incorrect.
* When upgrading plugins, run `git checkout` before `git pull`.
* Configure the plugin repositories so file permissions are ignored, and just re-correct permissions each time a plugin is updated.

To me the first option seems like the best approach, but it's also the one that will likely break a lot of plugins initially. @HashNuke @tuvistavie what do you think?

tuvistavie:
I think the first solution is definitely what we should go with.
I think we should proceed in several phases

1. Add tests for plugins
2. Warn the plugin authors (at least for the ones we have on the README)
3. Emit a warning when installing a plugin with no executable files
4. Remove the code changing the permission

What do you think?

Stratus3D:
@tuvistavie that sounds good. I'll put together a PR to update the plugin test.

akeating:
Thanks everyone, this seems a good solution.

Stratus3D:
[...]

PR #138, opened 2016-12-22
Check permissions on scripts in plugins' bin directories when testing

Stratus3D:
This is the first in a series of steps to address issue #124.

Plugins will now fail the test if the scripts in `bin/` are not executable.


tuvistavie:
LGTM! Thank you :smile:

PR #140, opened 2016-12-23
Add Haskell plugin

vic:
[asdf-haskell](http://github.com/vic/asdf-haskell) is a plugin for installing Haskell (via stack)


tuvistavie:
Nice, LGTM!

PR #141, opened 2016-12-24
Add couple of tests

Stratus3D:
Thanks for the additional tests @vic. Since tests currently take only 13 seconds I don't see any reason why we shouldn't add a few more.


tuvistavie:
LGTM, thank you :smile:

PR #142, opened 2016-12-24
Fix fish shell's deprecation warnings when using the `complete` command

crzrcn:
Attempting to tab-complete an asdf command would result in a deprecation warning being printed.

The --authoritative/-A and --unauthoritative/-u flags have been removed from the `complete` command.

Context: https://github.com/fish-shell/fish-shell/pull/3660


vic:
Nice, thanks 👍  :)

tuvistavie:
I did not know about this, thanks!

PR #151, opened 2017-02-07
Add my packer plugin

gozer:


Stratus3D:
It looks like the build is currently failing. Can you get the build working on Linux and OSX?

It also looks like your plugin uses on Ruby to sort versions. In the plugin guide we suggest only sorting versions if necessary. If sorting is necessary there is a simple function that can be used to sort versions https://github.com/asdf-vm/asdf/blob/master/docs/creating-plugins.md#binlist-all

gozer:
All great suggestions, I didn't take the time originally to read the plugin docs ;-(

I had just ripped the terraform plugin more or less blindly.

Converted to bash just fine, and I believe the tests will be passing shortly.

Thanks for the great feedback!

gozer:
There was a macosx backlog in Travis land, but builds are green now.

Stratus3D:
Yeah I waited for the OSX build to finish. Looks good now. Thanks!

You may also want to add some details to the readme. Show how to install the plugin, list the packer dependencies, and document any custom options.

PR #153, opened 2017-02-10
Test if plugin can be updated.

fazibear:
Just in case of
```
error: cannot pull with rebase: You have unstaged changes.
error: please commit or stash them.
```

### DIFF
--- a/lib/commands/plugin-test.sh
+++ b/lib/commands/plugin-test.sh
@@ -70,5 +70,10 @@ plugin_test_command() {
         fi
     done
 
+    (asdf plugin-update $plugin_name)
+    if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR @fazibear but this is something we've discussed on #124. I think it will be better to address the root cause (permission changes) rather than printing an error. Ideally a plugin update should only fail due to network issues. I think currently the error makes it to STDOUT when that happens.

fazibear:
I saw it, but testing it like this https://github.com/asdf-vm/asdf/pull/138 is pointless because it is already changed during plugin-add process.

Stratus3D:
@fazibear correct, and that has been fixed with #154.

PR #154, opened 2017-02-12
Don't change the permissions on files in a plugin's bin directory.

Stratus3D:
The next step in addressing #124. We should not alter plugin files at all.


tuvistavie:
Thanks, LGTM! :smile:

PR #156, opened 2017-02-12
export ASDF_DIR so that plugins can use it to refer to the ASDF DIR

ypid:
Also fixed `shellcheck` warnings in the file.

Ref: https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/asdf/asdf.plugin.zsh

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
 if [ "${BASH_SOURCE[0]}" != "" ]; then
-  current_script_path=${BASH_SOURCE[0]}
+  current_script_path="${BASH_SOURCE[0]}"
 else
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks! what does this line mean though? I'm pretty new to shellcheck.

    # shellcheck disable=SC2155,SC2164

ypid:
You are welcome!

It disables [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) and [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164). Now that I rechecked that, both warnings don’t have to be disabled but can instead be properly fixed. Done in #169.

Stratus3D:
Thanks!

PR #175, opened 2017-03-13
Avoid error when sysctl is missing hw.ncpu key

aeons:
When trying to use `asdf` in Alpine Linux in Docker, I got an error when trying to determine the number of CPUs, since Alpine has `sysctl` but the key `hw.ncpu` doesn't exist.

This adds an additional check for that case.


tuvistavie:
Looks good, thank you very much! :smile:

Issue #172, opened 2017-03-08
[Proposal] Command to update asdf itself?


Stratus3D:
It might be nice to have a command that would allow users to update asdf itself to the latest tagged version or the latest on the master branch. Right now it's a manual process that requires the user to:

     cd $ASDF_DIR
     git pull origin master # Or a specific tag

Might be nice to have a command that would do this for the user:

    asdf update-asdf # Checkout the most recent tag
    asdf update-asdf --latest # Pull the latest from the master branch

It could also warn the user if they switch to an older commit (e.g. moving from the master branch to the most recent tag that lacks some changes on master). What do you guys think? I guess we are only saving a couple commands, but it still might be nice to have an official way to update asdf.

tuvistavie:
I agree this would a good addition!

HashNuke:
👍  This would be awesome. We could just use `update` and `update --head`

PR #177, opened 2017-03-15
Asdf update command

Stratus3D:
This should complete #172.

### DIFF
--- a/lib/commands/update.sh
+++ b/lib/commands/update.sh
@@ -1,0 +1,28 @@
+update_command() {
+  local update_to_head=$1
+
+  (
+  cd $(asdf_dir)
+  pwd
+
+  if [ "$update_to_head" = "--head" ]; then
+    # Update to latest on the master branch
+    git checkout master || exit 1
+
+    # Pull down the latest changes on master
+    git pull origin master || exit 1
+    echo "Updated asdf to latest on the master branch"
+  else
+    # Update to latest release
+    git fetch --tags || exit 1
+    tags=$(git tag)
+
+    # Pick the newest tag
+    tag=$()
+
+    # Update
+    git checkout "$tag" || exit 1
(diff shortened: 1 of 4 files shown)


Stratus3D:
Updated this PR with tests for the update command.

lib/commands/update.sh
```diff
@@ -0,0 +1,33 @@
...
+    tag=$(echo $tags | cut -d ' '  -f1)
+
+    # Update
+    git checkout "$tag" || exit 1
+    echo "Updated asdf to release $tag"
+  fi
+  )
+}
+
+# stolen from https://github.com/rbenv/ruby-build/pull/631/files#diff-fdcfb8a18714b33b07529b7d02b54f1dR942
```
HashNuke: Haha. Thanks for adding appropriate credits :)

HashNuke:
Build breaks for some reason.

lib/commands/update.sh
Stratus3D: I actually stole this from the irdris plugin, which stole it from ruby build :p https://github.com/vic/asdf-idris/blob/master/bin/list-all#L6

Stratus3D:
Looks like `tail -r` to get the last item doesn't work on Linux. I switched to a simple sed find and replace expression to get the latest tag. We are already using sed in several places so it's guaranteed to work on any OS that asdf currently works on.

Stratus3D:
That fix was completely wrong. I've pushed the corrected the sed command.

Stratus3D:
This passes on both my Ubuntu laptop and my OSX laptop. I also ran the tests on my CentOS 6 server and tests passed there too. Not sure what is causing the tests to fail on Travis.

I tried using Vagrant but ran into issues, I'll see if I can get them sorted out tomorrow.

Issue #150, opened 2017-02-01
Pip doesn't get the correct version when virtualenv is activated


fcrespo82:
# System:
asdf 0.2.1
Python plugin installed with 2.7.10 and 3.6.0 versions

# Steps to reproduce:
```
$ asdf global 2.7.10
$ mkdir version3
$ cd version3
$ asdf local python 3.6.0
$ pip -V
```
Outputs the correct version 
`pip 9.0.1 from /Users/<username>/.asdf/installs/python/3.6.0/lib/python3.6/site-packages (python 3.6)`

```
$ virtualenv new
$ . new/bin/activate
$ pip -V
```
Outputs the WRONG version 
`pip 9.0.1 from /Users/<username>/version3/new/lib/python2.7/site-packages (python 2.7)
`

# Conclusion
I don't know if I'm doing something wrong, but I expected `pip` to get the version of the current local config.

If its by virtualenv's design, please add a command `asdf which python` to get the path of the current python so we can pass to virtualenv's `-p` flag.

Stratus3D:
@fcrespo82 `asdf current python` may give you the information you need to pass to `-p`. I'm not too familiar with Python, but that behavior may be intended. @tuvistavie is this the way pip and asdf-python are suppose to work?

tuvistavie:
`-p` option requires the path, so `asdf current` will not do for this case.

I explicitly pass the path to python when I create a new virtualenv:

```
virtualenv -p ~/.asdf/installs/python/3.5.2/bin/python myenv
```

it could also be done with

```
virtualenv -p $(asdf where python 3.5.2)/bin/python myenv
```

but this also requires to explicitly pass the version,
so I agree being able to change this into

```
virtualenv -p $(asdf which python) myenv
```

would be an improvement.

What do you guys think?

fcrespo82:
For workaround I am doing 

```
virtualenv -p $(asdf where python $(asdf current python))/bin/python myenv
```

Stratus3D:
@tuvistavie I think the addition of a `which` command would be a good idea. That would give us three related commands:

* `current` - print the current version and the path to the tool-version file
* `where` - show the path to the root of a specific version
* `which` - show the path to the current version of the specified executable

fcrespo82:
Related to #72

Stratus3D:
[...]

PR #178, opened 2017-03-16
Which command

### LINKED ISSUE #72: [Proposal] Command to get direct path of the binary

fcrespo82:
Hi guys, this is my attempt to do the which command it fixes #72 and #150.

I guess it could be better when returning the main binary, but I think would need a Plugin API change.

Hope it helps!

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -7,6 +7,7 @@ source $(dirname $(dirname $0))/lib/commands/install.sh
 source $(dirname $(dirname $0))/lib/commands/uninstall.sh
 source $(dirname $(dirname $0))/lib/commands/current.sh
 source $(dirname $(dirname $0))/lib/commands/where.sh
+source $(dirname $(dirname $0))/lib/commands/which.sh
 source $(dirname $(dirname $0))/lib/commands/version_commands.sh
 source $(dirname $(dirname $0))/lib/commands/list.sh
 source $(dirname $(dirname $0))/lib/commands/list-all.sh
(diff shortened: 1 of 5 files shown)


tuvistavie:
Thank you very much, and sorry for the delay.
I will try the branch locally, and let you know if I run into any issue.
If you get a moment, could you rebase on top of master as it seems we introduced some changes to `completions/asdf.bash` at some point.
Again, thank you :smiley:

fcrespo82:
Hi @tuvistavie, just updated the branch with the changes to `completions/asdf.bash`.
It should be able to merge without conflicts now.
Thank you.

Stratus3D:
Thanks @fcrespo82 ! Sorry it took so long to merge this.

Stratus3D:
It would be nice if we could pass any arbitrary name to this command. For example:

```
$ asdf which ruby
/home/trevor/.asdf/installs/ruby/2.3.0/bin/ruby
$ asdf which irb
No such plugin
```
It would be better if the second command returned `/home/trevor/.asdf/installs/ruby/2.3.0/bin/ruby`.

PR #182, opened 2017-03-25
Support system version

### LINKED ISSUE #55: Add system version
> Someday we should support "system" version as a way to use OS-installed stuff. Very useful for Python and other stuff.

Stratus3D:
This allows asdf to pass through to the system version of a give tool if it's version is set to `system` in the `.tool-versions` file. For example `ruby system` would result in `ruby` executing the version of Ruby on the path that is not managed by asdf. Or in the case of `ruby` not being present on the path, an error.

@tuvistavie did all the work for the feature, I just added the tests for it.

This should close #55

### DIFF
--- a/bin/private/asdf-exec
+++ b/bin/private/asdf-exec
@@ -18,33 +18,16 @@ fi
 IFS=' ' read -a versions <<< "$full_version"
 
 for version in "${versions[@]}"; do
-  IFS=':' read -a version_info <<< "$version"
+  install_path=$(find_install_path $plugin_name $version)
 
-  if [ "${version_info[0]}" = "ref" ]; then
-    install_type="${version_info[0]}"
-    version="${version_info[1]}"
-    install_path=$(get_install_path $plugin_name $install_type $version)
-  elif [ "${version_info[0]}" = "path" ]; then
-    # This is for people who have the local source already compiled
-    # Like those who work on the language, etc
-    # We'll allow specifying path:/foo/bar/project in .tool-versions
-    # And then use the binaries there
-    install_type="path"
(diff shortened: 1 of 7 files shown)


Stratus3D:
Looks like one test is failing on OSX. I'll test and fix it on Monday when I have access to a Mac.

Stratus3D:
Fixed the sed command that was incompatible with the sed version present on OSX. Tests are green now.

@tuvistavie feel free to merge if everything looks good. Once this is merged I'll tag release 0.3.0.

tuvistavie:
@Stratus3D Thank you very much for finishing working on this, I had this pending for quite a while. I am going to give it a try locally first, to be sure we did not forget anything :smile:

Stratus3D:
@tuvistavie I see you pushed a commit. Is there anything else that needs to be addressed?

tuvistavie:
I have been using it for a few days and everything seems to be working fine!
I also fixed my python plugin to support this, so let's merge! :smile:

Stratus3D:
I do the same thing. It's so nice being able to use any branch for day to day work.

nomasprime:
Hi, guys. This is exactly what I was looking for 👏 

Only thing is if the system, e.g., MacOS, has Python 2 and 3 I don't think `system` can handle this. Is there a way of adding something like `python 3.7.4 python3 python`?

nomasprime:
Would actually be nice if it could implicitly fallback to system to avoid additional wiring.

nomasprime:
Ah, you can pass `path:/src/elixir` 👌 Main issue sorted 🚀

Issue #181, opened 2017-03-24
[Proposal] Create a simple plugin repository for easier installation of plugins


doughsay:
I recently switched to oh-my-fish, and also wrote a simple plugin for it easily get started with asdf in fish: https://github.com/doughsay/omf-asdf

While working on my plugin, I learned of how oh-my-fish handles packages, and it's very simple and elegant.  I think asdf could maybe benefit from something similar.

In oh-my-fish, you just run `omf install <package>` to install new packages. It accomplishes this because they have a simple package repository here: https://github.com/oh-my-fish/packages-main. The installation command clones / updates that repo and checks for the given plugin name in the list of packages.  Packages are described by simple text files that looks like this:

```
type = plugin
repository = YOUR-PACKAGE-URL
maintainer = YOUR-NAME <YOUR-EMAIL>
description = YOUR-PACKAGE-DESCRIPTION
```

So, I'm proposing creating a new repository where we can create files for each plugin that exists right now, with the tool name being the filename and the repository field in the file being used when installing the plugin.  This would allow for a quick and easy:

```
asdf plugin-install elixir
```

Which would get the plugin repo url from the plugins repository.

Backwards compatibility with specifying a url should of course be maintained. This would just be an enhancement to the plugin-install command if no url is provided.

The benefits I see are mostly for more user adoption.  The easier it is for users to get started, the happier they'll be using asdf! :)

Thoughts?

Stratus3D:
That definitely would make it easier to use, but I'm not sure it would be worth the effort at this point. Installing plugins is usually a one time thing, so it's pretty minimal barrier to entry. It also makes it clear to the end user that Git is really what's doing all the work behind the scenes. 

@asdf-vm/maintainers what do you guys think?

Jwashton:
For what it's worth, I find myself installing plugins once a month or so. This could be because I'm setting up a new VM, or trying to learn a new language. Every time I have pull up the asdf readme so I can copy/pasta the appropriate link. I don't see this as a one-time thing, and I would absolutely appreciate this change.

I have only minimal bash experience, but I would be willing to contribute some time for this if that would help.

tuvistavie:
It would definitely be useful, but I do not really have time to work on this right now.
If somebody is willing to work on it and send a PR, I will be glad to review it.
If we are going to add this, I think we should also have a `plugin-list-all` command to show the user all the available plugins.

doughsay:
[...]

PR #183, opened 2017-03-26
Use a git repository for installing plugins.

doughsay:
This is my attempt at the proposal I opened: https://github.com/asdf-vm/asdf/issues/181

It's using a temporary repository under my account for now: https://github.com/doughsay/asdf-plugins.

Thoughts?  Feedback?

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -1,5 +1,5 @@
 MANAGE PLUGINS
-  asdf plugin-add <name> <git-url>     Add git repo as plugin
+  asdf plugin-add <name> [<git-url>]   Add a plugin
   asdf plugin-list                     List installed plugins
   asdf plugin-remove <name>            Remove plugin and package versions
   asdf plugin-update <name>            Update plugin
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -208,7 +208,7 @@ get_asdf_config_value() {
 }
 
 asdf_repository_url() {
-  echo "https://github.com/doughsay/asdf-plugins.git"
+  echo "https://github.com/asdf-vm/asdf-plugins.git"
 }
 
 initialize_or_update_repository() {
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -206,3 +206,29 @@ get_asdf_config_value() {
         get_asdf_config_value_from_file $default_config_path $key
     fi
 }
+
+asdf_repository_url() {
+  echo "https://github.com/doughsay/asdf-plugins.git"
+}
+
+initialize_or_update_repository() {
+  local repository_url=$(asdf_repository_url)
+  local repository_path=$(asdf_dir)/repository
+
+  if [ -d $repository_path ]; then
+    echo "updating plugin repository..."
+    (cd $repository_path && git pull)
+  else
+    echo "initializing plugin repository..."
+    git clone $repository_url $repository_path
+  fi
+}
+
+get_plugin_source_url() {
+  local plugin_name=$1
+  local plugin_config="$(asdf_dir)/repository/plugins/$plugin_name"
+
+  if [ -f $plugin_config ]; then
+    cat $plugin_config | grep "repository" | awk -F'=' '{print $2}' | sed 's/ //'
+  fi
+}
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -228,10 +228,9 @@ initialize_or_update_repository() {
 }
 
 get_plugin_source_url() {
-  local plugin_name
+  local plugin_name=$1
   local plugin_config
 
-  plugin_name=$1
   plugin_config="$(asdf_dir)/repository/plugins/$plugin_name"
 
 
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -217,7 +217,7 @@ initialize_or_update_repository() {
 
   if [ -d $repository_path ]; then
     echo "updating plugin repository..."
-    (cd $repository_path && git pull)
+    (cd $repository_path && git fetch && git reset --hard origin/master)
   else
     echo "initializing plugin repository..."
     git clone $repository_url $repository_path
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -212,23 +212,30 @@ asdf_repository_url() {
 }
 
 initialize_or_update_repository() {
-  local repository_url=$(asdf_repository_url)
-  local repository_path=$(asdf_dir)/repository
+  local repository_url
+  local repository_path
 
-  if [ -d $repository_path ]; then
+  repository_url=$(asdf_repository_url)
+  repository_path=$(asdf_dir)/repository
+
+  if [ -d "$repository_path" ]; then
     echo "updating plugin repository..."
-    (cd $repository_path && git fetch && git reset --hard origin/master)
+    (cd "$repository_path" && git fetch && git reset --hard origin/master)
   else
     echo "initializing plugin repository..."
-    git clone $repository_url $repository_path
+    git clone "$repository_url" "$repository_path"
   fi
 }
 
 get_plugin_source_url() {
-  local plugin_name=$1
-  local plugin_config="$(asdf_dir)/repository/plugins/$plugin_name"
+  local plugin_name
+  local plugin_config
 
-  if [ -f $plugin_config ]; then
-    cat $plugin_config | grep "repository" | awk -F'=' '{print $2}' | sed 's/ //'
+  plugin_name=$1
+  plugin_config="$(asdf_dir)/repository/plugins/$plugin_name"
+
+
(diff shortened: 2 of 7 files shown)


lib/utils.sh
```diff
@@ -206,3 +206,29 @@ get_asdf_config_value() {
...
+  echo "https://github.com/doughsay/asdf-plugins.git"
+}
+
+initialize_or_update_repository() {
+  local repository_url=$(asdf_repository_url)
+  local repository_path=$(asdf_dir)/repository
+
+  if [ -d $repository_path ]; then
+    echo "updating plugin repository..."
+    (cd $repository_path && git pull)
```
tuvistavie: For these this of update I think `git fetch && git reset --hard origin/master` is more predictable than `git pull`.

lib/utils.sh
```diff
@@ -206,3 +206,29 @@ get_asdf_config_value() {
...
+
+asdf_repository_url() {
+  echo "https://github.com/doughsay/asdf-plugins.git"
+}
+
+initialize_or_update_repository() {
+  local repository_url=$(asdf_repository_url)
+  local repository_path=$(asdf_dir)/repository
+
+  if [ -d $repository_path ]; then
```
tuvistavie: This will probably fail if `asdf_dir` contains spaces. Adding quotes should do the job.

tuvistavie:
Thank you very much for the PR, and sorry for the late reply.
I think this is very promising, and simple enough to be integrated into asdf.
I added a few comments.
Maybe it would be good if you could run [shellcheck](https://github.com/koalaman/shellcheck) on the code, and fix potential warnings.

doughsay:
Thanks for the review, I've addressed the feedback and run shellcheck.  There are many warnings produced by code I didn't touch; I've only removed all warnings from the code I added or modified.  I'd also like to add tests to make sure this works as expected, but I'm not familiar with bats...  How hard is it to test this, given that it relies on git?

tuvistavie:
Sorry for the delay and thanks for updating!
Sure, don't worry about the existing shellcheck warnings.

Our tests are not really pure unit tests, and some already rely on git and even on cloning remote repository, so I think it's fine to do the same thing for the parts of this feature that rely on git.

If you could just at least maybe add a test for the `asdf plugin-add nodejs`, I think we could maybe merge, and get some feedback before releasing the feature in the next version.

@asdf-vm/maintainers Thoughts?

vic:
Hm, interesting idea. Yes, please add some tests for this :), I like the idea in general, so I guess the list of available plugins would be inside the main asdf repo ? or shall we have another repo for that?. 

On my recent dockerized asdf project [asdf-alpine](https://github.com/vic/asdf-alpine) I'm holding the plugin url in a simple file like [this](https://github.com/vic/asdf-alpine/blob/elixir-1.5.0-rc.2-otp-20/elixir/plugin-repo) that gets read by a [tiny script](https://github.com/vic/asdf-alpine/blob/master/asdf-install-toolset).

doughsay:
@vic The idea would be to create a separate repo to list all available "official" plugins.  Contributors could open PRs to that repo to add new plugins.  Anyone wanting to add a not-yet-listed plugin could still use the `asdf plugin-add` command with a URL.  This PR relies on a repo that I created under my name to show the proof of concept, but I'd want an asdf maintainer to create `asdf-vm/asdf-plugins` before releasing this.  I haven't gotten to adding the tests because I've never dealt with bats before... I'll take a stab at it this morning to see if it's not too hard.

doughsay:
@vic @tuvistavie added a simple test for plugin-add with no URL provided.  Is that good enough?  It does rely on actually cloning the plugins repo and installing a real plugin, rather than a dummy plugin like the other tests.

vic:
@doughsay nice! Code looks ok :) no problem on actually clonning the asdf-repository and the real elixir plugin afterwards. 

I'm 👍  on this. Would you mind adding the current list of repo urls to the asdf-repository? Also please update the `README.md` file to mention the repository that the plugin-url is now optional. And perhaps also at `creating-plugins.md` as most plugin authors will need to read that.

We could move your repository to the asdf-vm org just before releasing this.

vic:
Oh forgot to mention, just add another test for the case where I dont provide the url and the plugin is not registered on the repository. It should just show the error message and exit 1. :)

vic:
Just wondering, once this gets merged, would it be possible to tell people, to just install asdf and afterwards go directly with `asdf list-all elixir` or even to `asdf install elixir 1.4.5` without having to *be explicit* about `plugin-add elixir` ? I mean, once we have the plugins repo, `plugin-add` is looks to me like only really useful for non-official plugins.

doughsay:
@vic I've added all the current plugins to the plugins repo here: https://github.com/doughsay/asdf-plugins. (I left out the `link` plugin though, since I see it's not really meant to be added on its own.)

I've also added the extra test and documentation updates you suggested.

As for your other comments, it does make sense that all plugin commands could automatically attempt to add a plugin if the plugin doesn't yet exist.  Consider typos though, would you want `asdf install elxr 1.4.5` to update the plugins repo and check if there's an elixr plugin before erroring out?  It's pretty fast, so I don't think it would be a big deal; just something to consider.

vic:
@doughsay awesome! Thank you very much. I guess this PR is close to ready.

The plugins repo looks nice, no problem on leaving `link` out. 

I'd say if only the `asdf install` command was aware of the plugins-repo it'd be an awesome first-time experience that just after downloading asdf, to be able to install any officially supported tool without further ado. 

So I guess it's up to you, if you'd like to implement this, for `asdf install foo 1.0` checkout the plugins-repo and look for foo in it. Maybe we could work on this in another Issue/PR. Also other thing we might consider for other PR is being able to list all known tools from the plugin-repo, perhaps `asdf plugin-list --remote`

Anyways, work on this branch looks like done to me, I've downloaded @doughsay asdf to give it a try and the new `plugin-add` feature works fine. 

@doughsay would you mind transferring the asdf-plugins repo to the asdf org? and then just update the url in the code? I guess that would be all for being able to merge this branch. Thank you :)  

@tuvistavie @asdf-vm/maintainers anything I might be missing?

HashNuke:
People, this is awesome work ~! ❤️ 💚 💙

doughsay:
Thanks a lot everyone, for making this awesome tool and for letting me contribute.  It's been fun!

@vic I tried transferring the ownership of the repository but apparently it won't let me unless I have permission to create new repositories under the organization.  If you'd like, you could just create a new empty repository and I'll copy-paste everything I've done into a PR for the new repo.

Also, I like your ideas for the `asdf install` and `asdf plugin-list --remote` commands; I think we should finish this PR as is, but I would be happy to try implementing those ideas next.

As soon as a new repo exists where I can upload the plugin configs, I'll change the URL in this PR.

Thanks!

doughsay:
Ah, I read the fine print a bit more carefully: "Transfer this repository to another user or to an organization where you have the ability to create repositories."

That means I could transfer the repo to a user easily, just not an org.  Would you like me to transfer it to you @vic, and then you can transfer to the org?

vic:
@doughsay  here's the new repo :) https://github.com/asdf-vm/asdf-plugins

tuvistavie:
Thanks for working on this, it looks great and will be super useful as we are having more and more plugins.
I think we can merge and start to try it out.
@vic @doughsay I agree it would be nice if we could check if there is an available plugin when we do not find it locally. `asdf plugin-list --remote` will be useful too. I will open a new issue for both of these features.

doughsay:
URL has been updated, should be ready to go! 👍

tuvistavie:
Great, thank you very much!!
Let's give it a try :smile: 

I opened two new tickets for the new features discussed here:
https://github.com/asdf-vm/asdf/issues/210
https://github.com/asdf-vm/asdf/issues/211

help.txt
```diff
@@ -1,5 +1,5 @@
 MANAGE PLUGINS
-  asdf plugin-add <name> <git-url>     Add git repo as plugin
+  asdf plugin-add <name> [<git-url>]   Add a plugin
```
Stratus3D: I think the help message here should be a little more detailed. Perhaps something like `Add a plugin from the plugin repo OR add a Git repo as a plugin by specifying name and repo url`?

lib/utils.sh
```diff
@@ -206,3 +206,35 @@ get_asdf_config_value() {
         get_asdf_config_value_from_file $default_config_path $key
     fi
 }
+
+asdf_repository_url() {
+  echo "https://github.com/asdf-vm/asdf-plugins.git"
+}
```
Stratus3D: Maybe move this to the top of the file under the `asdf_version` function since it's basically a constant.

Stratus3D:
No need to address those two comments I left during the review. I went and updated master with those changes. Thanks @doughsay !

Issue #92, opened 2016-09-08
asdf.sh script bug with bash OSX sessions


gregors:
on OSX El Capitan the asdf.sh logic to find the root directory doesn't work correctly since the terminal started outputting  "/Users/gregoryostermayr/.asdf Saving session... ...saving history...truncating history files... ...completed."

Stratus3D:
@gregors what version of asdf are you running? You can find out by either running `asdf --version` or by looking at the second line in ~/.asdf/lib/utils.sh

gregors:
"0.1" as per readme

Stratus3D:
I'm on Yosemite (10.10.5), and I am unable to replicate this issue. I know there is an bug in version 0.1 when it's used in a directory name that contains spaces, but it doesn't look like that is the same issue as yours. Are you using Zsh or Bash? Do you have your .bashrc/.zshrc posted somewhere so I could take a look at them?

gregors:
@Stratus3D you won't see this behavior on Yosemite only El Capitan - tries to be nice by adding a "global history"  aka Bash Sessions. see http://superuser.com/questions/975678/strange-output-from-terminal-exit-command-is-this-a-virus

gregors:
I'll take a stab at creating a fix and run it by you

Stratus3D:
Great! I don't have access to a machine running El Capitan right now so this is a huge help.

I'm hoping to tag the 0.2.0 release sometime in next couple of weeks, but I'll definitely want to get your fix merged before the release. If you have any questions about the codebase feel free to ask questions.

philipgiuliani:
I have the same problem on El Capitan :(

HashNuke:
@gregors did you find a fix? I'm not able to reproduce this on latest master with El Capitan.

philipgiuliani:
I fixed this bug by creating a file called `.bash_sessions_disable`. My sessions are still saved, so im not sure how this works.

@HashNuke could you check if you maybe have this file?

`touch ~/.bash_sessions_disable`

HashNuke:
@philipgiuliani I don't have the file. But I just came across the exact same issue on a colleague's computer last week and fixed it with some googling. I had to create that file.

I'm guessing the fix would have to be setting `SHELL_SESSION_HISTORY=0` whenever we run bash internally.

gregors:
@Stratus3D feedback appreciated

gregors:
[...]

PR #187, opened 2017-04-07
fix osx bash session logging

gregors:
this fixes the bug i reported in issue #92 I'm continuing to see when I have osx bash session history turned on

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -9,8 +9,10 @@ fi
 export ASDF_DIR
 ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
 
-# remove osx bash session logging output
+if [ -n "$BASH_VERSION" ]; then
+# remove osx bash session logging output if exists
 ASDF_DIR=${ASDF_DIR//[[:space:]]*Saving[[:alpha:][:blank:]]*completed./}
+fi
 
 export PATH="${ASDF_DIR}/bin:${ASDF_DIR}/shims:$PATH"
 
--- a/asdf.sh
+++ b/asdf.sh
@@ -8,6 +8,10 @@ fi
 
 export ASDF_DIR
 ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
+
+# remove osx bash session logging output
+ASDF_DIR=${ASDF_DIR//[[:space:]]*Saving[[:alpha:][:blank:]]*completed./}
+
 export PATH="${ASDF_DIR}/bin:${ASDF_DIR}/shims:$PATH"
 
 if [ -n "$ZSH_VERSION" ]; then


tuvistavie:
Thanks for the PR, and sorry for the delay.
What is this doing exactly? I am not familiar with this syntax.

Stratus3D:
I understand what the change is doing but I don't understand why this is necessary. Is it due to the fact we are capturing the output from a subshell that includes Bash session logging output? If that is the case this PR will likely not fix all the issues in asdf due to since we capture output from subshells in many places.

gregors:
@Stratus3D @tuvistavie yes - there is additional output due to Bash session logging on the latest version of OSX - but it's inline. The current version of ASDF still displays the error 

if I echo out my ASDF_DIR I get the following:
```
/Users/gregoryostermayr/.asdf Saving session... ...copying shared history... ...saving history... ...completed.
```

This is what get's inserted into the path. And obviously won't work. I think that most OSX experienced devs (at least in the Carbon Five office) are using Z Shell and don't see this at all.

The code is a regex with replacement - if the "Saving session.... " string is present in ASDF_DIR capture that string and remove it. It's pretty specific with the spaces and blanks so the likelihood that it would match something else is low.

Stratus3D:
@gregors I still don't how this one change fixes everything. If it's due to a subshell you should be encountering this issue in several other places. We use subshells a lot. Are you sure all the tests pass on your machine with just this one fix?

gregors:
@Stratus3D - I don't think the text gets added from a subshell (or if it does it's not from std out). This fixes it by:
1. recognizing that additional offending text exists
2. removing the offending text

such that:
```
/Users/gregoryostermayr/.asdf Saving session... ...copying shared history... ...saving history... ...completed.
```

gets changed to

```
/Users/gregoryostermayr/.asdf
```

Which is what ASDF_DIR should be set to.

I'll double check regarding additional tests -but regardless it still passes for the overall ci suite and right now this issue makes asdf unusable for any osx bash users who don't know how to manually modify asdf.sh or disable the bash sessions feature. So if there are other issues I'm definitely not seeing them as I do use ASDF on a daily basis.

gregors:
@Stratus3D I'm going to try spend some time to see if I can actually deduce how that text comes back and if we can somehow prevent that. I take it that prevention of the text coming back is a better implementation than removing it once it does, but the end goal is setting ASDF_DIR to the proper path.

Stratus3D:
@gregors what happens if you change the line that sets `ASDF_DIR` to this?

```
ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null && pwd)"
```

Does the behavior change at all?

Stratus3D:
Also, did you try setting `SHELL_SESSION_HISTORY=0`? @HashNuke mentioned something like that might fix it too.

Also, can you determine where the text is coming from? Is it STDOUT or STDERR?

gregors:
@Stratus3D setting SHELL_SESSION_HISTORY is not a fix - I want my session history. And even if it was a fix - it's a fix that is not beginner friendly. It requires additional knowledge -  as opposed to actually just fixing the issue inline.

The other logic you provided doesn't fix the issue.

Some friends also interested in fixing this issue pointed me here:
` /etc/bashrc_Apple_Terminal` seems to set up the output via a trap
https://unix.stackexchange.com/questions/282722/is-trap-inherited-by-a-subshell

gregors:
I wonder if there's a way to '`untrap` before the subshell returns ala https://stackoverflow.com/questions/31201572/how-to-untrap-after-a-trap-command

Stratus3D:
@gregors sorry, my point with `SHELL_SESSION_HISTORY` is that we could set it inside asdf so all asdf subshells would inherit it. You do not want commands run by asdf recorded as shell history anyway.

gregors:
@Stratus3D oh I see what you're saying

Stratus3D:
@gregors so another thing you could try to pinpoint the issue is split the expression up into multiple lines and see which one contains the logging string:

```
RELATIVE_ASDF_DIR=$(dirname "$current_script_path")
echo $RELATIVE_ASDF_DIR
ASDF_DIR="$(cd "$RELATIVE_ASDF_DIR" &> /dev/null && pwd)"
echo $ASDF_DIR
```

You could also try moving some things out of the subshell, for example:

```
cd "$RELATIVE_ASDF_DIR" &> /dev/null
ASDF_DIR="$(pwd)"
echo $ASDF_DIR
```

I would do this myself if I had access to an OSX machine with this issue, but unfortunately I don't and I'm not able to upgrade right now.

gregors:
@Stratus3D  `12 RELATIVE_ASDF_DIR=$(dirname "$current_script_path")` results with the proper full working path without the trap text

`ASDF_DIR="$(cd "$RELATIVE_ASDF_DIR" &> /dev/null && pwd)"`
Is what causes the trap text to appear ^^^^^^^^

Stratus3D:
Excellent! What happens if you do this?

```
RELATIVE_ASDF_DIR=$(dirname "$current_script_path")
ASDF_DIR="$(cd "$RELATIVE_ASDF_DIR" &> /dev/null && pwd)"
cd "$RELATIVE_ASDF_DIR" &> /dev/null
ASDF_DIR="$(pwd)"
```

Thanks for taking the time to test this. I looked into setting up builds on multiple OSX versions, but it doesn't seem like Travis CI supports that.

gregors:
@Stratus3D - splitting the logic like that results in no trap output 👍

Stratus3D:
Awesome! I'll merge the PR once you update it with those changes.

gregors:
@Stratus3D - that unfortunately introduces a massive bug - when you do open a sub shell or a new tab your starting directory is no longer `~` but in the directory of `asdf.sh`

Stratus3D:
Shoot, I forgot the context this code was in. Changing directories here is not a good idea, even if we change back, because that could affect other things the user is doing.

I'm not really sure how else do to this. We have similar logic in subshells other places in the code and it works fine. I'm not really sure what is special here. Until I know what trips up the shell history it's hard to make any other suggestions. Perhaps try this:

```
RELATIVE_ASDF_DIR=$(dirname "$current_script_path")
echo $RELATIVE_ASDF_DIR
ASDF_DIR="$(cd "$RELATIVE_ASDF_DIR"; pwd)"
echo $ASDF_DIR
```

Dropping the `&> /dev/null &&` might change things.

gregors:
@Stratus3D - no joy - now it appears with the junk

 - though RELATIVE_ASDF on my system does that have right value - though I can't confirm it always will be.

gregors:
I'm still pushing my for change (obviously) that does handle any bad input

- but another approach is to do what npm does and place an if statement checking for the ASDF before trying to set it. If you check for existence before running then at least bash users (on OSX) could set it manually once without having to patch your code on every update.

The problem currently is it runs for every new shell - so even if I set it correctly by hand - it gets overwritten (without my clean up logic)

gregors:
@Stratus3D I've added a check that would limit the check to only bash users. Did you think about my suggestion of checking for an explicit env file prior to any script running? Would you be open to that change in another PR?

rachel-carvalho:
This is also affecting me, is there anything I can do to help out to get this merged?

Stratus3D:
I believe this issue has been fixed in the latest release, v0.5.1. Please upgrade by running `asdf update`.

PR #188, opened 2017-04-08
Avoid bash error when no version is set

henrik:
Running "asdf current ruby" before installing any Rubies gave me:

    /Users/henrik/.asdf/lib/utils.sh: line 73: [: !=: unary operator expected
    No version set for ruby


Stratus3D:
Thanks @henrik !

PR #190, opened 2017-04-16
Query git for the current version tag (Fixes #189)

### LINKED ISSUE #189: asdf still shows version at 0.2.1 after update to 0.3.0
I just ran `asdf update` and got the following output:

`HEAD is now at d3a66d7... Merge pull request #173 from asdf-vm/prepare-for-0.3.0
Updated asdf to release v0.3.0`

After restarting the shell, `asdf --version` still displays 0.2.1

Jwashton:
Previously the version number was hard coded into `utils.sh`. This can be easy to miss when bumping version numbers elsewhere. This pull request asks git what the most recent tag in the tree is, reporting that answer instead of a static version number. The version number also contains what I suspect are

1. the number of commits since that version tag and
2. the mini-hash for the current commit.

```bash
$ asdf --version
v0.3.0-5-gbaad7dd
```

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -4,7 +4,11 @@ GREP_OPTIONS="--color=never"
 GREP_COLORS=
 
 asdf_version() {
-  echo "0.2.1"
+  # Move to the asdf repo, then report the current tag
+  (
+  cd $(asdf_dir)
+  git describe --tags
+  )
 }
 
 asdf_dir() {


tuvistavie:
I am not sure if this is the right way to fix this, as this will output something incorrect for non-tagged commits.
If somebody uses `master` branch, it would be better than `version` returns some information about it (maybe something `0.4.0-dev`).

Any idea how we could automatically check the tag and code version actually match? By the time the CI server runs the tag would already have been pushed, so I cannot think of a good way to make a precheck.

/cc @Stratus3D

Jwashton:
Non-tagged commits will report the most recent tag + the number of commits since that tag + plus the hash of the current commit. I'm very willing to believe that this is the wrong thing to show to the users, but I'm not sure I understand how this information is "incorrect".

tuvistavie:
Sorry, I confused the behavior of `git describe --tags` with the one of `git describe --abbrev=0 --tags`, which is why I thought it would give an incorrect output.
I think it is much better than the current hard-coded version we have, thank you!

@Stratus3D what do you think?

Stratus3D:
I think this is perfect! At first I was thinking it would be safer to stick with hardcoded versions. I think in practice this is better though, since users will be able to see the exact commit they are running, which should help when diagnosing issues.

PR #192, opened 2017-04-25
Added missing local and global to bash completion

sphinxc0re:


tuvistavie:
Great, thank you very much! :smile:

PR #204, opened 2017-07-03
Comment out 'set current version' example in readme

DavidOliver:
To match other command examples.


Stratus3D:
Thanks @DavidOliver !

PR #212, opened 2017-07-26
Add asdf-plugin-version metadata to shims

### LINKED ISSUE #200: reshim not working
I removed some gems and `reshim` won't remove the gem references:
```
$ which ruby
/usr/local/opt/asdf/shims/ruby
$ gem list

*** LOCAL GEMS ***

bigdecimal (default: 1.3.0)
bundler (1.15.1)
io-console (default: 0.4.6)
json (default: 2.0.2)
openssl (default: 2.0.3)
psych (default: 2.2.2)
rdoc (default: 5.0.0)

$ which kitchen
/usr/local/opt/asdf/shims/kitchen

$ cat $(which kitchen)
#!/usr/bin/env bash
# asdf-plugin: ruby
exec /usr/local/opt/asdf/libexec/private/asdf-exec ruby bin/kitchen "$@"
```
I would like `kitchen` to go away in this instance.

tuvistavie:
This adds `asdf-plugin-version` to all generated shims.
This metadata is used to remove the shim when necessary.

I think this should handle most cases we need:

* Remove a shim when removing a version using `asdf uninstall`
* Remove a shim which has been uninstalled using the language package manager when calling `asdf reshim $PLUGIN`

Of course this will only remove a shim in the case the binary exists in no other versions of the plugin, so for example if we install a binary `dummy` using versions `1.0` and `1.1`
of a plugin, and then remove version `1.0`, the shim will
correctly stay. However, if the shim was only installed
in version `1.0` and we remove this version, the shim
will be removed.

Please take a look and let me know if I missed something.

This should fix #200

### DIFF
--- a/lib/commands/reshim.sh
+++ b/lib/commands/reshim.sh
@@ -51,7 +51,7 @@ write_shim_script() {
     cp "$plugin_shims_path/$executable_name" "$shim_path"
   elif [ -f "$shim_path" ]; then
     if ! grep "# asdf-plugin-version: $version" "$shim_path" > /dev/null; then
-      sed "$shim_path" -i'' -e "s/\(asdf-plugin: $plugin_name\)/\1\n# asdf-plugin-version: $version/"
+     sed -i'' -e "s/\(asdf-plugin: $plugin_name\)/\1\\"$'\n'"# asdf-plugin-version: $version/" "$shim_path"
     fi
   else
     cat <<EOF > "$shim_path"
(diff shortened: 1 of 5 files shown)


tuvistavie:
@asdf-vm/maintainers
I don't have access to a mac, so help debugging this on OSX would be very welcome!

vic:
@tuvistavie just fixed some issues for making the tests pass on OS X. 

Everything looks fine, I'm merging this.

tuvistavie:
Thank you very much!

lib/commands/reshim.sh
```diff
@@ -93,29 +120,106 @@ generate_shims_for_version() {
...
+  IFS=' ' read -a all_bin_paths <<< "$space_separated_list_of_bin_paths"
+
+
+  for bin_path in "${all_bin_paths[@]}"; do
+    if [ -x "$install_path/$bin_path/$shim_name" ]; then
+      return 0
+    fi
+  done
+  return 1
+}
```
Stratus3D: I might be missing something here, but would `plugin_bin_still_exists` be a better name? This is checking for the presence of a binary in the plugin installation directory right?

Issue #194, opened 2017-05-05
Feature request: `asdf env` command that prints out the version info for every tool managed by asdf


Stratus3D:
I'm thinking it might be nice to have a command that can print out all of the information for each tool. We the command would output one line per tool and on each line we would print the tool name, version, place the version was set, and whether or not the version is installed. For example:

```
ruby 2.3.1 (set by ~/project/.ruby-version) installed
erlang 19.3 (set by ~/.tool-versions) not installed
```

This would make it easy to see the environment at a glance. It would also make it easy to see if the tool versions needed for a project you have just cloned are already installed on your system. Right now you can run `asdf install` and see what happens, but you can see what needs to be installed before actually installing the versions.

The `asdf where` and `asdf current` commands already provide similar output, but you must specify the tool name, so they are more limited in that you only see one tool at a time. You can't see all the tools until you run `asdf plugin-list`.

Stratus3D:
@asdf-vm/maintainers what do you think?

tuvistavie:
I agree being able to output all the versions at once could be helpful.
However, instead of increasing the number of commands, what would you think about adding a flag to `asdf current`, for example `asdf current --all` could provide this output.

Stratus3D:
Good idea. I've noticed we have quite a few similar commands, `current`, `where`, and soon, `which`.

I could add the `installed|not installed` field to the existing `current` command then just implement the `--all` flag as a loop over the existing function.

tuvistavie:
Solved by #219

PR #219, opened 2017-08-19
List all installed plugins when no argument is passed to current

tuvistavie:
`asdf current` currently needs the plugin name to show the version.
With the increasing number of plugins, I think it would be nice to have
a list of all the version currently in use, so I added this functionality to
`current`. For example, here is the output I now get on my machine:

```
elixir 1.5.1 (set by /home/daniel/.tool-versions)
erlang 20.0 (set by /home/daniel/.tool-versions)
golang 1.8 (set by /home/daniel/.tool-versions)
nodejs 6.10.2 (set by /home/daniel/.tool-versions)
python 3.6.1 2.7.13 system (set by /home/daniel/.tool-versions)
ruby 2.4.1 (set by /home/daniel/.tool-versions)
```

### DIFF
--- a/lib/commands/current.sh
+++ b/lib/commands/current.sh
@@ -1,4 +1,4 @@
-current_command() {
+plugin_current_command() {
   local plugin_name=$1
 
   check_if_plugin_exists $plugin_name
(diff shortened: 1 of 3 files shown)


Stratus3D:
Woot woot! Nice work with the new test helper functions!

Stratus3D:
This resolves #194 right?

tuvistavie:
@Stratus3D Thank you for the review. Yes, this resolves #194.

Issue #198, opened 2017-06-03
`--version` is doing something odd


vicendominguez:
Hello guys...

Working on the [asdf brew formula](https://github.com/Homebrew/homebrew-core/pull/14058)  I saw the next points:

* when you **clone** or **download** the `v0.3.0` tag then the `--version` argument always says: `0.2.1`
* when you download `master` from a **zip file** (`wget https://github.com/asdf-vm/asdf/archive/master.zip`) then the ``--version`` ouput is not there:

```
$ asdf --version
fatal: Not a git repository (or any of the parent directories): .git
```

Perhaps [this approach](https://github.com/asdf-vm/asdf/blob/master/lib/utils.sh#L6-L12) is not the better one when you are downloading the zip file :

```
asdf_version() {
  # Move to the asdf repo, then report the current tag
  (
  cd $(asdf_dir)
  git describe --tags
  )
}
```
Zip file downloading let us to download a concrete tag version.....so, any recommendation?

Stratus3D:
Thanks for reporting this. The version discrepancy was my mistake, I forgot to update the hardcoded version in utils.sh when I tagged. After which we moved to using `git describe --tags`. We felt it was easier and assumed everyone would have installed asdf via `git clone`. It may be better to go back to hardcoded values, and I can write a quick shell script to increment the versions (readme and utils.sh) and do the tag all at once.

Another issue here is likely updates - `asdf update` does a `git pull`, so if we want to support asdf without it's git repo we will have to change that function as well.

vicendominguez:
Hello Trevor, thx for answering.... I understand your point. 

I only would like to add some information here:

Updating action will not work via `brew` due to two good reasons:
 * philosophy: you are using `brew` to manage it and the updating makes sense with `brew` right? . Anyway this doesn't avoid the updating action but the next point is harder.
 * code: to achieve some formula rules I had to move `bin/private` folder while `brew` installs `asdf`. The native  "`update`" could break `asdf` then. 😢

Stratus3D:
@vicendominguez It would be nice if you didn't have to modify asdf to get it working with brew, but maybe we don't have a choice. Is there something we can do to make the brew formula simpler?

vicendominguez:
oh @Stratus3D 👋 , I would like it and I tried it but, as you see in the [PR](https://github.com/Homebrew/homebrew-core/pull/14058), the CI and the core manager didn't agree with that....

[...]

PR #220, opened 2017-08-22
Release script

Stratus3D:
# Summary

Add a release.sh that automates the work needed to release a new version of asdf. The script ensures the version doesn't already exist and also makes sure there aren't any uncommitted changes in the working directory. The script updates the files that contain the version, then commits the changes, and then tags a new version. 

All the maintainer needs to do to release a new version is:

1.  Run the script on the master branch - `./release.sh v0.3.1`
2. Then push the branch to GitHub - `git push origin master`

This will address part of #198 

There is one TODO comment in the script. That will be addressed in another PR that will resolve the rest of #198.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+
+# Unoffical Bash "strict mode"
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -euo pipefail
+#ORIGINAL_IFS=$IFS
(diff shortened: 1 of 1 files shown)


tuvistavie:
LGTM, thanks for adding this.

For the version, I think we could have a `VERSION` file at the top level of the repository and have `util.sh` read from it.
Then the release script would simply need to `echo "new_version" > VERSION` to update the version.

Stratus3D:
Good idea with the VERSION file. I will do that.

Issue #157, opened 2017-02-12
Please use shellcheck


ypid:
Programming in Shell can be difficult with all those things to think about. Kind of similar to Perl in this regard. Anyway. Please consider using [shellcheck](https://github.com/koalaman/shellcheck) to catch possible issues before they become problems.

PR #223, opened 2017-09-01
Shellcheck

### LINKED ISSUE #202: Shims not working when .tool-versions path contains spaces
If the `.tool-versions` file is present in a path that contains spaces and declares at least one versions, using a shim results in a series of `No such file or directory` grep errors and a failure to use the versions specified in that `.tool-versions` file.

Steps to reproduce (using the erlang plugin as an example):
1. Install asdf and asdf-erlang.
2. Create a directory with a space in the name. (If you really want to see the extent of this bug, give the directory a name with a star like `\*\ my\ directory`.)
3. Create `.tool-versions` in that directory with the entry `erlang 20.0`.
4. Run `erl` from that directory.

jonknapp:
# Summary

This brings [shellcheck](https://github.com/koalaman/shellcheck) support to the application by allowing linting of the project's source through the `./lint.sh` command. Linting is also added to Travis.

Fixes: Hopefully, this will fix #202 (which may not be fixed) as well as similar issues

## (Historical) Other Information

This is not ready in its current form as I've only updated issues found outside of the `lib` folder for the project. I'm also not 100% confident in my Bash scripting skills or in my familiarity with the project source, so I wanted to start a PR early on the subject in case someone is interested in assisting (or picking it up from me). I've been relying on the test suite to verify if functionality is still working after correcting suggested linting errors.

There will also be linting issues that I'm unsure how to solve and may add comments to ignore them in shellcheck. Feel free to make comments at any point if a different approach is desired to my changes or my ignored checks.

Solves #157


vic:
Hey, 

looks like this PR has a lot of changes, could you please rebase on current master?

jonknapp:
Rebase complete.

I have still not looked at the `lib/commands/*` files, but everything else is at least passing shellcheck or was ignored if I didn't know how to proceed.

As I originally stated, I don't have full confidence in the changes but I felt it was better to do some of the legwork to get the project passing with shellcheck. I love the idea of a universal version manager and wanted to contribute in some way. Feel free to use, modify, or throw away my work so that it aligns with the project's goals.

vic:
@jonknapp Thanks so much for your work on adding shellcheck! I've just shellcheckedd all of commands/*.sh and the test suite is green, so just let do a few manual tests (plugin-test) and I guess this would be ready for merge. 

Thanks again for working on this!

vic:
@asdf-vm/maintainers I've reviewed this PR, made sure the tests are passing and tested locally using this branch (manually installing plugins, shims worked and `asdf plugin-test` also worked).

So it would be great if anyone could also give a quick look at the PR just to be sure I'm not missing anything. I'd like to merge this if everything is ok.

tuvistavie:
Thank you very much and sorry for the delay.
It seems in the meanwhile `lib/utils.sh` has conflicted, could you fix the conflict and push again, please? Thanks.

jonknapp:
Things are all fixed up now. There was one conflict and the linter was getting caught up on text output for a help description (which I fixed by disabling the check for that instance).

I merged in `origin/master` since there was some activity since this PR started.

tuvistavie:
Thank you very much for this.
A last thing before merging, we do not need the shebang in files which
are not meant to be run as executables.
Could you please remove it from everything under lib and run `shellcheck` with `-s bash` option, which should avoid producing the warning.
https://github.com/koalaman/shellcheck/wiki/SC2148

Thanks!

jonknapp:
done

lib/commands/plugin-test.sh
```diff
@@ -1,69 +1,68 @@
...
+    if ! (asdf reshim "$plugin_name"); then
         fail_test "could not reshim plugin"
     fi
 
     if [ -n "$plugin_command" ]; then
-        (PATH="$ASDF_DIR/bin":"$ASDF_DIR/shims":$PATH eval "$plugin_command")
+        (PATH="$ASDF_DIR/bin:$ASDF_DIR/shims:$PATH" eval "$plugin_command")
+        local exit_code
         exit_code=$?
-        if [ $exit_code -ne 0 ]; then
```
Stratus3D: Did shellcheck suggest this? I was thinking `-ne` was a more widely supported condition.

lib/utils.sh
```diff
@@ -24,37 +27,40 @@ get_install_path() {
...
+  mkdir -p "$(asdf_dir)/installs/${plugin}"
 
-  if [ $install_type = "version" ]
+  if [ "$install_type" = "version" ]
   then
-    echo $(asdf_dir)/installs/${plugin}/${version}
+    echo "$(asdf_dir)"/installs/"${plugin}"/"${version}"
   else
-    echo $(asdf_dir)/installs/${plugin}/${install_type}-${version}
+    echo "$(asdf_dir)"/installs/"${plugin}"/"${install_type}"-"${version}"
```
Stratus3D: In both of these cases double quotes around the entire path should suffice.

lib/utils.sh
```diff
@@ -64,38 +70,41 @@ check_if_version_exists() {
...
-  if [ "$version" != "system" ] && [ ! -d $install_path ]; then
+  if [ "$version" != "system" ] && [ ! -d "$install_path" ]; then
     display_error "version $version is not installed for $plugin_name"
     exit 1
   fi
 }
 
 get_plugin_path() {
-  echo $(asdf_dir)/plugins/$1
+  echo "$(asdf_dir)"/plugins/"$1"
```
Stratus3D: Again, it's easier if you wrap the whole path in a single set of double quotes.

lib/commands/install.sh
```diff
@@ -20,33 +20,36 @@ install_command() {
...
 get_concurrency() {
   if which nproc > /dev/null 2>&1; then
-    echo $(nproc)
+    nproc
   elif which sysctl > /dev/null 2>&1 && sysctl hw.ncpu > /dev/null 2>&1; then
-    echo $(sysctl -n hw.ncpu)
+    sysctl -n hw.ncpu
   elif [ -f /proc/cpuinfo ]; then
-    echo $(grep -c processor /proc/cpuinfo)
+    grep -c processor /proc/cpuinfo
```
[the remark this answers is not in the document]
jonknapp: I'm not super well versed in bash scripting, so please feel free to let me know if something I suggest is incorrect. That said, I'm not sure why the `echo` is needed if the output of the `grep` command is sent to stdout/err already. It was a recommendation from `shellcheck` which may or may not apply here.

For the rest of comments, feel free to run this with shellcheck yourself to see the error output. I put a `lint.sh` file in the root of the project that lints the project files.

lib/commands/plugin-test.sh
jonknapp: Yea, I didn't make any changes that were not suggested by shellcheck. If they are syntactically correct I would tend to error on the side of shellcheck and there may be more info about why it was suggested on the rule definition doc page.

lib/utils.sh
jonknapp: Sounds good. There may be more instances of this as well. I'll poke around before committing the code.

lib/utils.sh
jonknapp: ditto.

jonknapp:
Made the corrections for unnecessary string quoting and added comments to the other feedback items. Feel free to fork and make additional changes (I believe asdf maintainers can change my Github branch) or let me know what else should be touched up.

Stratus3D:
Going to merge now. I'll try out locally later this week.

PR #224, opened 2017-09-03
Allow spaces in .tool_versions file path

aj-foster:
# Summary

This helps to address (though may not completely fix) issue #202. By adding quotes around two usages of a `$file_path` variable, the utility functions support parsing a legacy or `.tools_version` file located in a path with spaces.


tuvistavie:
Great, thank you very much.

PR #227, opened 2017-09-12
Brew upgrade fix

Stratus3D:
# Summary

This PR addresses a couple things that will make asdf integrate better with brew:

1. Moved back to a hardcoded version. Now the version is hardcoded in `VERSION`, this allows the `--version` flag to work, even if asdf lacks it's repository (which is the case with brew).
2. Add check for presence of a `asdf_updates_disable` file before attempting to upgrade asdf. If the file is present asdf will not be upgraded and a message is displayed. Brew will simply create this `asdf_updates_disable` file to disable that functionality so that users will use brew instead.

Refer to #198 for the reasoning behind this.

Fixes: #198 

Once this is approved and merged I will tag v0.3.1 (or perhaps v0.4.0 if you all would prefer that).


Stratus3D:
Agreed. I guess there probably are breaking changes in the interface, but I haven't actually checked. I'll bump the version to 0.4.0 and then merge and tag when I get a chance.

Issue #201, opened 2017-06-13
Opt-in for XDG base directory support


franklinyu:
This is a feature request. If `$ASDFRC` is set, use `$ASDFRC` instead of `$HOME/.asdfrc`. Do the same to `$HOME/.tool-version`.

Alternatively, if `$ASDF_CONF_DIR` is set, use `$ASDF_CONF_DIR/config` as `.asdfrc`, and `$ASDF_CONF_DIR/tool-version` as `.tool-version`.

Reference: [Arch Linux Wiki](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support)

hugopeixoto:
Right now it seems that you're able to specify the location of `asdfrc` through a variable named `$AZDF_CONFIG_FILE`. Not sure why the `Z` in the name. Could/should this be renamed to `$ASDF_CONFIG_FILE` or `$ASDFRC`? This is an undocumented feature, but renaming might break existing usages.

I was able to patch asdf to use a variable `$ASDF_TOOL_VERSIONS`. It doesn't affect in any way legacy files lookup, but I don't think it needs to. I created a PR, but I still need to add some tests.

PR #228, opened 2017-09-14
Add support for $ASDF_DEFAULT_TOOL_VERSIONS_FILENAME

hugopeixoto:
# Summary

Instead of just looking in `$HOME/.tool-versions`, allow the default
tool-versions file location to be specified through the variable
`$ASDF_TOOL_VERSIONS_FILENAME`.

Fixes: #201 

## Other Information

I still need to add tests, but in the meantime some feedback might be useful.


hugopeixoto:
I added some tests now. This should be ready, I'll be using it locally for a while to see if anything breaks.

I am checking for `$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` after checking the `$HOME` directory because there might be legacy version files in there, and this variable only controls the `.tool-versions` filename. This means that the file `$HOME/.tool-versions` will override `$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` file. Not sure if this is ideal.

vic:
Hey @hugopeixoto, could you resolve the merge conflicts and update the PR ? thanks ;)

Stratus3D:
Sorry for the late review @hugopeixoto. PR looks great! We will need to document this at some point though.

PR #231, opened 2017-09-25
Use origin when updating tags

hugopeixoto:
# Summary

`update_command` is using `git pull origin master`, when updating with `--head`,
but using `git fetch --tags` when updating to a tag, without specifying the remote.

This PR makes it use `git fetch origin --tags` instead.

This was preventing me from successfully running tests locally, as it was defaulting to another remote to which I currently don't have access.


Stratus3D:
Thanks!

Issue #166, opened 2017-03-02
Versioning of plugins


rmoriz:
Please consider adding functionality to install a specific version/to update to a specific version of a plugin. Also please consider versioning/tagging plugins in their repos.

The recent changes in the `nodejes` plugin broke a lot of setups.

Stratus3D:
@rmoriz good idea. I had thought about this before but decided it wasn't high priority since I hadn't heard any complaints. This is the first I've heard about it causing a problem, sorry about the trouble.

This will require a rewrite of the plugin install and plugin update commands, but it's definitely something we need to do. We will also need to tag all the plugins before making this change to asdf.

vic:
I guess the easiest change would be to allow `plugin-add` to take a git branch/ref name (default to master) and use that when cloning the plugin.

skotchpine:
[Just did a patch for this](https://github.com/asdf-vm/asdf/pull/234). Clones plugin into `<plugin-name>-<branch-name>` instead of `<plugin-name>`.

stephanos:
This would be a great addition! Just now the `elixir` plugin broke and we can only install the broken HEAD instead of a previous version. (see https://github.com/asdf-vm/asdf-elixir/issues/90 for details)

Stratus3D:
Questions:

* What commands should support plugin versions? `asdf plugin-add` is an obvious one, but it may be useful for other commands to take a plugin version argument too.
* When a plugin is added with a specific commit SHA as the version, how would plugin updates work?
* When a plugin is added with a specific branch as the version, how would plugin updates work?
* Should the plugin update command accept a version argument too?

dvic:
My 2 cents.

> What commands should support plugin versions? asdf plugin-add is an obvious one, but it may be useful for other commands to take a plugin version argument too.

I think just adding it to `asdf plugin-add`is sufficient (at least for my usecases).

> When a plugin is added with a specific commit SHA as the version, how would plugin updates work?

It doesn't, it prints a info message maybe? (that the plugin can't be updated because it's pinned to a specific SHA commit)

> When a plugin is added with a specific branch as the version, how would plugin updates work?

It simply checks if the branch (or tag) points to the same SHA commit, if not, it uses that SHA.

> Should the plugin update command accept a version argument too?

If it's ok to "downgrade", then yes, it could be useful in situations like #90.

jthegedus:
>When a plugin is added with a specific commit SHA as the version, how would plugin updates work?

[...]

PR #234, opened 2017-10-03
add starter plugin version feature (Addresses Issue #166)

skotchpine:
# Plugin Versioning

This isn't a finished product, but a first draft. It adds a versions to `plugin-add` via git branches. With this change, `plguin-add` takes an optional third argument `<git-branch>`. If a branch is given and is not `master`, then the plugin is cloned to `<plugin-name>-<git-branch>` instead of `<plugin-name>`.


lib/commands/plugin-add.sh
```diff
@@ -28,7 +35,7 @@ plugin_add_command() {
     display_error "Plugin named $plugin_name already added"
     exit 1
   else
-    if ! git clone "$source_url" "$plugin_path"; then
+    if git clone -b "$plugin_branch" "$source_url" "$plugin_path"; then
```
vic: I guess you removed the `!` un-intentionally, the script should fail if it's not able to clone the repo.

tomdavidson:
@skotchpine why branches rather than tags?

skotchpine:
@tomdavidson No reason. Why tags rather than branches?

tomdavidson:
Why closed? Vendoring the plugins would be great.

Branches just point to HEAD and are a moving target. In many cases tags vendor strong enough that a separate artifact registry is not needed. SCMs typically have built in features for tags https://github.com/asdf-vm/asdf/releases and there is an whole ecosystems of semver'd git tagged releases.

theoretick:
Any chance we can get this PR re-opened? Or happy to open a fresh one if the original author is unwilling.

Having some basic version locking would be great in order to prevent failures due to plugins being updated. In the meantime we have to use something a bit hackier like `cd`ing to the installation path and checking out a given SHA

jthegedus:
If this were to be reopened, I would like to lookup the default branch of the repo instead of assuming `master` given the recent changes to GitHub et al moving to `main`.

I need to finish #800 before we modify any of this too.

jthegedus:
@theoretick Yes please! Initial install being either remote HEAD, remote <branch> or remote <tag> is (I think) enough dials to satisfy people.

If you want to open a new PR I will happily review, extra points for tests ;)

theoretick:
Sorry for the delay! Opened https://github.com/asdf-vm/asdf/pull/916 as follow-up with a relevant test

Issue #236, opened 2017-10-10
Install using parent directory .tool-versions


DustinChaloupka:
### Steps to reproduce

- Create a directory with a currently uninstalled plugin/version in the `.tool-versions` file
- Create a child directory
- `cd` into child directory
- Run `asdf install`

### Expected behavior

I would expect the `asdf install` to use the parent directory's `.tool-versions` file since if the plugin/version is already installed it does work propertly.

### Actual behavior

`asdf install` errors about not having a `.tool-versions` file present or the name/version of the tool supplied.

### Environment

**OS**: MacOS

**asdf version**: 0.4.0

PR #237, opened 2017-10-10
Add installing from a parent directory's .tool-versions

DustinChaloupka:
# Summary

Adds installing from a parent directory's `.tool-versions` file if found.

Fixes: #236 

## Other Information

I looked quickly to see if there was anything like this in the `utils.sh` but did not see anything. The `while` loops is almost a copy of the one in `find_version` but wanted to keep the addition without refactoring too much. If there is a better way to go about doing this I am open to suggestions.

### DIFF
--- a/lib/commands/install.sh
+++ b/lib/commands/install.sh
@@ -37,10 +37,9 @@ get_concurrency() {
 }
 
 install_local_tool_versions() {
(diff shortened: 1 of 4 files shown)


tuvistavie:
Thank you very much!!

PR #245, opened 2017-10-24
Readme: format install instructions better

rstacruz:
# Summary

Update install instructions on the Readme to make it easier to understand :heart:

> image

## Other Information

I'd like to propose adding Fish + Fisherman integration via https://github.com/rstacruz/fish-asdf — but I figured it'd be best to clean this part up before adding more variants in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,25 +11,35 @@ Copy-paste the following into command line:
 git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.4.0
 ```
 
-Depending on your OS, run the following
-```bash
-# For Ubuntu or other linux distros
-echo -e '\n. $HOME/.asdf/asdf.sh' >> ~/.bashrc
-echo -e '\n. $HOME/.asdf/completions/asdf.bash' >> ~/.bashrc
+Depending on your OS and shell, run the following:
 
-# OR for Mac OSX
-echo -e '\n. $HOME/.asdf/asdf.sh' >> ~/.bash_profile
-echo -e '\n. $HOME/.asdf/completions/asdf.bash' >> ~/.bash_profile
-```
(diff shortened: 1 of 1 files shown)


vic:
Hey, thanks for taking the time to improve the README. And yes, separating the code snippets for each shell variant is nicer as people can just copy the one they need. Could you just resolve the conflict and we could merge this.

Regarding fish, we could have a link to your fish plugin, I'm not very sure as what difference/improvements does the plugin has over the fish scripts we currently have. If your plugin renders our scripts obsolete, we could remove them, @tuvistavie opinions on this?

tuvistavie:
Thank you for the changes, I think this is indeed simpler for new users.

> If your plugin renders our scripts obsolete, we could remove them, @tuvistavie opinions on this?

I think we should keep things as is for this part.
There are many different plugin managers for all shells, so I would rather keep asdf plugin manager agnostic, and keep the necessary files in the asdf repository so that anyone can use asdf without the need for other external tool or plugin.
If people want to write wrappers or plugins for asdf, this is of course very welcome, and we could maybe have a page in the wiki for these kind of contributions (I do have a plugin for fish too https://github.com/tuvistavie/fish-asdf), but I think it should not be part of asdf itself.

rstacruz:
> Regarding fish, we could have a link to your fish plugin, I'm not very sure as what difference/improvements does the plugin has over the fish scripts we currently have. 

Modifying your config files is not a good idea if you're using [fisherman](http://fisherman.github.io/).

rstacruz:
Conflicts resolved! btw this PR doesn't include `fish-asdf`, it's still the same info in the readme just reformatted.

tuvistavie:
Thank you!

Issue #242, opened 2017-10-19
duplicated shims and .tool-versions file suffixed with `-e-e-e`


bschaeffer:

akeating:
I've seen this a before. Here is a snippet:
```
> ls ~/.asdf/shims/
> ncu
-rwxr-xr-x    1 adam  staff   142B Oct 12 11:36 ncu-e-e-e-e
-rwxr-xr-x    1 adam  staff   140B Oct 12 11:39 ng
-rwxr-xr-x    1 adam  staff   196B Oct 12 11:36 node
-rwxr-xr-x    1 adam  staff   137B Oct 12 11:36 node-e-e-e-e-e-e-e-e-e-e-e
-rwxr-xr-x    1 adam  staff   195B Oct 12 11:36 npm
-rwxr-xr-x    1 adam  staff   185B Oct 12 11:39 npm-check-updates
-rwxr-xr-x    1 adam  staff   156B Oct 12 11:36 npm-check-updates-e-e-e-e
-rwxr-xr-x    1 adam  staff   136B Oct 12 11:36 npm-e-e-e-e-e-e-e-e-e-e-e
-rwxr-xr-x    1 adam  staff   165B Oct 12 11:36 npx
-rwxr-xr-x    1 adam  staff   136B Oct 12 11:36 npx-e-e-e-e-e-e-e-e-e-e
```
I think it's related to some _filename too long_ errors. Will try to produce some steps to repro.

vic:
Definitely a bug-e-e-e

vic:
yes, please, try to find a way to properly reproduce this :)

akeating:
Some quick testing this morning ...
```
> rm -rf ~/.asdf/shims/*
> asdf current nodejs
8.7.0 (set by /Users/adam/.tool-versions)
> asdf reshim nodejs
> ls ~/.asdf/shims/
-rwxr-xr-x   1 adam  staff   171B Oct 23 07:21 ncu
-rwxr-xr-x   1 adam  staff   142B Oct 23 07:21 ncu-e
-rwxr-xr-x   1 adam  staff   140B Oct 23 07:21 ng
-rwxr-xr-x   1 adam  staff   196B Oct 23 07:21 node
-rwxr-xr-x   1 adam  staff   167B Oct 23 07:21 node-e
-rwxr-xr-x   1 adam  staff   195B Oct 23 07:21 npm
[...]

PR #250, opened 2017-10-26
Duplicate shims

Jwashton:
# Summary

The MacOS version of `sed` requires you to provide a backup extension if you use the `-i` flag. Due to some weirdness in the implementation `-i'' -e` ignores the `''` and uses `-e` as the backup extension. This caused duplicate shims and `.tool-versions` files to show up with `-e` on the end of the filename.

Fixes: #242 

## Other Information

This problem only seems to show up if you don't provide a version to the reshim command. `asdf reshim elixir` would trigger it, but `asdf reshim elixir 1.5.2` would not. This inspired me to add a second dummy version to the reshim test, which does make it consistently fail on my machine without these fixes. However, by the time we reach the problematic sed command I don't see any difference in inputs or execution between user-provided versions or auto-discovered versions, so the exact reasoning there remains a mystery.

### DIFF
--- a/lib/commands/reshim.sh
+++ b/lib/commands/reshim.sh
@@ -58,7 +58,7 @@ write_shim_script() {
     cp "$plugin_shims_path/$executable_name" "$shim_path"
   elif [ -f "$shim_path" ]; then
     if ! grep "# asdf-plugin-version: $version" "$shim_path" > /dev/null; then
-     sed -i'' -e "s/\(asdf-plugin: $plugin_name\)/\1\\"$'\n'"# asdf-plugin-version: $version/" "$shim_path"
+     sed -i '' -e "s/\(asdf-plugin: $plugin_name\)/\1\\"$'\n'"# asdf-plugin-version: $version/" "$shim_path"
     fi
   else
     cat <<EOF > "$shim_path"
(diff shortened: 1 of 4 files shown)


Jwashton:
So, inspired by the failing test in CI I tried playing with this in on linux. Alas, I haven't found a way to get both platforms to be happy.

| Command                    | MacOS | Fedora |
| :------------------------- | :---: | :----: |
| `sed -i'' -e ...`          | N     | Y      |
| `sed -i '' -e ...`         | Y     | N      |
| `sed --in-place='' -e ...` | N     | Y      |
| `sed -i='' -e ...`         | N     | N      |

Any other ideas I should be trying?

jdashton:
Three options, @Jwashton: 

1. Detect the OS (or even run a command to try it and see what happens on the current box) and adjust the `-i ''` and `-i''` accordingly.
2. Actually make a backup with a valid extension and then delete it.
3. Stop using `-i`.

Even though it's easy to yell at macOS, this is probably an issue on other BSD boxes too.

afraser:
I have the same problem except I'm seeing `.bak` get appended to my shims.

CleanShot 2021-09-14 at 13 49 24@2x

asdf vv0.8.1-a1ef92a
macOS Catalina 10.15.7

Stratus3D:
I'm running a similar version:

```
asdf --version
v0.8.1-47e8fb0
```

And I am on macOS Catalina 10.15.7 and I do not observe this issue.

Issue #248, opened 2017-10-26
Realpath: command not found


tristandruyen:
### Steps to reproduce
Pull newest master into `~/.asdf` on OS X

### Expected behavior

asdf works as expected

### Actual behavior

`/Users/tristandruyen/.asdf/lib/utils.sh: line 246: realpath: command not found`

found a brewtab which adds realpath although i would prefer not to have this dependency or at least tell people about it in the README.md

### Environment

**OS**: OS X

**asdf version**: master

tuvistavie:
Oh, completely forgot `realpath` is not available by default on OS X.
Thanks for the catch, I just reverted the commit for now.

PR #251, opened 2017-10-28
Add test that greps source code for banned commands.

Stratus3D:
# Summary

Add test that greps source code for banned commands. Currently I'm only looking for the realpath command, but there are likely many other commands we shouldn't be using if we want to ensure cross platform compatibility.

This should prevent issues like #248 from occurring in the future.

## Other Information

We can add to this test by adding commands to the array of banned commands defined at the top of the file.

### DIFF
--- a/test/banned_commands.bats
+++ b/test/banned_commands.bats
@@ -1,0 +1,22 @@
+#!/usr/bin/env bats
+
+load test_helpers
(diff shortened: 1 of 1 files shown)


tuvistavie:
Thanks, this is very helpful!

Issue #255, opened 2017-12-09
Keep changing to .asdf directory if oh-my-zsh last_working_dir plugin on


mracos:
### Steps to reproduce

On `zsh`:
- Install `oh-my-zsh`
- Add to your `.zshrc` a `plugins=(last_working_dir)`

### Expected behavior

When creating a new shell, change to the last directory automatically 

### Actual behavior

Keeps changing to `.asdf`

### Environment

**OS**: ArchLinux

**asdf version**: v0.4.0

**shell**: zsh

Stratus3D:
I believe this has been fixed by https://github.com/asdf-vm/asdf/pull/334.

mracos:
Fixed here for me! Awesome, thanks!
Feel free to close it. :D

PR #256, opened 2017-12-09
Fix keeps changing oh-my-zsh last_working_dir

mracos:
# Summary

Fix keeps changing oh-my-zsh last_working_dir

Fixes: #255

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -7,7 +7,7 @@ else
 fi
 
 export ASDF_DIR
-ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
+ASDF_DIR="$(dirname "$(readlink -m "$current_script_path")")"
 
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR @mracos ! Is there any way we could do this without using `readlink -m`? `readlink` doesn't support that option on OSX.

I need to figure out why tests didn't catch this.

mracos:
As I don't have a mac I do not know how to test it.
But maybe we can just do `dirname $current_script_path` (but i don't know the implications and the bugs that this bring or some workaround with `pushd && popd`
what do you say? :thinking:

mracos:
Even commenting out the ASDF_DIR="$(dirname..." line makes all the tests pass, is that right?

trs:
Related: https://github.com/asdf-vm/asdf/pull/280

Stratus3D:
I believe this issue has been fixed in https://github.com/asdf-vm/asdf/pull/334. `cd` is no longer used in the subshell.

PR #262, opened 2017-12-15
Fix issue when having Elixir and R plugins installed

MarkVillacampa:
# Summary

When having both Elixir and R plugins installed, and an Elixir version set in `~/.tool-versions`, trying to fetch the current version of R would fail with:

```bash
$ asdf current r
version elixir is not installed for r
```

This happens because `exlixir 1.0` would be a match for `grep "r " ~/.tool-versions`. Adding `^` to the `grep` command to search from the begining of the line line fixes this issue.

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -214,7 +214,7 @@ parse_asdf_version_file() {
 
   if [ -f "$file_path" ]; then
     local version
-	version=$(grep "${plugin_name} " "$file_path" | sed -e "s/^${plugin_name} //")
+	version=$(grep "^${plugin_name} " "$file_path" | sed -e "s/^${plugin_name} //")
     if [ -n "$version" ]; then
       echo "$version"
       return 0


Stratus3D:
Thanks for catching the bug and contributing a fix!

I'd like to merge this in as soon as possible, but it looks like these changes caused a few of the tests to fail. Can you take a look at the failing tests and see if you can figure out why they failed? I'd take a look at the tests myself, but I don't have the time right now.

It might also be nice to write a unit test for this bug, to prevent regressions. We use `bats` for testing, and it's pretty simple to write unit tests that assert the asdf commands work as expected.

Stratus3D:
Thanks for the bug fix @MarkVillacampa !

I had to change one thing to get the tests passing. There was one sed command that replaced the old version with the new version, but the replacement string inserted ^ and $ on the line. The new string isn't regex, it's just a plain string, so I had to remove those characters.

PR #271, opened 2018-01-13
add to PATH once. Fixes #261

### LINKED ISSUE #261: Add to PATH only once
I use Z Shell as both a login shell and a main shell, so `.zshrc` gets sourced twice = I get a PATH with the following contents:

```
/home/user/.asdf/bin
/home/user/.asdf/shims
/home/user/.asdf/bin
/home/user/.asdf/shims
/usr/local/sbin
/usr/local/bin
/usr/bin
```

Not a big deal functionally, but this makes it harder to parse a PATH with eyes, when, for example, troubleshooting some ordering issue. Putting `source "${HOME}/.asdf/asdf.sh"` in `.zprofile` is not an option, since `source "${HOME}/.asdf/completions/asdf.bash"`, which needs to remain in `.zshrc`, depends on it. Please, consider doing [something of this sort](https://unix.stackexchange.com/a/124447/21976).

jthegedus:
# Summary

asdf currently adds itself to PATH even if it's already present. This fix checks whether asdf is within PATH before adding if not found.

Fixes: #261

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -8,7 +8,9 @@ fi
 
 export ASDF_DIR
 ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
-export PATH="${ASDF_DIR}/bin:${ASDF_DIR}/shims:$PATH"
+
+[[ ":$PATH:" != *":${ASDF_DIR}/bin:"* ]] && PATH="${ASDF_DIR}/bin:$PATH"
+[[ ":$PATH:" != *":${ASDF_DIR}/shims:"* ]] && PATH="${ASDF_DIR}/shims:$PATH"
 
 if [ -n "$ZSH_VERSION" ]; then
   autoload -U bashcompinit


asdf.sh
```diff
@@ -8,7 +8,9 @@ fi
 
 export ASDF_DIR
 ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
-export PATH="${ASDF_DIR}/bin:${ASDF_DIR}/shims:$PATH"
+
+[[ ":$PATH:" != *":${ASDF_DIR}/bin:"* ]] && PATH="${ASDF_DIR}/bin:$PATH"
+[[ ":$PATH:" != *":${ASDF_DIR}/shims:"* ]] && PATH="${ASDF_DIR}/shims:$PATH"
```
jthegedus: The checks for `/bin` and `/shims` are separate entries as I've noticed over the years that my PATH ordering gets shuffled around which one can only assume means other tools are editing and inserting between items in the PATH list.

Stratus3D:
Awesome thanks! I will merge as soon as the build finishes.

jeremejevs:
Thanks!

PR #272, opened 2018-01-14
Add `asdf current` to README and `asdf -h`

### LINKED ISSUE #257: Document command `asdf current`
By happy accident I found out that `asdf current` (without specifying a plugin) is a valid command that is recognized by `asdf` and prints a list of all currently used versions (and by which `.tool-versions`-files they are set).

But this is not documented: Neither `asdf -h` nor `Readme.md` are currently listing this information.

I think it would be a simple and useful clarity fix to document this command, because it is very useful to debug the complex version-interactions that certain programming projects rely on nowadays.

jthegedus:
# Summary

Add `asdf current` to README and `asdf -h`

PS. sorry about the markdown formatting, [Prettier](https://prettier.io/) ran on save

Fixes: Closes #257

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # asdf [![Build Status](https://travis-ci.org/asdf-vm/asdf.svg?branch=master)](https://travis-ci.org/asdf-vm/asdf)
+
 ### _extendable version manager_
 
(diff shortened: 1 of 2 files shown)


Stratus3D:
Thanks!

PR #273, opened 2018-01-14
Add print repo URLs with `asdf plugin-list` and `adsf plugin-list-all`

### LINKED ISSUE #268: Add option to the `plugin-list` command to show the plugin repo URLs
When I list my local plugins, it only shows me a list of the names each plugin is registered under. I want to be able to see the plugin repo URLs in addition to the names. Adding a flag to the existing command seems like the easiest way of doing this. Maybe a `-v` flag?

jthegedus:
# Summary
Add print repo URLs with `asdf plugin-list` and `adsf plugin-list-all`.
Fixes #268 

## Other Information

I decided not to use a flag as the raw output is itself not too useful and the terminal can afford the space to always print the URL. I'm happy to update to require a flag, but personally, I think it just adds more cruft to `-h` and README and is ultimately distracting for newcomers.

* Use printf for formatting
* Add error when plugin index at `/asdf_dir/repository/plugins` is not found
* add example of output to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,8 @@ asdf plugin-add <name> <git-url>
 ```bash
 asdf plugin-list
 # asdf plugin-list
+# java            https://github.com/skotchpine/asdf-java.git
+# nodejs          https://github.com/asdf-vm/asdf-nodejs.git
 ```
 
 ##### Remove a plugin
--- a/help.txt
+++ b/help.txt
@@ -1,8 +1,8 @@
 MANAGE PLUGINS
   asdf plugin-add <name> [<git-url>]   Add a plugin from the plugin repo OR, add a Git repo 
                                        as a plugin by specifying the name and repo url
-  asdf plugin-list                     List installed plugins
-  asdf plugin-list-all                 List plugins registered on asdf-plugins repository
+  asdf plugin-list                     List installed plugins with repository URLs
+  asdf plugin-list-all                 List plugins registered on asdf-plugins repository with URLs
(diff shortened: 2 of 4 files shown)


jthegedus:
I dev on a Ubuntu machine with Bash. Please don't tell me OSX doesn't have `printf` :cry:

jthegedus:
Output:
asdf-indicator-if-installed

jthegedus:
Apologies for adding to the OSX queue. I'll keep this in mind for future contributions.

jthegedus:
I added a `--urls` flag.

I've been using `printf` with string formatting to make the output nicer. This has made the tests a bit more cumbersome as the whitespace between some items is variable now. This formatting though should reveal regressions of formatting in the CLI.

asdf-fixed-flag

test/plugin_commands.bats
```diff
@@ -18,6 +18,7 @@ teardown() {
   [ "$status" -eq 0 ]
 
   run plugin_list_command
+  # whitespace between 'elixir' and url is from printf %-15s %s format
```
jthegedus: this can be removed. I just don't want to trigger yet another osx Travis build

tuvistavie:
Thank you very much for this PR and sorry for the delay.

> I still think that asdf plugin-list should always show the URLs. The internal calling of this command should specify a flag to not return the URLs. IMO

Some scripts might be dependant on the current behavior, so let's go with the `--url` option for now.

jthegedus:
I did come across the problem of other scripts depending upon this output, most of these commits were backtracking :sweat_smile:

PR #278, opened 2018-01-23
Update Bash Completions

kennyp:
# Summary

- Add missing commands to completion
- Auto-completion for `plugin-add`. Only include plugins not included.
- Auto-completion for `update`


kennyp:
@HashNuke or @Stratus3D thoughts?

Stratus3D:
Thanks!

Issue #279, opened 2018-01-23
asdf: command not found


antonovga:
MacOS

previously installed `rvm` and trying to install `asdf`
getting error `asdf: command not found`

$PATH "mess" after adding to `.bash_profile`
```
PATH=/Users/grg/.asdf
Saving session...
...copying shared history...
...saving history...
...completed./bin:/Users/grg/.asdf
Saving session...
...copying shared history...
...saving history...
...completed./shims:/Users/grg/.rvm/gems/ruby-2.5.0/bin:/Users/grg/.rvm/gems/ruby-2.5.0@global/bin:/Users/grg/.rvm/rubies/ruby-2.5.0/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/share/dotnet:/usr/local/go/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Users/grg/.rvm/bin
``` 

quick fix from https://github.com/asdf-vm/asdf/issues/98
`touch $HOME/.bash_sessions_disable`
fixes the problem

Maybe add this as a last point to the installation guide for macos?

Stratus3D:
This is related to #92.

Yes, I think adding it to the README would be a good idea. We should note the versions of OSX that require this change in the README.

antonovga:
Another solution
inside `.bash_profile` add
`export PATH="$HOME/.asdf/bin:$HOME/.asdf/shims:$PATH"`

I didn't understand why `asdf.sh`, which purpose is exactly as above string, brakes env when adding to `.bash_profile`

apisaka:
I've solved this with `reshim` as `asdf reshim <plugin_name>`

English3000:
I'm getting `bash: asdf: command not found` as well.

Followed the instructions. This is my `.bash_profile`:
```
source "$HOME/.bashrc"
export PATH="/usr/local/opt/qt@5.5/bin:$PATH"
# don't work
. $HOME/.asdf/asdf.sh

. $HOME/.asdf/completions/asdf.bash
```

Stratus3D:
@English3000 and you are sure the file is being sourced and that `$HOME/.asdf/asdf.sh` exists?

English3000:
```
~$ $HOME
-bash: /Users/amarkskatz: is a directory
~$ pwd
/Users/amarkskatz
~$ cd .asdf
~/.asdf ((v0.4.3))$ ls -A
.git              asdf.sh
[...]

PR #280, opened 2018-01-25
fix Macos nuance with bash sessions messing $PATH

antonovga:
# Summary
On MacOs there is a known problem with bash sessions which brake env vars.
Changing method of getting script dir fixes this problem

Method for retrieving current script dir in PR works with symlinks and "sourcing script".

On MacOs there is no `readlink`(MacOs is so MacOs) so we have to use `greadlink`

`2>/dev/null` - suppressing command not found for `readlink` under MacOs

Fixes: #279

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -7,7 +7,7 @@ else
 fi
 
 export ASDF_DIR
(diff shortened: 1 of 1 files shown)


Stratus3D:
A simpler approach to this was taken in #334 and I think it fixes this same issue. `cd` is no longer used. Now we only rely on `dirname`.

PR #287, opened 2018-02-13
Adds curl(1) as a dependency in Ubuntu

fxn:
Ubuntu does not ship with `curl(1)` pre-installed.


Stratus3D:
I don't believe curl is required by asdf. Is there a scenario where it is required?

fxn:
Oh, right, don't know exactly what should go in this particular repository.

I needed `curl` and also `build-essentials` to compile Ruby in a fresh install of Ubuntu. Should the Ruby plugin document them instead maybe?

Stratus3D:
@fxn yes, that's probably where it belongs. This repo only documents asdf, not any of the plugins. If the asdf-ruby plugin needs curl and it's not documented please open a PR there. Thanks!

Issue #291, opened 2018-02-23
Version error should mention asdf


pikeas:
### Steps to reproduce

```
$ asdf install minikube 0.25.0
$ minikube
No version set for minikube
```

### Expected behavior

    No version set for minikube: please run `asdf <global|local> minikube <version>`

It took me a bit to realize I had an asdf problem and not a minikube problem.

PR #293, opened 2018-02-25
Suggest action when no version is set

aj-foster:
# Summary

These changes give additional information when no version of a plugin is currently set. Using the suggested text from `@pikeas`, it recommends the `asdf` command a user should run.

Fixes: #291

### DIFF
--- a/bin/private/asdf-exec
+++ b/bin/private/asdf-exec
@@ -12,7 +12,7 @@ check_if_plugin_exists "$plugin_name"
 full_version=$(get_preset_version_for "$plugin_name")
 
 if [ "$full_version" == "" ]; then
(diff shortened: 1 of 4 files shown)


Stratus3D:
@aj-foster nice work! This is much more user friendly.

I'm going to merge this now since it is a step in the right direction, but I see a few things that could be improved further:

* That one line is duplicated several times. We should probably define a utility function somewhere that takes a tool name and prints the no version message. e.g. something like `print_no_version_set $plugin_name`. That way the message is defined in one place.
* This doesn't work with the `asdf which` command. Not sure what the deal is, but we likely have some things we need to fix for that command. See this screenshot:
screen shot 2018-02-25 at 1 16 43 pm

Issue #292, opened 2018-02-24
`current` command does not show details on versions set by env variables


Stratus3D:
### Expected behavior

The `asdf current` command should show the environment variable if a version is being set with an environment variable. The output should look something like this:

```
$ ASDF_ERLANG_VERSION=18.2 asdf current
...
erlang         18.2    (set by ASDF_ERLANG_VERSION environment variable)
...
```

### Actual behavior

Instead the command just says "set by <version>", which doesn't help the user at all since it doesn't say where or how.

```
$ ASDF_ERLANG_VERSION=18.2 asdf current
...
erlang         18.2    (set by 18.2)
...
```

### Environment

**OS**: affects all
**asdf version**: 0.4.2

PR #294, opened 2018-02-25
Specify when a version is set via environment variable

aj-foster:
# Summary

This change creates the following output when a version is set using an environment variable:

```
$ ASDF_ERLANG_VERSION=19.2 asdf current
erlang         19.2    (set by ASDF_ERLANG_VERSION environment variable)
...
```

Fixes: #292


Stratus3D:
Thanks @aj-foster ! This looks great!

Issue #296, opened 2018-02-25
Improve output of `asdf which`


Stratus3D:
The `asdf which` command doesn't work the same way other commands do when a tool version is not set. Refer to this screenshot for an example of how the output differs:

![screen shot 2018-02-25 at 1 16 43 pm](https://user-images.githubusercontent.com/1520926/36644960-c2f86b96-1a2f-11e8-9865-3c34c995fa3e.png)

I'm not sure how the command is implemented, but it should probably share the same error message printing code as the other commands. We should probably define a utility function somewhere that takes a tool name and prints the no version message. e.g. something like `print_no_version_set $plugin_name`. That way the no version message is defined in one place instead of being duplicated across the codebase.

#293

aj-foster:
#297 addresses this issue, but I did not use the `Fixes` auto-close feature because there's still refactoring to be done. (Probably by someone with better understanding of the overall architecture than me.)

PR #297, opened 2018-02-25
Refactor "no version set" messages

aj-foster:
# Summary

This PR refactors my previous changes by drawing the "no version set" messages into a utility. It also addresses the odd output from `asdf which` that was caused by running a version check in a subshell. A new test reflects this desired behavior.

Refer: #296

## Other Information

Further refactoring is deserved in `which_command`. Such was likely attempted previously with the unintended consequence of hiding the version check from the user. Also, given how `check_if_version_exists` works currently, the final `else` clause of `which_command` will never run.

/cc @Stratus3D


Stratus3D:
Thanks @aj-foster !

PR #298, opened 2018-02-26
Fix `plugin-list-all` test

aj-foster:
# Summary

The `plugin-list-all` test file was not run by `bats` due to the name of the file (`.sh` instead of `.bats`). This has been fixed, and the contents of the expected output now reflect the actual output (including the repository).


Stratus3D:
Good catch, Thanks!

Issue #288, opened 2018-02-19
v0.4.2 attaches asdf paths at the end of $PATH rather than at beginning


akatov:
### Steps to reproduce

updated asdf from `<= 0.4.1` to `0.4.2`

### Expected behavior

```bash
$ echo $PATH | tr ':' '\n'
/Users/me/.asdf/bin
/Users/me/.asdf/shims
/usr/local/bin
```

### Actual behavior

```bash
$ echo $PATH | tr ':' '\n'
/usr/local/bin
/Users/me/.asdf/bin
/Users/me/.asdf/shims
```

### Environment

**OS**: `Darwin 16.7.0 Darwin Kernel Version 16.7.0: Thu Jan 11 22:59:40 PST 2018; root:xnu-3789.73.8~1/RELEASE_X86_64 x86_64 i386 MacBookPro14,3 Darwin`

**asdf version**: `0.4.2`

mlitwiniuk:
can confirm, for now I had to manually alter `PATH`

Stratus3D:
Likely introduced by #271.

Stratus3D:
I'm unsure why this would be caused by #271 though, it appears to still be adding asdf to the beginning of PATH.

jthegedus:
I can't reproduce on Ubuntu 17.10

**asdf version**: `0.4.1`
```
/home/james/.local/bin
/home/james/.asdf/bin
/home/james/.asdf/shims
/usr/local/sbin
/usr/local/bin
/usr/sbin
/usr/bin
/sbin
/bin
/usr/games
/usr/local/games
/snap/bin
/opt/yarn-[version]/bin
/home/james/.asdf/installs/nodejs/8.9.4/.npm/bin
/home/james/.fzf/bin
```

then ran `asdf update` and reloaded shell.

**asdf version**: `0.4.2`
```
/home/james/.local/bin
/home/james/.asdf/shims
/home/james/.asdf/bin
/usr/local/sbin
/usr/local/bin
/usr/sbin
/usr/bin
/sbin
/bin
/usr/games
/usr/local/games
/snap/bin
/opt/yarn-[version]/bin
/home/james/.asdf/installs/nodejs/8.9.4/.npm/bin
/home/james/.fzf/bin
```

Stratus3D:
[...]

PR #303, opened 2018-03-06
add or move asdf to front of PATH

jthegedus:
# Summary

Some OSs and shells reorder items within PATH, this can result in the `asdf` entries being moved to the wrong place. This seems to occur on macOS and possibly due to the Tmux login shell option, see here for a discussion: https://github.com/thoughtbot/dotfiles/issues/587#issuecomment-369335238 .

`asdf` detects there are entries already in PATH and so does not prepend them to PATH, but due to the reordering by other tools, it stops functioning as intended.

## Proposed fix
1. Remove any `asdf` entries within PATH, regardless of their position.
2. Prepend the entries to PATH.

For PATHs without `asdf` nothing will be removed. Then the entries will be prepended.
For PATHs with `asdf` entries, they will be removed, even if in the correct position. Then the entries will be prepended.

Fixes: #288 

## Other Information

I noted in https://github.com/asdf-vm/asdf/pull/271 that PATH reordering happens but it didn't occur to me to fix it this way at the time :man_shrugging: sorry?

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -9,8 +9,17 @@ fi
 export ASDF_DIR
 ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
 
-[[ ":$PATH:" != *":${ASDF_DIR}/bin:"* ]] && PATH="${ASDF_DIR}/bin:$PATH"
-[[ ":$PATH:" != *":${ASDF_DIR}/shims:"* ]] && PATH="${ASDF_DIR}/shims:$PATH"
+# Add asdf to PATH
+#
+# if in $PATH, remove, regardless of if it is in the right place (at the front) or not.
+# replace all occurrences - ${parameter//pattern/string}
+ASDF_BIN="${ASDF_DIR}/bin:"
+ASDF_SHIMS="${ASDF_DIR}/shims:"
+[[ ":$PATH:" == *"${ASDF_BIN}"* ]] && PATH="${PATH//$ASDF_BIN/}"
+[[ ":$PATH:" == *"${ASDF_SHIMS}"* ]] && PATH="${PATH//$ASDF_SHIMS/}"
+# add to front of $PATH
+PATH="${ASDF_DIR}/bin:$PATH"
+PATH="${ASDF_DIR}/shims:$PATH"
 
 if [ -n "$ZSH_VERSION" ]; then
   autoload -U bashcompinit
--- a/asdf.sh
+++ b/asdf.sh
@@ -18,8 +18,8 @@ ASDF_SHIMS="${ASDF_DIR}/shims:"
 [[ ":$PATH:" == *"${ASDF_BIN}"* ]] && PATH="${PATH//$ASDF_BIN/}"
 [[ ":$PATH:" == *"${ASDF_SHIMS}"* ]] && PATH="${PATH//$ASDF_SHIMS/}"
 # add to front of $PATH
-PATH="${ASDF_DIR}/bin:$PATH"
-PATH="${ASDF_DIR}/shims:$PATH"
+PATH="${ASDF_BIN}$PATH"
+PATH="${ASDF_SHIMS}$PATH"
 
 if [ -n "$ZSH_VERSION" ]; then
   autoload -U bashcompinit
--- a/asdf.sh
+++ b/asdf.sh
@@ -15,8 +15,8 @@ ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
 # replace all occurrences - ${parameter//pattern/string}
 ASDF_BIN="${ASDF_DIR}/bin:"
 ASDF_SHIMS="${ASDF_DIR}/shims:"
-[[ ":$PATH:" == *"${ASDF_BIN}"* ]] && PATH="${PATH//$ASDF_BIN/}"
-[[ ":$PATH:" == *"${ASDF_SHIMS}"* ]] && PATH="${PATH//$ASDF_SHIMS/}"
+[[ "$PATH:" == *"${ASDF_BIN}"* ]] && PATH="${PATH//$ASDF_BIN/}"
+[[ "$PATH:" == *"${ASDF_SHIMS}"* ]] && PATH="${PATH//$ASDF_SHIMS/}"
 # add to front of $PATH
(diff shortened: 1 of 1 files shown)


asdf.sh
```diff
@@ -9,8 +9,17 @@ fi
...
-[[ ":$PATH:" != *":${ASDF_DIR}/bin:"* ]] && PATH="${ASDF_DIR}/bin:$PATH"
-[[ ":$PATH:" != *":${ASDF_DIR}/shims:"* ]] && PATH="${ASDF_DIR}/shims:$PATH"
+# Add asdf to PATH
+#
+# if in $PATH, remove, regardless of if it is in the right place (at the front) or not.
+# replace all occurrences - ${parameter//pattern/string}
+ASDF_BIN="${ASDF_DIR}/bin:"
+ASDF_SHIMS="${ASDF_DIR}/shims:"
+[[ ":$PATH:" == *"${ASDF_BIN}"* ]] && PATH="${PATH//$ASDF_BIN/}"
+[[ ":$PATH:" == *"${ASDF_SHIMS}"* ]] && PATH="${PATH//$ASDF_SHIMS/}"
```
Stratus3D: What's the purpose of the colon prefix here? I think I understand why the trailing colon is needed - it's because the `ASDF_SHIMS` string has a colon as it's last character so if we need to match the last path in PATH we'd need it to end with a colon too.

jthegedus: Do you mean

Stratus3D: It's fine with me either way.

jthegedus:
So I'm going to hit merge now https://www.youtube.com/watch?v=-W6as8oVcuM

gohanlon:
Thanks for tackling this issue, @jthegedus. Sorry I didn't get to looking at this PR earlier, but from reading the changes, I think there are some issues left to address. I'm not particularly knowledgeable or experienced at shell scripting, so please check carefully for any mistakes I may make below.

The expression that checks if a `PATH` segment is already present unintentionally matches a partial path segment. Consider the following:

```
% MY_PATH="a:bc:d"
% [[ "$MY_PATH:" == *"c:"* ]] && echo "uh oh"
uh oh
```

Here's a simple approach for matching only a whole `PATH` segment:

```
% MY_PATH="a:bc:d"
% [[ ":$MY_PATH:" == *":c:"* ]] && echo "uh oh"
% [[ ":$MY_PATH:" == *":bc:"* ]] && echo "yay"
yay
% [[ ":$MY_PATH:" == *":a:"* ]] && echo "yay"
yay
```

(This is actually how it was in e2a98e13450efb1df93e34ea99b5204bf1e6dff2, but was changed in b0342d2f78a4fcbe95e151adacf1d58ead087597.)

Note that the first segment of a `PATH` lacks a leading colon, and the last segment lacks trailing colon. Padding with a leading and trailing colon — `":$MY_PATH:"` in immediately preceding example — allows a single expression to match the first, last, or any middle segment of the `PATH` string. This works by "normalizing" the `PATH` string such that all segments begin and end with a colon.

Another issue is that the substitution expression in this PR won't properly remove `ASDF_BIN` (or `ASDF_SHIMS`) from the `PATH` when `ASDF_BIN` is the last `PATH` entry:

```
% MY_PATH="a:b:c"
% echo ${MY_PATH//c:/}
a:b:c
```

Thinking aloud, here's how I'd approach this. Note that while I have thought this through fairly carefully, I haven't actually tested any of the following.

First, to make it easier to keep track of all the colons flying around, define `ASDF_BIN` and `ASDF_SHIMS` without a trailing colon:

```
ASDF_BIN="${ASDF_DIR}/bin"
ASDF_SHIMS="${ASDF_DIR}/shims"
```

We'll use our trick normalize the `PATH` string:

```
tmp=":$PATH:"
```

Checking if `PATH` already contains asdf's segments isn't necessary. Instead, we can rely our substitution expression to remove each of asdf's `PATH` segments if they are present and do nothing if they are not present.

Now, we remove each of asdf's PATH segments, removing one colon along with each. If we were to remove both the leading and trailing colon with each segment either we'd break our normalized `PATH` string (when matching the first or last segment) or we'd inadvertently merge two segments together (when matching a middle segment):

```
tmp="${tmp//:$ASDF_BIN/} 
tmp="${tmp//:$ASDF_SHIMS/}
```

Finally, we can remove the leading colon we added, and then remove the trailing colon we added:

```
tmp=${tmp#:}
tmp=${tmp%:}
```

Now we have a valid `PATH` string:

```
PATH=tmp
```

And of course, now we can add asdf's `PATH` segments:

```
PATH="$ASDF_BIN:$ASDF_SHIMS:$PATH"
```

The only unhandled case I can think of is when for some reason `PATH` is empty at the start of the script. I think everything except the final `PATH` assignment would be okay. The problem with the final assignment is that the final `PATH` would end with `:`, which I think is the same as ending with `:.`, which poses a security risk.

If asdf is comfortable assuming the system has `tr`, here's  safer way to set the `PATH`:

```
PATH=`echo $ASDF_BIN $ASDF_SHIMS $PATH | tr ' ' ':'`
```

I personally can't think of any *NIX system that doesn't include`tr` — probably some very stripped down systems don't have it (like minimal Linux distros that are popular for use with Docker). Here's a pure bash solution (from [Stack Overflow](https://stackoverflow.com/questions/1527049/join-elements-of-an-array)):

```
function join_by { local IFS="$1"; shift; echo "$*"; }

PATH=$(join_by : $ASDF_BIN $ASDF_SHIMS $PATH)
```

jthegedus:
@gohanlon You're right! Thanks for the thorough explanation! I'm not too experienced with Bash myself (as you can see) and manually testing without being able to reproduce the initial error was obviously not thorough enough :sweat_smile: 

I'll look into adding back the `:` as a quick fix and work on a more concrete solution to this with some updated tests for these edge cases you've raised. If it's okay I'll tag you to get your feedback on the more concrete solution when I PR something :smile:

gohanlon:
Sounds great. Thanks, @jthegedus!

Stratus3D:
FYI @jthegedus we are already using `tr` in asdf when constructing version variable names in `utils.sh`, so feel free to use it anywhere else it could be helpful. Unless of course you find a platform you think we should support that lacks `tr`, in which case we will need to refactor our existing code.

Thanks @jthegedus !

Stratus3D:
And thanks @gohanlon for the thorough explanation!

PR #304, opened 2018-03-08
begin stabalising asdf prepending to PATH logic

jthegedus:
# Summary

as per suggestion from here - https://github.com/asdf-vm/asdf/pull/303#issuecomment-371231967 :
* remove colon from `ASDF_` variables
* use `:` to normalize PATH
* match against whole `ASDF_` PATH segments
* add `:` back to string replacement - this still doesn't match against asdf segments when last in PATH
* add `:` back to asdf/PATH merging as it's no longer in the `ASDF_` vars

## Other Information

This fix is a stopgap to address the more serious problems (matching partial segments) raised in https://github.com/asdf-vm/asdf/pull/303#issuecomment-371231967 . I plan to address the rest of the issues is another PR with tests for each edge case.

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -15,8 +15,8 @@ ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
 # replace all occurrences - ${parameter//pattern/string}
 ASDF_BIN="${ASDF_DIR}/bin:"
 ASDF_SHIMS="${ASDF_DIR}/shims:"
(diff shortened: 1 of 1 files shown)


jthegedus:
@gohanlon it would be good to get your eyes on this too before it goes through.

gohanlon:
Looks good to me!

jthegedus:
Cheers all :smiley:

Stratus3D:
Just tagged a new version that includes all these fixes - https://github.com/asdf-vm/asdf/blob/master/CHANGELOG.md

PR #306, opened 2018-03-29
add a more robust check for `bins` in `shims` dir for fish shell users

ipatch:
# Summary

Add a more robust check so the `$asdf_dir/{bin,shims}` directories aren't repeatedly added when launching sub shells from the current shell, or launching a program such as tmux.

## Other Information

Just a quick little fix. 👍

cheers 🍻
Chris


Stratus3D:
@ipatch looks good to me!

@tuvistavie you are a fish user right? Can you take a look at this? I don't use fish shell.

tuvistavie:
LGTM!

Issue #308, opened 2018-03-30
"No version set" message should go to stderr


neilmb:
### Steps to reproduce

I thought I would stop using `asdf` for a particular project
```
$ rm .tool-versions
```

Then I ran a command that redirected its stdout to a file
```
$ gshuf march_signup_domains_unique.txt | pipenv run python crawlme.py > march_signup_crawl.json
```
where `pipenv` was installed by `asdf`.

### Expected behavior

The `asdf` error message about "No version set ..." should have been written to stderr so that it still would have still appeared when stdout was redirected.

### Actual behavior

This showed no error message because the "No version set" message was printed to stdout and was swallowed into the output file
```
$ head march_signup_crawl.json
No version set for python; please run `asdf <global | local> python <version>`
```
### Environment

**OS**: Mac OS X

**asdf version**: v0.4.3

PR #309, opened 2018-03-30
Print "No version set" message to stderr

neilmb:
# Summary

Print "No version set" message to stderr instead of stdout.

Fixes: https://github.com/asdf-vm/asdf/issues/308


Stratus3D:
Thanks! I think this is a good change, but it looks like one of the unit tests is failing due to this message not being printed to STDOUT anymore. Can you fix the test?

neilmb:
I have a fix, let me add it to this PR...

Issue #305, opened 2018-03-26
Use exit code 127 on missing version


endorama:
Hello, 

first of all thank you for this project. I'm using `asdf` since a couple of month and is really awesome! There is little thing that I think could be improved, and I can provide a PR if you're interested.

Currently when you install a plugin it adds a shim in the path. This makes impossible to test if the command is available using "usual" checks, because is always available.
Currently after displaying the error message the command exists with [`-1` or `1`](https://github.com/asdf-vm/asdf/search?utf8=%E2%9C%93&q=display_no_version_set&type=). 
The Advanced BASH Scripting guide has some [codes with special meaning](http://tldp.org/LDP/abs/html/exitcodes.html), and from my searches seems to be the *de-facto* standard for exit codes in shell scripting ( even zsh uses the `127` for missing commands ).

What is the reason for the current behaviour? Could be aligned with the special exit codes from the BASH Scripting guide ( thus using `127` as exit code if no version is found )? Thank you.

### Steps to reproduce

- Install a plugin ( es `kops` )
- Run any plugin with no version set

### Expected behavior

- Warning message is printed
- Exit code is `127`

### Actual behavior

- Warning message is printed
- Exit code is `255` or `1`

### Environment

**OS**: Ubuntu 16.04

**asdf version**: v0.4.3

Stratus3D:
@endorama yes those exit codes can be updated. I don't think an exit code of 255 or 1 is wrong per se, but I understand why 127 would make sense. If you create a PR that makes the exit codes more logical I will not object to it. Though because it changes the codes it may not be wise to release it in a patch version. We may need to do a minor release since this could break existing code that integrates with asdf.

PR #310, opened 2018-04-05
Change exit code

endorama:
# Summary

As discussed in #305, this PR aligns exit codes for plugin not installed and plugin version not set to the one specified in the [BASH Advanced scripting guide](http://tldp.org/LDP/abs/html/exitcodes.html).

Fixes: #305 

## Other informations

Code `127` has been used to exit when the requested plugin version is not installed, resulting in a  "command not found" error.  
Code `126` has been used to exit when there is no version set for the requested plugin, resulting in a "Command invoked cannot execute" error.

Thank you!

### DIFF
--- a/bin/private/asdf-exec
+++ b/bin/private/asdf-exec
@@ -13,7 +13,7 @@ full_version=$(get_preset_version_for "$plugin_name")
 
 if [ "$full_version" == "" ]; then
   display_no_version_set "$plugin_name"
-  exit -1
+  exit 126
 fi
 
 # shellcheck disable=SC2162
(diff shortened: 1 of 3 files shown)


Stratus3D:
Thanks for the PR! It looks like one of the tests checks the exit code returned from one of the commands and is failing with these changes. Can you update the PR so the test passes? We use `bats` for tests.

endorama:
@Stratus3D Ops :) Sorry I forgot about `bats`. 

I fixed the failing test, but I wasn't able to find any test for `asdf-exec`, where most of the changes are. Should I add it? I tried, but I don't understand I would I be able to run a "fake" plugin for testing this.

Thanks!

Stratus3D:
@endorama thanks for fixing the test! No worries about tests for asdf-exec. I'll try and write some when I get a chance.

Issue #307, opened 2018-03-29
Add ability to list all installed versions of all packages


rlipscombe:
I'd like to get a list of all installed versions of all packages.

```
$ asdf list
No plugin given
```

It'd be nice if it iterated through the installed plugins and listed the installed versions for each. I cobbled the following together:

```
$ asdf plugin-list | xargs -I{} sh -c "echo {}; asdf list {}"
elixir
No versions installed
erlang
19.3.6.7
golang
No versions installed
nodejs
No versions installed
python
No versions installed
ruby
2.2.7
2.3.4
```

It'd be nice if something like this was built in.

Christopher-Bui:
Should this be closed now?

Stratus3D:
Thanks @Christopher-Bui. Yes, this has been implemented.

PR #311, opened 2018-04-16
List command without parameter lists all installed packages with versions

hlhr:
# Summary
* Add feature to command `asdf list` to list all installed plugins with versions
* Add test to test the new behaviour, change the title of the existing tests

Fixes: Add ability to list all installed versions of all packages #307 

## Other Information
Feel free to give me feedback (specially for the bash code style)


Stratus3D:
This looks great @hlhr ! Thanks for taking the time to write a new test for this!

Issue #276, opened 2018-01-22
Feature Idea: asdf install to also install plugins


PragTob:
First of all, thanks for `asdf` - it's a pleasure to work with! Especially when using multiple languages. Also love how it's gradually gotten better (love installing plugins by name!)

This is just an idea for a little feature improvement that I'd like to see.

### Steps to reproduce

Get a repo with a `.tool-versions` of many different plugins. Run `asdf install`

```
$ cat .tool-versions 
rust 1.23.0
crystal 0.24.0
nodejs 9.3.0
elixir 1.6.0
haskell 8.2.2
clojure 1.8.0
golang 1.9.2
ruby 2.5.0
```

### Expected behavior

I'd have loved that despite installing all the versions of plugins I already got it'd also install the plugins themselves (i.e. I don't have `haskell` installed, first install it then the specified version).

### Actual behavior

On the first plugin it doesn't have installed, asdf aborts.

```
$ asdf install
rust 1.23.0 is already installed
# lots of stuff
No such plugin # hit haskell here
```

### Environment

**OS**: Linux Mint 18.3

**asdf version**: 0.4.1

Stratus3D:
I don't have much time right now but I do want to weigh in on this. I have some concerns about the security aspects of this. Here is a scenario:

1. Malicious code gets into a plugin in our plugins repo/list (quite possible, since we don't control most of them)
2. asdf user clones down a git repository for an open source project. Project contains a `.tool-versions` file that references the asdf plugin in the plugin repo containing malicious code.
3. User runs `asdf install`. Malicious code in plugin is invoked and can do anything on the users system.

Notice the user did not run any code in the software project they just cloned down. Maybe they don't actually trust the code at all. But it's not obvious to the user that the `.tool-versions` file is actually dictating what code is downloaded and run on their system when they run `asdf install`.

While this may not be very likely, it is possible and poses a very real security risk. I know that this issue just builds on the previous plugin-repository work, but I think this is probably a bad idea. If we do decide to move forward with this there must be a configuration setting added to asdf to disable this functionality entirely, so that paranoid people like me can sleep at night. I manually install all my asdf plugins so I know exactly what's installed on my system (https://github.com/Stratus3D/dotfiles/blob/master/scripts/setup.sh#L127). I do not let the plugin repository dictate what gets installed.

Stratus3D:
[...]

PR #313, opened 2018-04-24
feat: pass --add to asdf install to add plugins in .tool-versions

trs:
## Summary

`asdf install --add` will add any plugins from `.tool-versions` before installing them (if they are in the repo), saving time and effort when adding plugins listed there.

Fixes: https://github.com/asdf-vm/asdf/issues/276

Let me know if more needs to be done to satisfies security concerns or anything.

### DIFF
--- a/lib/commands/install.sh
+++ b/lib/commands/install.sh
@@ -93,7 +93,7 @@ install_add_tool_plugin() {
   plugin_path=$(get_plugin_path "$plugin_name")
 
   if ! [ -d "$plugin_path" ]; then
-    plugin_add_command $plugin_name
+    plugin_add_command "$plugin_name"
   fi
 }
 
--- a/lib/commands/install.sh
+++ b/lib/commands/install.sh
@@ -14,8 +14,15 @@ install_command() {
   local plugin_name=$1
   local full_version=$2
 
-  if [ "$plugin_name" = "" ] && [ "$full_version" = "" ]; then
-    install_local_tool_versions
+  if [[ ( "$plugin_name" = "" || "$plugin_name" = "--add" ) && "$full_version" = "" ]]; then
+    local add_plugins
+    if [[ $plugin_name = "--add" ]]; then
+      add_plugins=true
+    else
+      add_plugins=false
+    fi
+
+    install_local_tool_versions "$add_plugins"
   elif [[ $# -eq 1 ]]; then
     display_error "You must specify a name and a version to install"
     exit 1
@@ -37,6 +44,8 @@ get_concurrency() {
 }
 
 install_local_tool_versions() {
+  local add_plugins=$1
+
   local plugins_path
   plugins_path=$(get_plugin_path)
 
@@ -54,6 +63,11 @@ install_local_tool_versions() {
       plugin_version_and_path="$(find_version "$plugin_name" "$search_path")"
 
       if [ -n "$plugin_version_and_path" ]; then
+
+        if [ "$add_plugins" = true ]; then
+          install_add_tool_plugin "$plugin_name"
+        fi
+
         local plugin_version
         some_tools_installed='yes'
         plugin_version=$(cut -d '|' -f 1 <<< "$plugin_version_and_path")
@@ -73,6 +87,15 @@ install_local_tool_versions() {
   fi
 }
 
+install_add_tool_plugin() {
+  local plugin_name=$1
+  local plugin_path
+  plugin_path=$(get_plugin_path "$plugin_name")
+
+  if ! [ -d "$plugin_path" ]; then
+    plugin_add_command $plugin_name
+  fi
+}
 
 install_tool_version() {
   local plugin_name=$1


danhper:
Thanks for the PR and very sorry for the huge delay.
This feature has been asked for a couple of times and I think this could be helpful.
@asdf-vm/core Any opinion on this?

trs:
No worries! I've rebased this against master.
Let me know if there's anything else I can help with.

jthegedus:
@danhper I believe the main reasons for not adding this so far was because there may be multiple plugins used for the same lang/framework/tool. For instance https://github.com/Banno/asdf-hashicorp

danhper:
@trs Thanks a lot!

@jthegedus I am not entirely sure I get it. If the plugin does not exist, this would just be equivalent to running

```
asdf plugin-add PLUGIN_NAME
```

which will use the plugin registered in our asdf-plugins repo with this name.
If somebody wants a different plugin with the same name, he can always run `asdf plugin-add PLUGIN_NAME URL` and then even if the `--add` flag is passed, the plugin will not be updated or anything.
Am I missing something here?

kvokka:
I like the idea of this feature.

If the user want to check the the repos for security, he should do it anyway.

With out this feature all earlier mentioned security [problems](https://github.com/asdf-vm/asdf/issues/276#issuecomment-359666776) are at the same place- it just help to reduce the annoying installation process.

Asdf tool does not contain any tool, which have 2 official repo, so it does not looks like a problem.

This branch is outdated, if you want, I can rebase it and add missing docs.

danhper:
@trs Thanks for rebasing!
@Stratus3D @jthegedus I think this is a good addition, any opinion?

jthegedus:
My only opinion is that I am keen to see this land! Will certainly help me onboard new users at work.

danhper:
Ok, great! Feature-wise I think it's good but we do need some tests.
@trs Do you think you would have time to add tests for this feature, please?

aliuk2012:
It's been over a year now, is there any progress on this feature. It would make a huge difference to be able to confidently know that when joining an existing project that uses asdf that you could install all its tools with two or less commands instead of manually running `asdf plugin add ...` a number of times.

jthegedus:
@aliuk2012 Sorry this has lagged, we've been making our way through some of the open PRs and trying to finish off our `0.8.0` release. I'll commit to getting this in soon (likely after `0.8.0` release) as I also want this to land.

I think the only remaining questions around this is if it's possible to get some test cases to measure the effect of the `--add` flag on this command. If someone has the time over the next week to add those that would be appreciated, else I will get to it soon.

jthegedus:
Further thoughts on this feature request

- How would missing system dependencies be handled for plugins? The python plugin for instance requires both `pyenv` deps and `python build` deps
- How are plugins installed that are not in `asdf-plugins`? perhaps your company have their own internal plugins or forks of public plugins vetted by security which download from internal mirrors of public repos?

I was too hasty in my previous comment to commit this to the next major release as there's **increasingly more scenarios where this feature introduces more problems for the user than solves**.

The _issues_ it solves are actually required setup of the tool. There's certainly room to move and improve, but I don't think this is necessarily the right solution (happy to be convinced otherwise).

So in a complete 180 I'm going to remove the milestone tag.

Stratus3D:
> I was too hasty in my previous comment to commit this to the next major release as there's increasingly more scenarios where this feature introduces more problems for the user than solves.

> The issues it solves are actually required setup of the tool. There's certainly room to move and improve, but I don't think this is necessarily the right solution (happy to be convinced otherwise).

I agree with @jthegedus on this. This pull request solves the easy part of the problem - running `asdf plugin-add <plugin>` - and doesn't address the hard part that users will have to perform manually - installing dependencies and setting up the environment for the tool. asdf is just a version manager it, doesn't setup the environment.

I suggest we close this PR. I understand things like this would make developer easier for many folks, but I don't think this is the right approach. I think **asdf should provide small, focused, stable API so that other developers can build their own tooling around asdf for things such as environment setup**. I'm happy to provide input if @trs or other developers want to build additional tooling around asdf's current API.

jthegedus:
>I suggest we close this PR.

codyrobbins:
I wish this could be revisited. I feel like the step of adding plugins is entirely superfluous: `asdf` knows exactly what plugins are installed and from `.tool-versions` it knows which ones need to be installed—why do I have to manually go through the motions of running an incantation myself for no reason? For context, I’m primarily looking at this from the viewpoint of onboarding new developers to a project and it looks like a number of people above are too. Sure, instructions for this can go in a readme or a bootstrapping script, but… why? It can just happen automatically.

I understand the security implications of automatically running code, but I think strategies for dealing with this were properly addressed in #276 in terms of making this behavior opt-out via `.awsrc` or opt-in via an `--add` flag. And I think closing this because it doesn’t solve all possible problems is kind of silly. This change would be an improvement—even if there are all sorts of various edge cases like plugin dependencies and environment configuration that it doesn’t address. But why does it need to?

I really love `asdf` but the conscious choice not to merge this seems like making perfect the enemy of the good.

ngyewch:
@codyrobbins I've got a simple tool for this. I ran into the same problem myself. https://github.com/ngyewch/asdf-helper

sheldonhull:
I ended up switching everything to [aqua](https://github.com/aquaproj/aqua) which fit the bill, and has some additional steps to harden security. Binary installs, pinning versions, very easy to contribute new tools, checksum verification, cross-platform, global vs project installs. 

It doesn't handle runtimes, like Python, so asdf is better for that, but since I work with Go, it ended up taking care of every tool I need. If you are in a different ecosystem with runtime tooling dependencies for specific versions of python, ruby, etc, then asdf is probably a better fit for those. For all other binary based tools, highly recommend checking out aqua.

PR #315, opened 2018-04-30
Include plugin name in error message if plugin doesn't exist

mxie:
# Summary

Previously, if we ran a command like `asdf install` and we encountered a
plugin that didn't exist, asdf produces an error saying "No such
plugin". Without knowing which plugin it could be referring too, we'd
have to manually go through each plugin in `.tool-versions` to find the
culprit.

With this commit, we'll now also include the plugin name as part of the
messaging for easier debugging.

## Other information

Before:

```sh
$ asdf install
No such plugin
```

After:

```sh
# assuming `doesnt-work 1.2.3 is in .tool-versions`
$ asdf install
No such plugin: doesnt-work
```

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -54,14 +54,16 @@ list_installed_versions() {
 }
 
 check_if_plugin_exists() {
+  local plugin_name=$1
+
   # Check if we have a non-empty argument
   if [ -z "${1}" ]; then
     display_error "No plugin given"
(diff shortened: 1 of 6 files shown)


tuvistavie:
This is indeed way better, thank you very much!

mxie:
Thank you for the quick turnaround and the merge!

PR #317, opened 2018-05-06
add bash for error: command 'asdf' not found

sgnd:
Just add this command `exec $SHELL` to update the shell. I am running with Ubuntu 18.04. If `exec $SHELL` is not updated, you can't run asdf in terminal.


Stratus3D:
This change would need to be made to the 3 other code snippets under the "Setup" section in the readme since it affects all of them. It may be better to just say something along the lines of "after running these commands make sure to restart your shell" at the bottom of the setup section.

Issue #322, opened 2018-05-25
`asdf plugin-add` should have a different return code for already-installed plugins


the-mikedavis:
### Steps to reproduce

```shell
$ asdf plugin-list
elixir
...
$ asdf plugin-add elixir
Plugin named elixir already added
$ echo $?     # return code of last command
1
$ asdf plugin-add aoestnuhaosnetu
Plugin aoestnuhaosnetu not found in repository
$ echo $?
1
```

### Expected behavior

Trying to install a plugin that's already added should return a different code than trying to install a plugin that doesn't exist. This will make it easier to script asdf installations in prod environments. It doesn't really matter what the exit code is; just that it's different from a non-existent plugin.

IMHO, if the plugin's already there, asdf should return `0`. The user assumes calling `asdf plugin-add <plugin>` will result in asdf having that plugin if the call was successful. If the plugin's already there, that assumption is not violated. Therefore the return code should be successful (`0`).

### Actual behavior

Attempting to install plugins that are already added and installing plugins that don't exist both have the same return code of `1`.

### Environment

**OS**: MacOS High Sierra, Zsh, git-clone installation of `asdf`.

**asdf version**: v0.4.3

PR #323, opened 2018-05-25
Make the exit code of an already installed plugin distinct

the-mikedavis:
When `plugin-add` is called on a plugin that's already installed, asdf will return `0`.

```shell
$ asdf plugin-add elixir
[ clones repository, clones asdf-elixir ]
$ echo $?       # last exit code
0
$ asdf plugin-add elixir
Plugin named elixir already added
$ echo $?
0
```

Connects asdf-vm/asdf#322.

# Summary

As described in asdf-vm/asdf#322, there should be different exit codes for the cases of running `asdf plugin-add <plugin>`:

case | current exit code
----------------|----------------------
new install, successful |  0
already installed | 1
failed to install | 1

To be clear: **I care about case 2** (_already installed_).

Having the same exit code (`1`) for the second and third cases makes it difficult to write scripts to install asdf in production environments (e.g. via Ansible) because running `asdf plugin-add elixir` twice will fail the playbook.

In my opinion, `0` is the proper exit code for running `asdf plugin-add <plugin>` when `<plugin>` is already installed because with a successful exit code, the user should be able to assume that `<plugin>` is now available.

## Other Information

I am open to debate on whether this exit code should be some other value (e.g. `2`), but not `1`.

### Current workaround

```shell
plugin=elixir
if ! asdf plugin-list | grep $plugin > /dev/null; then
   asdf plugin-add $plugin
fi
```

### DIFF
--- a/lib/commands/plugin-add.sh
+++ b/lib/commands/plugin-add.sh
@@ -26,7 +26,7 @@ plugin_add_command() {
 
   if [ -d "$plugin_path" ]; then
     display_error "Plugin named $plugin_name already added"
-    exit 1
+    exit 0
   else
     if ! git clone "$source_url" "$plugin_path"; then
       exit 1
--- a/lib/commands/plugin-add.sh
+++ b/lib/commands/plugin-add.sh
@@ -26,7 +26,7 @@ plugin_add_command() {
 
   if [ -d "$plugin_path" ]; then
     display_error "Plugin named $plugin_name already added"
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Failing error codes are alright by me. I don't know if there's some perfect exit code for this. I imagine 2 is fine; as long as it's in the documentation.

Stratus3D:
Thanks!

PR #326, opened 2018-05-31
Fix for #319

### LINKED ISSUE #319: .asdf/lib/utils.sh: line 173: ASDF_DOTNET-CORE_VERSION: bad substitution
### Steps to reproduce

Using asdf v0.4.3

```
$ asdf plugin-add dotnet-core

$ asdf install dotnet-core 2.1.200

$ asdf global dotnet-core 2.1.200

$ dotnet version
/root/.asdf/lib/utils.sh: line 173: ASDF_DOTNET-CORE_VERSION: bad substitution

Welcome to .NET Core!
---------------------
Learn more about .NET Core @ https://aka.ms/dotnet-docs. Use dotnet --help to see available commands or go to https://aka.ms/dotnet-cli-docs.
...
```

### Expected behavior

`bad substitution` error should not occur

### Actual behavior

See above

### Environment

**OS**: Linux Debian Stretch & Fedora

**asdf version**: 0.4.3

### Comment

I was working on a golang-dep plugin and came across this bug. I figgured this was likely due to the `-` hyphen/dash in the plugin name. Dotnet-core was the only official plugin that also has a hyphen/dash so I installed it to confirm.

I believe the issue is due to bash not allowing hyphen/dash in a variable name.

```
$ version_env_var=ASDF_GOLANG-DEP_VERSION
$ echo ${!version_env_var}
bash: ASDF_GOLANG-DEP_VERSION: bad substitution
$ version_env_var=ASDF_GOLANG_DEP_VERSION
$ echo ${!version_env_var}

```

Updating line .asdf/lib/utils.sh:171 `tr '[:lower:]-' '[:upper:]_'` seems to fix the issue but this means setting a version via environment variable will require the user to replace any hyphen/dash with an underscore.

brennanfee:
Fixed #319 by forcing all dashes to underscores in the plugin name.
Shell variables don't allow dash.

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -170,7 +170,7 @@ display_no_version_set() {
 get_version_from_env () {
   local plugin_name=$1
   local upcase_name
-  upcase_name=$(echo "$plugin_name" | tr '[:lower:]' '[:upper:]')
+  upcase_name=$(echo "$plugin_name" | tr '[:lower:]-' '[:upper:]_')
   local version_env_var="ASDF_${upcase_name}_VERSION"
   local version=${!version_env_var}
   echo "$version"


Stratus3D:
Awesome!

PR #328, opened 2018-05-31
Fix another spot for plugin-name env var lookup

brennanfee:
After fixing #319 I wondered if there were any other spots where the same/similar bug existed.  I found one other location at line 126 in utils.sh.


Stratus3D:
Thanks! We should probably write tests for the environment variable logic. And also maybe extract logic into a function.

Issue #330, opened 2018-05-31
Add spacing in front of versions for "asdf list" to enhance readability


brennanfee:
### Steps to reproduce

Run "asdf list"

### Expected behavior

When the lines are written the versions are slightly indented to help make the output more readable.

### Actual behavior

At present all of the lines are written flush with the left of the screen making it more difficult to distinguish between plugins version versions.

### Environment

**asdf version**: v0.5.0

Stratus3D:
Fixed by #331.

It might be worth while to eventual check and see if the output is being printed to screen or piped to another program. The whitespace is not needed when piping this info elsewhere.

PR #331, opened 2018-05-31
Fix for #330 - "asdf list" spacing

brennanfee:
In order to make the output of "asdf list" a bit easier to read, this commit adds two spaces to the front of each version being written so as to indent them under the plugin.

### DIFF
--- a/lib/commands/list.sh
+++ b/lib/commands/list.sh
@@ -26,7 +26,7 @@ display_installed_versions() {
 
   if [ -n "${versions}" ]; then
     for version in $versions; do
-      echo "$version"
+      echo "  $version"
     done
   else
     display_error 'No versions installed'
--- a/test/list_command.bats
+++ b/test/list_command.bats
@@ -18,7 +18,7 @@ teardown() {
   run install_command dummy 1.0
   run install_command dummy 1.1
   run list_command
-  [ "$(echo -e "dummy\n1.0\n1.1")" == "$output" ]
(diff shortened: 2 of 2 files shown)


Stratus3D:
Looks like tests are failing due to these changes. Can you update the tests as well?

brennanfee:
Sure, will do.

brennanfee:
So, I fixed the problem but we are getting a bunch of linting errors that seem unrelated.  The bad thing is I can't reproduce those issues locally.  Is it possible there are differences between the shellcheck available for Linux and the one available for OSX?

brennanfee:
So, looking into the shellcheck issue it looks like OSX brew has updated to the very latest shellcheck (0.5.0) while Ubuntu is still on 0.4.6.  My guess is that the newer one is finding more issues than before.

brennanfee:
@Stratus3D Any recommended course for this?  None of the "new" issues found are in the area I modified.  Should we just wait until the other linting issues are resolved or can we ignore them and push this anyway?

Time permitting I may make an attempt to fix the linting issues primarily.  But time is, as always, limited.  Besides, that would likely need to be in a separate group of PRs.

Stratus3D:
Thanks for fixing the tests. Looks great! I am going to wait on merging this until I get the asdf build fixed (in a separate PR). It does look like the rug was pulled out from under us again with shellcheck on Travis CI.

Stratus3D:
I fixed the build! Merging these changes now.

PR #334, opened 2018-06-07
Fixes #333, there was a unnecessary cd issued (typo?) within a $ subs…

### LINKED ISSUE #333: Throwing "-bash: shell_session_update: command not found"
### Steps to reproduce

1. Install latest version `0.5.0`
2. Open a new login shell

### Expected behavior

```
Last login: Wed Jun  6 18:25:39 on ttys009
 ~ $
```

### Actual behavior

```
Last login: Wed Jun  6 18:25:39 on ttys009
-bash: shell_session_update: command not found
 ~ $
```

### Environment

```
$ asdf --version
v0.5.0

$ bash --version
GNU bash, version 4.4.23(1)-release (x86_64-apple-darwin16.7.0)
Copyright (C) 2016 Free Software Foundation, Inc.

$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G1408

```

rafaelliu:
Fixes: #333

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -7,7 +7,7 @@ else
 fi
 
 export ASDF_DIR
-ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
+ASDF_DIR="$(dirname "$current_script_path" &> /dev/null || exit 1; pwd)"
 
 # Add asdf to PATH
 #
--- a/asdf.sh
+++ b/asdf.sh
@@ -7,7 +7,8 @@ else
 fi
 
 export ASDF_DIR
-ASDF_DIR="$(dirname "$current_script_path" &> /dev/null || exit 1; pwd)"
+ASDF_DIR="$(dirname "$current_script_path")"
+[ -d "$ASDF_DIR" ] || return
 
 # Add asdf to PATH
 #


rafaelliu:
I was being too simplistic on my first fix. Problem is still the same, but my fix killed the check for directory existence.

This uncovered another issue: if `ASDF_DIR` fails, `ASDF_BIN`  is set to `/bin`. So condition will match:

```
[[ ":$PATH:" == *":${ASDF_BIN}:"* ]] && PATH="${PATH//$ASDF_BIN:/}"
```

And all `/bin:` will be stripped from path.

We should probably not fail silently though, but we can't `exit` or we will crash terminal initialization

asdf.sh
```diff
@@ -7,7 +7,8 @@ else
 fi
 
 export ASDF_DIR
-ASDF_DIR="$(cd "$(dirname "$current_script_path")" &> /dev/null || exit 1; pwd)"
+ASDF_DIR="$(dirname "$current_script_path")"
+[ -d "$ASDF_DIR" ] || return
```
Stratus3D: The previous code executed `exit 1` if the path was not a directory. This code just executes `return`, which is not the same thing. I think this should be updated to be `[ -d "$ASDF_DIR" ] || exit 1`.

Stratus3D:
Thanks for the PR @rafaelliu ! I made that one change and then merged it.

asdf.sh
rafaelliu: Thanks for merging back upstream!

The reason for `return` is that, since asdf.sh is being sourced on .bashrc, if we `exit` we will crash user's session and they'll will never be able to spawn a shell.

Agree that `return` would fail silently and it would be better to explicit message user

Issue #275, opened 2018-01-15
Move ASDF install directory


paradox460:
I have ASDF installed using Homebrew. I recently updated to `0.4.1`. Since homebrew versions everything into cellar folders, and 0.4.0 and 0.4.1 are different folders in the cellar, they cannot share installs without hacking (symlinks).

An envar to set the ASDF install directory location would be very welcome.

PR #335, opened 2018-06-10
Install data outside asdf installation directory

josacar:
# Summary

This reads a custom configuration variable `install_type`, that can be set to `user` and changes the installation directory from where`asdf` installation is to `$HOME/.asdf`. 

Fixes: #275 

## Other Information

This fixes the problem that users experience using `asdf` with `brew`

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -17,8 +17,10 @@ ASDF_DIR="$(dirname "$current_script_path")"
 # replace all occurrences - ${parameter//pattern/string}
 ASDF_BIN="${ASDF_DIR}/bin"
 ASDF_SHIMS="${ASDF_DIR}/shims"
+ASDF_USER_SHIMS="${HOME}/.asdf/shims"
 [[ ":$PATH:" == *":${ASDF_BIN}:"* ]] && PATH="${PATH//$ASDF_BIN:/}"
 [[ ":$PATH:" == *":${ASDF_SHIMS}:"* ]] && PATH="${PATH//$ASDF_SHIMS:/}"
+[[ ":$PATH:" == *":${ASDF_USER_SHIMS}:"* ]] && PATH="${PATH//$ASDF_USER_SHIMS:/}"
 # add to front of $PATH
 PATH="${ASDF_BIN}:$PATH"
 PATH="${ASDF_SHIMS}:$PATH"
--- a/asdf.sh
+++ b/asdf.sh
@@ -17,7 +17,7 @@ ASDF_DIR="$(dirname "$current_script_path")"
 # replace all occurrences - ${parameter//pattern/string}
 ASDF_BIN="${ASDF_DIR}/bin"
 ASDF_SHIMS="${ASDF_DIR}/shims"
-ASDF_USER_SHIMS="${HOME}/.asdf/shims"
+ASDF_USER_SHIMS="${ASDF_DATA_DIR:-$HOME/.asdf}"
 [[ ":$PATH:" == *":${ASDF_BIN}:"* ]] && PATH="${PATH//$ASDF_BIN:/}"
 [[ ":$PATH:" == *":${ASDF_SHIMS}:"* ]] && PATH="${PATH//$ASDF_SHIMS:/}"
 [[ ":$PATH:" == *":${ASDF_USER_SHIMS}:"* ]] && PATH="${PATH//$ASDF_USER_SHIMS:/}"
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -23,17 +23,40 @@ asdf_repository_url() {
   echo "https://github.com/asdf-vm/asdf-plugins.git"
 }
 
+get_install_type(){
+  local install_type
+  install_type=$(get_asdf_config_value "install_type")
+
+  echo "$install_type"
+}
+
+asdf_data_dir(){
+  local data_dir
+
+  if [ "$(get_install_type)" = "user" ]; then
+    data_dir="${HOME}/.asdf"
+  else
+    data_dir="$(asdf_dir)"
+  fi
+
+  echo "$data_dir"
+}
+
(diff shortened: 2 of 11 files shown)


josacar:
@stratus3d how does it look to you?

btburton42:
what is the status of this? just updated `asdf` via `brew` and had to copy `plugins` and `install` folders to the new versioned dir. 

would managing asdf outside of homebrew be best for the time being?

mwean:
I would rather use an environment variable than a config flag that doesn't allow arbitrary locations, but at this point, I just need something to work with homebrew... @Stratus3D is there anything with this PR that needs to change before you can approve it?

Stratus3D:
Sorry all, I've been behind on asdf PRs recently. Reviewing this PR now. This is definitely something that is needed.

lib/utils.sh
```diff
@@ -23,17 +23,40 @@ asdf_repository_url() {
...
+  local data_dir
+
+  if [ "$(get_install_type)" = "user" ]; then
+    data_dir="${HOME}/.asdf"
+  else
+    data_dir="$(asdf_dir)"
+  fi
+
+  echo "$data_dir"
+}
```
Stratus3D: I'd rather see this function pull the directory name from an environment variable, and default to something reasonable if it's not set. We could look for `$ASDF_DATA_DIR` and default to `"$HOME/.asdf"` if it is not set. The install type logic may limit us in the future as users may want to customize the directory installations are placed in.

lib/utils.sh
```diff
@@ -42,8 +65,11 @@ list_installed_versions() {
   local plugin_path
   plugin_path=$(get_plugin_path "$plugin_name")
 
+  local install_dir
+  install_dir="$(asdf_data_dir)/installs"
```
Stratus3D: I don't think this variable is needed. This could be combined with line 72 and assigned to `plugin_installs_path`.

lib/utils.sh
Stratus3D: Alternatively we could look in the config file for the path, and allow the user to set it there. However reading the config file would make this function slower, and it's used by the shims, so we want to keep it as fast as possible. I'm leaning towards not putting anything in the config file for this.

mwean:
I'm working on a version based on an env variable. I'll try to push it up later today.

asdf.sh
```diff
@@ -17,8 +17,10 @@ ASDF_DIR="$(dirname "$current_script_path")"
 # replace all occurrences - ${parameter//pattern/string}
 ASDF_BIN="${ASDF_DIR}/bin"
 ASDF_SHIMS="${ASDF_DIR}/shims"
+ASDF_USER_SHIMS="${HOME}/.asdf/shims"
```
Stratus3D: If we switch to pulling from an environment variable we will need to update this code as well. Something along the lines of:

```
ASDF_USER_SHIMS="${ASDF_DATA_DIR:-$HOME/.asdf}"
```

mwean:
I'm unclear why there are function calls like `mkdir -p  "$(asdf_dir)/plugins"` instead of `mkdir -p "${ASDF_DIR}/plugins"`

It seems like `asdf-exec` doesn't have `asdf.sh` sourced, so the variables aren't available? I'm wondering if there's a better way to set up the variables for all cases?

lib/utils.sh
josacar: I was trying to make it more semantic, I can merge it with L72 :+1:

josacar:
@Stratus3D @mwean I just changed the way the asdf_data_dir is sourced

lib/utils.sh
```diff
@@ -23,17 +23,40 @@ asdf_repository_url() {
   echo "https://github.com/asdf-vm/asdf-plugins.git"
 }
 
+get_install_type(){
+  local install_type
+  install_type=$(get_asdf_config_value "install_type")
+
+  echo "$install_type"
+}
```
Stratus3D: This function isn't used anymore.

Stratus3D:
Works on my machine! :shipit:

lib/utils.sh
```diff
@@ -43,12 +66,12 @@ list_installed_versions() {
...
   local plugin_installs_path
-  plugin_installs_path=$(asdf_dir)/installs/${plugin_name}
+  plugin_installs_path="$(asdf_data_dir)/installs/${plugin_name}"
 
   if [ -d "$plugin_installs_path" ]; then
-	# shellcheck disable=SC2045
+  # shellcheck disable=SC2045
     for install in $(ls -d "${plugin_installs_path}"/*/ 2>/dev/null); do
-		basename "$install" | sed 's/^ref-/ref:/'
+    basename "$install" | sed 's/^ref-/ref:/'
```
Stratus3D: Indentation seems off here.

Stratus3D:
I've gone ahead and merged this, but I think there are some things we need to address before tagging a new version with these changes:

* Remove unused get_install_type function
* Document `ASDF_DATA_DIR` environment variable in readme
* Correct indentation for loop in utils.sh
* Evaluate potential issues that could come arise from setting `ASDF_DATA_DIR` after sourcing `asdf.sh` in the shell. `asdf.sh` would see it as unset and default to `~/.asdf`, whereas the asdf commands would see it as set and would look elsewhere. This would result in shims not being on the path, but asdf commands still seeing the installed versions and the shims.

lib/utils.sh
josacar: There are a mix of tabs and spaces :S , maybe my editor messed it up

josacar:
@Stratus3D @mwean let me know if you want me to address those concerns in a new PR

Stratus3D:
@josacar yes feel free, it will probably be a couple days before I am able to come back to this. I opened https://github.com/asdf-vm/asdf/issues/361

PR #340, opened 2018-07-11
Fix issue with Fish shell shim loading script

friedenberg:
Currently, `asdf.fish` uses incorrect arguments for Fish's `contains` and `test` command builtins. 

This fixes the script to use valid arguments patterns.


Stratus3D:
@tuvistavie you are a fish user right? Can you check this out?

tuvistavie:
Oh, the check in the current version is indeed broken, thanks for the fix @friedenberg :smiley:

PR #341, opened 2018-07-25
Allow for at least one space when printing versions in asdf current

antonafMSFT:
# Summary

When `asdf current` prints, plugins with long names or long version strings lose the space between plugin name and version or plugin version and the note ("(set by ...)").
For example, for the `java` plugin:
`java           oracle-8.181(set by PATH_HERE/.tool-versions)`

Personally, and this is a matter of preference, `asdf current` or other tools shouldn't attempt to pretty print and should instead produce machine-readable output that would be easy for the user to format.
Should this opinion be adopted by the project, this PR would devolve to simply adding a simple space between the two pieces of data being `printf`ed.

### DIFF
--- a/lib/commands/current.sh
+++ b/lib/commands/current.sh
@@ -19,14 +19,18 @@ plugin_current_command() {
     printf "%s\\n" "$(display_no_version_set "$plugin_name")"
     exit 126
   else
-    printf "%-8s%s\\n" "$version" "(set by $version_file_path)"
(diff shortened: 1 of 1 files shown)


Stratus3D:
It looks ShellCheck is complaining about the `printf` calls on OSX and failing the build. Can you check the calls and make sure they work the same on Linux and OSX?

danhper:
I think this is fixed in recent versions.
Please open a new PR if it is not.

PR #348, opened 2018-08-24
Fix fish shell completions

tuvistavie:
# Summary

It seems that fish recently changed how it handles white spaces in completions, so I was getting weird completions.
This should hopefully fix it.  

## Other Information

If anybody with on macOS using fish could check everything is working fine, it would be great.
Thank you.

### DIFF
--- a/completions/asdf.fish
+++ b/completions/asdf.fish
@@ -26,9 +26,15 @@ function __fish_asdf_arg_at -a number
     echo $cmd[$number]
 end
 
+function __fish_asdf_list_versions -a plugin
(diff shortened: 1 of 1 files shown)


Stratus3D:
I think this may have been caused by #331. `asdf list` will indent versions now, so lines will now start with whitespace.

tuvistavie:
@Stratus3D Oh, I did not realize this change to `asdf list` was recent. Either way, the completion for `plugin-add` was horrible due to changes on the fish side, so this should fix both issues.
Thanks for the feedback, I'll merge!

PR #355, opened 2018-09-15
Make data directory configurable

mwean:
This was inspired by #335, but uses an optional variable instead of a config value to specify a data directory.

# Summary

This adds the `ASDF_DATA_DIR` variable (which defaults to `~/.asdf`), that holds data that should persist across versions of asdf (plugins, installs, shims, etc.).

Fixes: #275


mwean:
We should probably add more tests to this, but I'm not 100% sure how best to attack them and didn't want to make the effort unless we want to move forward with this direction.

Stratus3D:
I already merged the other PR for this feature since it had been updated to use the environment variable as well. Is there anything else here that would still be good to have in master?

PR #358, opened 2018-09-17
Make help text reflect that <name> and <version> are optional in `asdf install` command

nate:
The help text shows `<name>` and `<version>` as required options to `asdf install` when they, as a group, are optional. This pull request fixes that.


Stratus3D:
Thanks! I can't believe I missed this.

PR #360, opened 2018-09-18
Fix reshim to generate shims only for executable files and not directories.

teixeiradiego:
# Summary

Change reshim to generate shims only for executable files and not directories. 

Problem: When you have sub-directories inside bin path directory, it uses to create shims for the executable file and for all directories inside the path and it causes the error: "No such command in [version] of [plugin-name]"

Fixes: No open issue found with this error description.

### DIFF
--- a/lib/commands/reshim.sh
+++ b/lib/commands/reshim.sh
@@ -143,7 +143,7 @@ generate_shims_for_version() {
       # because just $executable_file gives absolute path; We don't want version hardcoded in shim
       local executable_path_relative_to_install_path
       executable_path_relative_to_install_path="$bin_path"/$(basename "$executable_file")
-      if [ -x "$executable_file" ]; then
+      if [[ (-f "$executable_file") && (-x "$executable_file") ]]; then
         write_shim_script "$plugin_name" "$version" "$executable_path_relative_to_install_path"
       fi
     done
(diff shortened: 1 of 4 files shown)


Stratus3D:
Can you write a unit test to exercise the buggy behavior? I didn't know this was an issue.

teixeiradiego:
> Can you write a unit test to exercise the buggy behavior? I didn't know this was an issue.

It`s possible to exercise this buggy behavior just changing the install script of dummy plugin like this:

`#!/usr/bin/env bash

mkdir -p "$ASDF_INSTALL_PATH"
env > "$ASDF_INSTALL_PATH/env"
echo "$ASDF_INSTALL_VERSION" > "$ASDF_INSTALL_PATH/version"

# create the dummy executable
mkdir -p "$ASDF_INSTALL_PATH/bin"
cat <<EOF > "$ASDF_INSTALL_PATH/bin/dummy"
#!/usr/bin/env python

import sys

print r"This is Dummy ${ASDF_INSTALL_VERSION}! %s %s" % (sys.argv[2], sys.argv[1])

EOF
chmod +x "$ASDF_INSTALL_PATH/bin/dummy"
mkdir "$ASDF_INSTALL_PATH/bin/python"`

I changed the dummy to be a python script and created a python folder inside. When install plugin, it generates shims for "dummy" and for the directory "python" then it causes problem to use the system installed python, trying to use the shim that points to an non existing script, but a directory.

Stratus3D:
Thanks!

PR #364, opened 2018-09-24
Fix user shims stored now in ASDF_DATA_DIR

josacar:
# Summary

Fixes shims PATH stored under ASDF_DATA_DIR


Stratus3D:
Thanks!

josacar:
Maybe we can remove old shims, as now they are always in data dir

Stratus3D:
Now that I think about this, the user is going to have to restart all their shell sessions after upgrading to the version that includes these changes. I think we can remove the old shims, but either way the user is going to have to restart their shells after the upgrade.

PR #365, opened 2018-09-27
Document asdf data dir

### LINKED ISSUE #361: Document ASDF_DATA_DIR
* Remove unused get_install_type function
* Document `ASDF_DATA_DIR` environment variable in readme
* Correct indentation for loop in utils.sh
* Evaluate potential issues that could come arise from setting `ASDF_DATA_DIR` after sourcing `asdf.sh` in the shell. `asdf.sh` would see it as unset and default to `~/.asdf`, whereas the asdf commands would see it as set and would look elsewhere. This would result in shims not being on the path, but asdf commands still seeing the installed versions and the shims.

Hopefully I'll get around to addressing these points this week, and I'll tag a new version next weekend. But feel free to address any of these things if you like.

_Originally posted by @Stratus3D in https://github.com/asdf-vm/asdf/pull/335#issuecomment-421879751_

josacar:
# Summary

Fixes #361

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ legacy_version_file = yes
 
 * `ASDF_CONFIG_FILE` - Defaults to `~/.asdfrc` as described above. Can be set to any location.
 * `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` - The name of the file storing the tool names and versions. Defaults to `.tool-versions`. Can be any valid file name.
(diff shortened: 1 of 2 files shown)


Stratus3D:
Thanks @josacar ! This looks great!

Stratus3D:
Actually it might be better to move the new documentation to under the `## Environment Variables` heading since it is in regards to an environment variable, and because _hopefully_ it's a setting most asdf users won't have to customize so it should be lower in the readme.

josacar:
Ok, I'll move it now

josacar:
I've rebased it and squashed the change that you suggested

Stratus3D:
Thanks!

PR #371, opened 2018-10-10
commands/current: add space before parenthesis

phunehehe:
It seems to have been lost in 88d47bbd69dc49a601fda84c87d6380d15244daa.
Probably by accident?

I also pulled the literal bits into the format string. It looks more
readable to me that way. I'm happy to remove that if it's not desired.

### DIFF
--- a/lib/commands/current.sh
+++ b/lib/commands/current.sh
@@ -19,7 +19,7 @@ plugin_current_command() {
     printf "%s\\n" "$(display_no_version_set "$plugin_name")"
     exit 126
   else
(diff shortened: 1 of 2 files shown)


Stratus3D:
I think that is a good change though, the raw values are kept separate from the format string.

phunehehe:
Glad you like the idea. I'm going to fix the test errors. Maybe I'll add a new test for long version names.

phunehehe:
@Stratus3D how about now?

Stratus3D:
Thanks!

PR #376, opened 2018-10-17
Add the default asdf_data_dir to path

m-nagae:
# Summary

If fish shell users install the asdf by homebrew, can't use commands in $HOME/.asdf/shims.
So, I added the path.

## Other Information

information here.

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -1,6 +1,6 @@
 #!/usr/bin/env fish
 
-set -l asdf_data_dir (dirname (status -f))
+set -l asdf_data_dir $HOME/.asdf (dirname (status -f))
 
 # we get an ugly warning when setting the path if shims does not exist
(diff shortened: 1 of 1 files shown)


Stratus3D:
What happens if `$ASDF_DATA_DIR` isn't the default path? What if they have customized it themselves? Would this still work?

m-nagae:
Maybe doesn't work that case. I'll fix.

m-nagae:
@Stratus3D I fixed. Would you please check it?

grahamannett:
Yeah, this will fix asdf not working at all with asdf.  Im not sure what the ideal layout is but brew putting everything in /usr/local/opt/asdf and asdf putting the installs/plugins/shims/etc in ~/.asdf by default seems like it could be bad.  this patch fixed it but should the asdf stuff be managed and configureable by the homebrew formula?

Issue #274, opened 2018-01-14
Documentation update


jthegedus:
Hey all maintainers and contributors, thanks for the great tool!

I like the challenge of writing clear and concise docs and in an effort to learn this tool better I thought I would contribute to the docs.

The Contribution.md file states:

>The wiki exists but is in a state of disrepair. If you see something that can be improved please submit a pull request or edit the wiki.

I *like* projects that use the Wiki page for GitHub as instructions are in a single place and not spread across GitHub and some docs site. Having said that, there are drawbacks to using GitHub's Wikis and many pros to a separate docs site. So before starting, I would like to get an idea of your intentions/wants for the docs and list some of my thoughts on doc layout and docs from other projects.

### Docs in README
* docs immediately seen on GitHub. Most people look here first for docs.
* in sync with the main repo. Easier to maintain docs coverage of new feats or bug fixes as they're added.
* README becomes very large as commands, options etc increase
* some repos use the README to link to the docs and just explain how to contribute here.

### Docs in Wiki
* can be easily linked from the top of the README
* removes a lot of noise from README so README can be used to explain where to find info - [Prettier's README](https://github.com/prettier/prettier) is the best example of this IMO.
* do you include Contribution guidelines in Wiki or in repo/contributing.md?
* where does "Development" guidelines go? Do they get merged with Contribution guidelines?

### Docs on website (say asdf.io)
SAME as above 'Docs in Wiki" except:
* Docs managed within the same repo as the tool and therefore easy to keep in sync 
* Easy to brand
* Easier to add docs searching
* Renders nicely everywhere (GitHub Wikis are not great on mobile)

### My suggestions
* the README should be short and to the point, like the [Prettier README](https://github.com/prettier/prettier).
* IMO a Docs website is the way to go if you maintain it within the same repo as the tool so you can keep the tool and docs in sync. I'd suggest, and happily create a https://www.gatsbyjs.org/ site for this. 
**EDIT** [
I know this seems like overkill but this would go along way with persuading new users. Especially if combined with gifs showing the perf of automatic version switching etc
]
* If you prefer the option where the docs are in a separate repo from the tool, then the Wiki would require less work to get going than creating a docs site.

@HashNuke
@tuvistavie
@Stratus3D
@vic

tuvistavie:
@jthegedus Thank you very much for the issue.
While I personally do not really have time to help with a docs website, I am open to the idea.

jthegedus:
[...]

PR #387, opened 2018-10-30
Document shims

### LINKED ISSUE #215: how to source scripts from asdf?
I'm using Python's Virtualenvwrapper package, installed with python 3.5.2 via asdf, and I'm having an issue with properly sourcing the `.virtualenvwrapper.sh` file.

Essentially, `~/.asdf/shims/virtualenvwrapper.sh` is just a proxy that runs an `exec` line, as so:
```shell
exec /Users/nate/.asdf/bin/private/asdf-exec python bin/virtualenvwrapper.sh "$@"
```
however, because I don't want to _execute_ the shell script, but rather source it, attempting to run the command simply causes my terminal to exit prematurely.

My current workaround is to do
```shell
. $(asdf where python $(asdf current python))/bin/virtualenvwrapper_lazy.sh
```

It works, but it's a bit hacky. Is there a better way to do this?

mig4:
# Summary

Explain how shims work and how they're used to provide executables to
the environment. Also explain how to use scripts that are meant to be
sourced instead of executed.

Closes #215

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,38 @@ The version format is the same supported by the `.tool-versions` file.
 ASDF_ELIXIR_VERSION=1.4.0 mix test
 ```
 
+### Shims
+
+When asdf installs a package it creates shims for every executable program in
+that package in a `$ASDF_DATA_DIR/shims` directory (default `~/.asdf/shims`).
+This directory being on the `$PATH` (by means of `asdf.sh` or `asdf.fish`) is
+how the installed programs are made available in the environment.
+
+The shims themselves are really simple wrappers that `exec` a helper program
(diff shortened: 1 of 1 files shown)


Stratus3D:
This is great! Thanks for taking the time to write this up! I'm going to merge this now, but I think we may be eventually splitting things up and moving a lot of these details out of the readme and into other files.

Stratus3D:
This is related to https://github.com/asdf-vm/asdf/issues/274

mig4:
Sure, yeah I've seen there's a plan to split the docs, I'm happy enough to follow up later once that's started and help move things into right place. Thanks for merging! :+1:

PR #388, opened 2018-10-30
Optimise listing all plugins

mig4:
# Summary

When checking if a plugin is installed (to display a flag) it is not
necessary to loop through all installed plugins, since we already have
a name we can check directly.

Also expand test case to test this code path too.


Stratus3D:
Awesome! This is a nice performance improvement! I wish we had benchmarking for asdf. I'm sure there would be a noticeable difference with this change.

PR #389, opened 2018-10-30
Make where command default to current version

mig4:
# Summary

Allow not specifying the version on command line in which case fall back
to looking it up.

Also add tests for the `where` command.


Stratus3D:
Awesome! Thanks for taking the time to write tests for that command!

PR #392, opened 2018-10-31
Fix fish shell init snippet

mig4:
# Summary

- add `bin` and `shims` directories from asdf installation to PATH
- add `shims` directory from `$ASDF_DATA_DIR` (default `$HOME/.asdf`) to
  PATH
- don't create directories in init script (directories are only added to
  PATH if they exist, besides there is no warning when a directory does
  not exist (perhaps it was a bug that was fixed))

This brings the fish snippet closer to the one for sh.

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -1,13 +1,15 @@
 #!/usr/bin/env fish
 
-set -l asdf_data_dir (dirname (status -f))
+set -l asdf_dir (dirname (status -f))
+set -l asdf_data_dir (
+  if test -n "$ASDF_DATA_DIR"; echo $ASDF_DATA_DIR;
+  else; echo $HOME/.asdf; end)
 
-# we get an ugly warning when setting the path if shims does not exist
-mkdir -p $asdf_data_dir/shims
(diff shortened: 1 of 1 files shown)


danhper:
Hi, thanks for the PR.
This would break compatibility with Ubuntu xenial, as the default fish version on it is 2.2 while `string` was introduced in 2.3.
I don't mind changing the init script but we should at least keep it compatible with 2.2, we don't want to break existing setups.

asdf.fish
```diff
@@ -1,13 +1,15 @@
 #!/usr/bin/env fish
 
-set -l asdf_data_dir (dirname (status -f))
+set -l asdf_dir (dirname (status -f))
+set -l asdf_data_dir (
+  if string length -q -- $ASDF_DATA_DIR; echo $ASDF_DATA_DIR;
```
danhper: Please use `test -n "$ASDF_DATA_DIR"` instead of `string length`

mig4: done

danhper:
Great, thank you!

PR #398, opened 2018-11-09
Fix README typo of maintainer Daniel Perez's username

jlw:
# Summary

While updating the README in #397, I noticed that the link text for Daniel Perez did not match the href, so I fixed it as well.

_Would you prefer one commit with a number of tweaks to the README, or accept/reject each on it's own?_


danhper:
Oh, good catch. I did not realize. Thank you!

> Would you prefer one commit with a number of tweaks to the README, or accept/reject each on it's own?

Unless there is something controversial that might not be merged, a single PR would be just fine.
Thanks for helping out!

PR #403, opened 2018-11-16
support .tool-versions content even if it does not end in a newline

burnettk:
# Summary

````
# works
echo "nodejs 8.12.0" > .tool-versions

# does not work
echo -n "nodejs 8.12.0" > .tool-versions
````

I'm really not sure how we ended up with a .tool-versions file without a trailing newline, but it caused a significant amount of confusion. :)

## Other Information

Implementation for fix borrowed from https://unix.stackexchange.com/questions/418060/read-a-line-oriented-file-which-may-not-end-with-a-newline. I'm hoping the existing test coverage will prevent a regression.

Thanks for asdf-vm; it rocks!

### DIFF
--- a/lib/commands/install.sh
+++ b/lib/commands/install.sh
@@ -40,7 +40,7 @@ install_local_tool_versions() {
   local asdf_versions_path
   asdf_versions_path=$(find_tool_versions)
   if [ -f "${asdf_versions_path}" ]; then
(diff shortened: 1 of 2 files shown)


Stratus3D:
Awesome! Thanks so much for taking the time to write a test @burnettk !

PR #404, opened 2018-11-16
Silence errors during tab completion

rliebz:
# Summary

Currently, when a message is printed to `stderr` during tab completion, it appears inline, causing an awkward user experience. This is easily reproducible by trying to tab complete with a fresh clone of `asdf` (no plugins installed):

asdftaberror

The fix redirects all `stderr` to `/dev/null` during tab completion, so if there is an error when trying to suggest completions, `asdf` will simply suggest nothing rather than report out a message.

## Other Information

I think this is the most user-friendly approach, but we could also remove the "Oohes nooes" message from `plugin-list` to silence an expected error, while still reporting unexpected errors during tab completion, such as:
```
asdf install golang <TAB> /root/.asdf/plugins/golang/bin/list-all: line 34: curl: command not found
```

### DIFF
--- a/completions/asdf.bash
+++ b/completions/asdf.bash
@@ -8,7 +8,7 @@ _asdf () {
   local prev
   prev=${COMP_WORDS[COMP_CWORD-1]}
   local plugins
-  plugins=$(asdf plugin-list | tr '\n' ' ')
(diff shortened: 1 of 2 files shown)


danhper:
This is indeed better, thank you very much for the PR!

Issue #402, opened 2018-11-12
System Version Broken


mwean:
### Steps to reproduce

```bash
> asdf plugin-add ruby
> asdf local ruby system
> ruby -v
```

### Expected behavior

Print out the system ruby version

### Actual behavior

Hangs indefinitely

### Environment

**OS**: macOS 10.13.6

**asdf version**: 0.6.0

I tracked the issue down to the `get_executable_path` function, where it removes the shim directory from the path and tries to find the executable if the version is `system`:

https://github.com/asdf-vm/asdf/blob/dab04107ff88a34d95b09a44a94320d27a2af470/lib/utils.sh#L230-L231

The problem is it's only removing `ASDF_DIR` and not `ASDF_USER_SHIMS`. I was going to make a PR just adding that, but it seems really weird that we have both defined. I would expect there to be one shims directory, either defined by the user or a default. What do you think?

Stratus3D:
You are correct. There should only be one shims directory. My guess is that line should be `...s|$ASDF_DATA_DIR/shims||g...`. This is likely a bug.

mwean:
Sorry, I totally forgot to include the main point of the issue:

https://github.com/asdf-vm/asdf/blob/dab04107ff88a34d95b09a44a94320d27a2af470/asdf.sh#L19-L20

So we have a variable for _both_ shims directories, which seems wrong to me. I would expect to have one, which would be used here.

mwean:
Hmm I hadn't thought about compatibility with earlier versions... On the other hand, it seems like a lot of things would be broken already if most of the code is using the new variable. I would prefer just getting rid of the old one everywhere.

Stratus3D:
@rliebz support for the new `asdf_data_dir` feature is not backwards compatible, as there was no equivalent functionality in older versions. I think `$(asdf_data_dir)` is what we want.

PR #406, opened 2018-11-26
Make get_executable_path use `ASDF_DATA_DIR`

Stratus3D:
# Summary

Use `ASDF_DATA_DIR` in `get_executable_path` function if it is set. Otherwise fall back to `$HOME/.asdf`.

Fixes: #402

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -228,7 +228,7 @@ get_executable_path() {
   check_if_version_exists "$plugin_name" "$version"
 
   if [ "$version" = "system" ]; then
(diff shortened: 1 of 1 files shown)


mwean:
This is good, but shouldn't we either use or get rid of these variables:

https://github.com/asdf-vm/asdf/blob/dd5a8036f302b49bcfa7b4319eeaacfe2065eb5a/asdf.sh#L19-L20

Stratus3D:
Oh yes, that should be fixed as well. Only user shims should be needed.

PR #408, opened 2018-11-27
Remove unused asdf shims directory

Stratus3D:
# Summary

Only one shims directory needs to be added to `PATH`.

Related to #406

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -16,14 +16,11 @@ ASDF_DIR="$(dirname "$current_script_path")"
 # if in $PATH, remove, regardless of if it is in the right place (at the front) or not.
 # replace all occurrences - ${parameter//pattern/string}
 ASDF_BIN="${ASDF_DIR}/bin"
-ASDF_SHIMS="${ASDF_DIR}/shims"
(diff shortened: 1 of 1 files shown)


danhper:
Would this work if the user has `ASDF_DIR` set but `ASDF_DATA_DIR` is not set?
I not I guess we should first check for `$ASDF_DATA_DIR` then for `$ASDF_DIR` and default to `$HOME/.asdf`

Stratus3D:
Yes I think so, `$ASDF_DATA_DIR` is set to `ASDF_USER_SHIMS="${ASDF_DATA_DIR:-$HOME/.asdf}/shims"` so it will default to `$HOME/.asdf`, which is what we want when `$ASDF_DATA_DIR` is not set by the user.

rliebz:
The way I've installed `asdf` on my machine has `ASDF_DIR` set (to somewhere other than `$HOME/.asdf`) and `ASDF_DATA_DIR` not set at all. For me, `${ASDF_DIR}/shims` was never actually created on my machine, only `$HOME/.asdf/shims`. 

This change appears to work fine after some local testing.

PR #411, opened 2018-12-02
Fix issues with update command

rliebz:
# Summary

`asdf update` should be using `ASDF_DIR`, the directory of the git repo, rather than `ASDF_DATA_DIR`, the location of various data.

The following changes are included as a result of this fix:
- `cd` into `${ASDF_DIR}` instead of `$(asdf_data_dir)` before performing an update
- Make sure that the directory is part of a `git` repo before performing an update. This gives a better error message, and works for homebrew installations as well
- Add a test case for the above behavior
- Fix test setup to work for certain `asdf` installations. Details below

## Other Information

I have a somewhat unique setup in that I have `asdf` installed as a git submodule in my dotfiles, but I keep the `ASDF_DATA_DIR` separately as the default of `$HOME/.asdf`. The `update_command` tests were failing for me unrelated to this change because they use the `.git` directory for a submodule, which doesn't work the same way as a fresh `.git` directory. The current test set up is more system agnostic.

### DIFF
--- a/lib/commands/update.sh
+++ b/lib/commands/update.sh
@@ -2,9 +2,9 @@ update_command() {
   local update_to_head=$1
 
   (
-  cd "$(asdf_data_dir)" || exit 1
+  cd "${ASDF_DIR}" || exit 1
 
-  if [ -f asdf_updates_disabled ]; then
+  if [ -f asdf_updates_disabled ] || ! git rev-parse --is-inside-work-tree &> /dev/null; then
     echo "Update command disabled. Please use the package manager that you used to install asdf to upgrade asdf."
   else
     do_update "$update_to_head"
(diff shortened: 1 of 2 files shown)


Stratus3D:
Thanks for taking the time to figure this all out @rliebz ! This looks great. Do you know if this issue is affecting homebrew users? I'd guess it would be.

rliebz:
Yeah, on my machine I get the following error:
```
fatal: not a git repository (or any of the parent directories): .git
```

I looked through the git blame for `asdf_updates_disabled`—my guess is that the intention was for homebrew to drop a file with that name in with the repo during installation, but from what I can tell, that doesn't actually happen. I couldn't find any reference in this repo or the homebrew formula to creating this file, so it seems likely it was just missed.

Since the fix in this PR reports the same error for any non-git installation, which includes homebrew, this should be the fix. In theory we might not need the `asdf_updates_disabled` check anymore, but I figured it could be better for backward compatibility to keep it just in case I missed anything, and there isn't really a downside apart from the maintenance cost of an extra piece of logic in the code.

PR #419, opened 2018-12-28
Continue list even when version is not found

Raphx:
# Summary

Previous implementation exits abruptly when no version is installed for
a plugin. This prevented the list command from listing the versions for
some other plugins.

This commit allows list command to continue executing even when no
version is installed for some plugins.

## Other Information

Example, plugin a with 1.0, b with none, and c with 2.0.

Previous implementation:

```
$ asdf list

a
  1.0
b
No versions installed
```

After commit changes:

```
$ asdf list

a
  1.0
b
No versions installed
c
  2.0
```

### DIFF
--- a/lib/commands/list.sh
+++ b/lib/commands/list.sh
@@ -30,6 +30,5 @@ display_installed_versions() {
     done
   else
     display_error 'No versions installed'
-    exit 1
   fi
 }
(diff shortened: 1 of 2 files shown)


Stratus3D:
Thanks @Raphx ! Good catch, these changes look good.

Issue #425, opened 2019-01-10
Homebrew setup instructions are incorrect


d4rky-pl:
The setup instructions are incorrect - the Homebrew formula installs things a bit differently.
The paths (in case of Bash) are:

```
. /usr/local/opt/asdf/asdf.sh
. /usr/local/opt/asdf/etc/bash_completion.d/asdf.bash
```

ttilberg:
I tried installing asdf for the first time tonight, and it did not work (as this issue suggests). I was digging into it on my system and left some notes for someone who is more fluent in brewing (and/or this project).

https://github.com/asdf-vm/asdf/issues/428#issuecomment-455022420

Hopefully this is helpful for someone.

pikeas:
Same issue here, after running `brew install asdf` on Mojave 10.14.2, and before sourcing either script, running `asdf` outputs the following:

```
$ asdf
cat: /usr/local/VERSION: No such file or directory
version:

cat: /usr/local/help.txt: No such file or directory
```

Oscarteg:
The issue @ttilberg linked to has helped me solve the same problem. 

TLDR: put `export ASDF_DIR=$(brew --prefix asdf)` in your `.bash_profile` or `.zshrc` or whatever you are using.

dboune:
**Note**: If a homebrew user already has bash completion configured for homebrew (see: [configuring-completions-in-bash](https://docs.brew.sh/Shell-Completion#configuring-completions-in-bash)), or has properly installed the `bash-completion` or `bash-completion@2` package, sourcing the completions manually is unnecessary.

danhper:
Is this still an ongoing issue?
If so, could a brew user be kind enough to send a PR with working instructions, please? 
Thank you very much!

dboune:
@danhper The way the docs are structured focuses on installation via Git. If I were to update the docs for this purpose, I'd be inclined to also add tabs to the **Install asdf-vm** section with Homebrew specific installation instructions, so that the instructions in **Add to your Shell** have a bit more context available.

Installing via Git would naturally not require Homebrew specific instructions, whether or not Homebrew is used.

Thoughts?

danhper:
@dboune I agree with this. I suppose installing via Homebrew is a common choice for macOS user so having a dedicated section is, I think, a good idea.
However, it seems that docsify does not support nested tabs so I am not sure what is the best way to present this cleanly.

dboune:
[...]

PR #426, opened 2019-01-12
Add instructions for brew install

heydonovan:
# Summary

This will add instructions on how to install asdf when using the fish shell and when it is installed using homebrew instead of git. Fish is now at 3.0, which means it now supports && as well.

Fixes: https://github.com/asdf-vm/asdf/issues/425

## Other Information

Do we still need instructions on bash? It also doesn't look like the `asdf.fish` file get executed upon install, as I'm still unable to use the shims. This is on a fresh install too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,20 @@ Depending on your OS and shell, run the following:
 
 * Fish:
 
+  If installed using `git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.6.2`:
+
   ```bash
   echo 'source ~/.asdf/asdf.fish' >> ~/.config/fish/config.fish
-  mkdir -p ~/.config/fish/completions; and cp ~/.asdf/completions/asdf.fish ~/.config/fish/completions
+  mkdir -p ~/.config/fish/completions && cp ~/.asdf/completions/asdf.fish ~/.config/fish/completions
   ```
+  
+  If installed using `brew install asdf`:
+  
+  ```bash
+  echo 'source /usr/local/share/fish/vendor_completions.d/asdf.fish' >> ~/.config/fish/config.fish
+  mkdir -p ~/.config/fish/completions && cp /usr/local/share/fish/vendor_completions.d/asdf.fish ~/.config/fish/completions
+  ```
+  
 Restart your shell so that PATH changes take effect. (Opening a new terminal
 tab will usually do it.)
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Depending on your OS and shell, run the following:
   If installed using `brew install asdf`:
   
   ```bash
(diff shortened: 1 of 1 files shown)


heydonovan:
I'm also unable to get the shim working correctly, even after sourcing it. Any ideas?

```
$ source /usr/local/share/fish/vendor_completions.d/asdf.fish
$ echo $PATH
/usr/local/opt/coreutils/libexec/gnubin /usr/local/opt/grep/libexec/gnubin /usr/local/bin /usr/bin /bin /usr/sbin /sbin
```

jthegedus:
@heydonovan sorry to be the bearer of bad news, but we released the docs site and changed how we use the README so these changes will need to be moved.

The README is strictly for introducing the project to newcomers.

Documentation to install and use the site is stored in the `docs/` folder. Check out the site https://asdf-vm.github.io/asdf/#/ to see where it would be appropriate to add this change.

Additionally, it would be great to get another macOS user to validate your changes before merging this as I have erroneously added Homebrew instructions before to others peril :disappointed: (lack of an Apple machine)

d4rky-pl:
@jthegedus I don't use Fish shell. I've reported a similar problem but with Bash. Just double-checked on a fresh Mojave machine that never had asdf installed. The paths I've reported in #425 are still correct.

d4rky-pl:
The check I did right now was on zsh but the paths are identical (the only difference was `bash_profile` vs `zshrc`)

README.md
```diff
@@ -59,10 +59,20 @@ Depending on your OS and shell, run the following:
...
   echo 'source ~/.asdf/asdf.fish' >> ~/.config/fish/config.fish
-  mkdir -p ~/.config/fish/completions; and cp ~/.asdf/completions/asdf.fish ~/.config/fish/completions
+  mkdir -p ~/.config/fish/completions && cp ~/.asdf/completions/asdf.fish ~/.config/fish/completions
+  ```
+  
+  If installed using `brew install asdf`:
+  
+  ```bash
+  echo 'source /usr/local/opt/asdf/asdf.fish' >> ~/.config/fish/config.fish
+  mkdir -p ~/.config/fish/completions && cp /usr/local/share/fish/vendor_completions.d/asdf.fish ~/.config/fish/completions
```
danhper: Completions are not installed automatically when using brew? I don't know who manages the formula but it seems worth fixing.
Either way, I would go with a symlink rather than a copy, otherwise, completions will not get updated when brew updates asdf's version.

heydonovan:
It might be better to fix this issue upstream. Then, it truly is one command to install:

https://github.com/Homebrew/homebrew-core/blob/master/Formula/asdf.rb

I'll try and work on this over the weekend, now that I'm using `asdf` at work again.

README.md
cesarandreu: Copying or symlinking completions isn't required.

danhper: Great, thanks for confirming this.

Issue #431, opened 2019-01-16
Shims dont work if same executable is provided by many plugins


vic:
### Steps to reproduce

Install two tools that create shims for the same executable. For example both `nodejs` and `graalvm` provide their respective `npm` executable. Same should happen with `java` and `graalvm`.

In project A:

```shell
cd A
echo graalvm 1.0.0-rc5 > .tool-versions
asdf install
```

In project B:

```shell
cd B
echo nodejs 10.7.0 > .tool-versions
asdf install
```

### Expected behavior

In project A, executing `npm` should use the `graalvm` executable.
In project B, executing `npm` should use the `nodejs` executable.


### Actual behavior

In project B, executing `npm` fails with asdf complaining that no version is set for `graalvm`.

This is because the `npm` shim was created first for the `graalvm` plugin. A horrible workaround was to delete my `~/.asdf/shims/{npm,node,js}` and `asdf reshim nodejs`. But this would be needed everytime I switch between project A and B.

### Proposed solution

Shims should determine which tool I'm trying to invoke.

Currently the shims we generate have the following lines:

`cat ~/.asdf/shims/npm`                                                                                                                                                                                                              
```shell
#!/usr/bin/env bash
# asdf-plugin: graalvm
# asdf-plugin-version: 1.0.0-rc7
# asdf-plugin-version: 1.0.0-rc5
exec ~/.asdf/bin/private/asdf-exec graalvm bin/npm "$@"
```

Note that `asdf-exec` takes `graalvm` as the first argument. This prevents the same shim to work with other tools like `nodejs`.

I'm proposing we change the meta comments `asdf-plugin-version` to something like:

```shell
#!/usr/bin/env bash
# asdf-plugin: graalvm 1.0.0-rc7
# asdf-plugin: graalvm 1.0.0-rc5
# asdf-plugin: nodejs 10.7.0
exec ~/.asdf/bin/private/asdf-exec npm "$@"
```

And `asdf-exec` no longer taking a tool-name as first-arg.

When the current project (like `B`) specifies a tool version (like nodejs-10.7.0, but no graalvm), then `asdf-exec` should find out that it needs to execute nodejs's npm binary.


### Environment

**OS**: OSX

**asdf version**: master

vic:
Should be solved by #432

Stratus3D:
I like the solution you have proposed here. Great work @vic!

PR #432, opened 2019-01-18
New shim metadata to allow many plugins with same executable names

vic:
Features

* New shim version meta-data allows shims to not depend on a particular plugin
  nor on its relative executable path (#431)
  Upgrading requires shim re-generation and should happen automatically
  by `asdf-exec`:
    `rm -rf ~/.asdf/shims/` followed by `asdf reshim`
* Added lots of tests for shim execution.
  We now make sure that shim execution obeys plugins hooks like
  `list-bin-paths` and `exec-path`.
* Shim exec is now performed by a new `bin/private/asdf-tool-exec` that might
  be faster for most common use case: (versions on local .tool-versions file)
  but fallbacks to slower `get_preset_version_for` which takes legacy formats
  into account.
* Shim exec recommends which plugins or versions to set when command is
  not found.

Fixed Bugs

* Allow many plugins to provide shims with same executable name (#431)

### DIFF
--- a/bin/private/asdf-exec
+++ b/bin/private/asdf-exec
@@ -1,54 +1,16 @@
 #!/usr/bin/env bash
 
+# remove this asdf-exec file when v0.7 becomes the stable version
+echo "asdf is self upgrading shims to new asdf-tool-exec"
+
+asdf_dir="$(dirname "$(dirname "$(dirname "$0")")")"
 # shellcheck source=lib/utils.sh
-source "$(dirname "$(dirname "$(dirname "$0")")")/lib/utils.sh"
+source "$asdf_dir/lib/utils.sh"
+rm "$(asdf_data_dir)"/shims/*
+"$asdf_dir"/bin/asdf reshim
+shim_name=$(basename "$2")
 
-plugin_name=$1
-executable_path=$2
+echo "asdf: now running $shim_name"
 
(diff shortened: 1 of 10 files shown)


vic:
Ok, I've been working on this feature lately and it implied a major refactor on how the shims metadata is kept in shims. 

I've been using this branch locally for the last three days (for my development needs) and everything seems to be ok (at least for my nodejs, elixir, scala, rust installations). 

It would be great if someone could help testing this branch. Added lots of tests for the new `bin/private/asdf-tool-exec` and everything should be pretty much smooth and way faster.

bin/private/asdf-exec
```diff
@@ -1,54 +1,16 @@
...
+rm "$(asdf_data_dir)"/shims/*
+"$asdf_dir"/bin/asdf reshim
+shim_name=$(basename "$2")
 
-        # unset everything, we don't want to pollute
-        unset ASDF_INSTALL_TYPE
-        unset ASDF_INSTALL_VERSION
-        unset ASDF_INSTALL_PATH
-      fi
+echo "asdf: now running $shim_name"
```
Stratus3D: This could be problematic if it happened when a shimmed command got executed by a users shell script, and the script relied on the output of the command. This extra text could break exist shell code in other scripts.

PR #434, opened 2019-01-19
user configurable pre-post command hooks

vic:
Suppose a `foo` plugin is installed and provides a `bar` executable.
The following hooks will be executed when set in `.asdfrc`:

```shell
 pre_asdf_install_foo = echo will install foo version ${1}
 post_asdf_install_foo = echo installed foo version ${1}

 pre_asdf_reshim_foo = echo will reshim foo version ${1}
 post_asdf_reshim_foo = echo reshimmed foo version ${1}
    
 pre_foo_bar = echo about to execute command bar from foo with args: ${@}
 post_foo_bar = echo just executed command bar from foo with args: ${@}
```

[Based on changes](https://github.com/vic/asdf/compare/new-shim-meta..command-hooks) from #432

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -310,7 +310,7 @@ get_asdf_config_value_from_file() {
   fi
 
   local result
-  result=$(grep -E "^\\s*$key\\s*=" "$config_path" | awk -F '=' '{ gsub(/ /, "", $2); print $2 }')
+  result=$(grep -E "^\\s*$key\\s*=\\s*" "$config_path" | head | awk -F '=' '{print $2}' | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
   if [ -n "$result" ]; then
     echo "$result"
   fi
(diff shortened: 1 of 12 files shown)


lib/utils.sh
```diff
@@ -406,3 +406,149 @@ resolve_symlink() {
...
+
+
+asdf_run_hook() {
+  local hook_name=$1
+  local hook_cmd
+  hook_cmd="$(get_asdf_config_value "$hook_name")"
+  if [ -n "$hook_cmd" ]; then
+    asdf_hook_fun() {
+      unset asdf_hook_fun
+      ev'al' "$hook_cmd" # ignore banned command just here
```
Stratus3D: Wouldn't just `$hook_cmd` be equivalent to this code?

colinta:
@vic I've been trying to get a post hook installed (auto-reshim after running `pip`) but either I'm misunderstanding the documentation or something is wrong... I checked that `pip` is being run via asdf:
```
which pip
/Users/colinta/.asdf/shims/pip
```
And I added this to ~/.asdfrc
```
post_python_pip = echo Just executed pip
```
but `pip install foo` doesn't run that command. 🤷‍♂

Stratus3D:
@colinta since `pip` is the executable I think you'd need something like `post_python_pip = echo Just executed pip`

vic:
@colinta Looks like a bug in documentation now, since those post-command-hooks [have already been removed](https://github.com/asdf-vm/asdf/pull/502/files)

Issue #436, opened 2019-01-22
Update command disabled


tkossak:
### Steps to reproduce
```
$ which asdf
/home/kossak/.asdf/bin/asdf
$ asdf --version
v0.6.3
$ asdf update
Update command disabled. Please use the package manager that you used to install asdf to upgrade asdf.
```

### Expected behavior

asdf was installed using `git clone` so I hoped `asdf update` would switch to proper commit. Does it mean I should checkout proper commit manually or should `asdf update` work?

### Actual behavior

Error: `Update command disabled. Please use the package manager that you used to install asdf to upgrade asdf.`
What is interesting, when I `cd` into asdf git dir it seems to be working:
```
$ cd ~/.asdf
$ asdf update
HEAD is now at 1886e00... Update version to 0.6.3
Updated asdf to release v0.6.3
$ cd /
$ asdf update
Update command disabled. Please use the package manager that you used to install asdf to upgrade asdf.
```

### Environment

**OS**: Linux Mint 18.3 Sylvia

**asdf version**: 0.6.3

Stratus3D:
I am not able to reproduce this. Is your update command only working inside the `~/.asdf` directory? Or does it work in other places too?

tkossak:
I just tried updating on my other computer and the same thing happened:

```
$ asdf --version
v0.6.0
$ asdf update
remote: Enumerating objects: 328, done.
remote: Counting objects: 100% (328/328), done.
remote: Compressing objects: 100% (80/80), done.
remote: Total 496 (delta 253), reused 310 (delta 244), pack-reused 168
Receiving objects: 100% (496/496), 133.03 KiB | 752.00 KiB/s, done.
Resolving deltas: 100% (270/270), completed with 41 local objects.
From https://github.com/asdf-vm/asdf
   f9bb44f..8ccfda2  master     -> origin/master
 * [new tag]         v0.6.1     -> v0.6.1
 * [new tag]         v0.6.2     -> v0.6.2
 * [new tag]         v0.6.3     -> v0.6.3
Previous HEAD position was aa5f7f1 Update version to 0.6.0
HEAD is now at 1886e00 Update version to 0.6.3
Updated asdf to release v0.6.3
$ asdf update
Update command disabled. Please use the package manager that you used to install asdf to upgrade asdf.
```

Stratus3D:
This is the conditional for disabled updates:

```
[...]

PR #439, opened 2019-01-24
fix "asdf update" on fish

tkossak:
Export `ASDF_DIR` variable for fish shell, so `asdf update` works properly no matter the current folder.

Fixes: https://github.com/asdf-vm/asdf/issues/436


Stratus3D:
Thanks! Now `asdf.fish` matches `asdf.sh` in this regard.

PR #440, opened 2019-01-25
Significant documentation site improvements

jthegedus:
# Summary

This contains fixes and significant improvements to the documentation site.

* fixes some routing issues due to erroneous docsify config
* fixed links and layout of 404 pg
* improved the language on the landing page as per #438 
* propose a NEW_README.md tackling concerns with #438  and #422 
* use docsify specific md to emphasise some text
* add docsify-themeable plugin to improve loading experience and future proof for customisation
* improve use of `asdf-vm` over `asdf`
* I also added myself to the maintainer list :wink: 

## Other Information

Included in this PR is a file `docs/NEW_README.md` which is my proposed README for this repo. The intention was to have all documentation in a single place, the site, and simplify the readme bringing the focus to explaining the tool rather than instructing users on how to use it.

I will create a separate PR for the README, but in the meantime the landing page composes this new readme with the ballad.

Issue #438, opened 2019-01-24
Project introduction not clear enough


coka:
Hi! :wave:

Your project was mentioned in a recent HN thread, and the [comments](https://news.ycombinator.com/item?id=18987739) included this bit:

>I have to read the README twice to see what is it about. What is extensible version manager? Managing what exactly. So I guess it's the reason asdf is not popular
>>It replaces gvm/rbenv/pyenv/nvm/etc. with a single CLI. Saves you from having to use multiple shell completion scripts, bashrc entries and CLIs. Useful if you work with multiple dev languages on your computer.
>>>And nowhere in the README tells me about that. Your first sentence immediately makes it clear

I have to admit that this person _does_ have a point. Maybe you could explicitly mention which tools this project replaces, what benefits does replacing them bring, as well as provide a comparison chart (mentioning features such as the one discussed in #367)? I understand that these things are clear to some people from the get-go, but I think that a better introduction would help onboard new asdf tinkerers.

Stratus3D:
Thank you for reporting this. I do think the readme has recently changed, and the current wording doesn't really convey what asdf's strengths are. We need to update the first few lines of the readme. Looks like the Hacker News comments have some good suggestions on how to improve it.

jthegedus:
@coka I believe this has been address appropriately in #440 and #441 

Please re-open this issue if you feel otherwise :smiley:

Please take a look at the new README and documentation site http://asdf-vm.github.io/asdf/#/ Feedback or additions are encouraged!

coka:
Yeah, much better! :smiley: Good work!

PR #441, opened 2019-01-25
De-dupe readme and docs site

jthegedus:
# Summary

This is the last major part of the work done in #422 and #440 

As discussed in #422 this change will separate out the **WHAT & WHY** from the **HOW**. The **proposed** readme is simply information for newcomers about asdf-vm and why they would use it, redirecting to the documentation site for actual instructions.

It would be good if all @asdf-vm/core were aware of this change before it merged so we are on the same page going forward about additions to the readme.

### The new readme

I've already used the proposed readme as content on the home page of the docs site. Here is what it would look like on GitHub https://github.com/asdf-vm/asdf/blob/master/docs/NEW_README.md Here it is on the site currently: https://asdf-vm.github.io/asdf/#/?id=why-use-asdf-vm

### why no quickstart in the proposed readme?
I had considered having a "quickstart" section in the new readme, but unlike a simple npm package, it's not a single line setup, and the doc site was made specifically so people could select their terminal system and get the exact information, no more, no less. Because of this, the **readme is sparse, clear and direct**, similar to the [prettier](https://github.com/prettier/prettier#opinionated-code-formatter) or [docsify](https://github.com/docsifyjs/docsify/) readme's, which are successful and popular tools with no complaints about the lack of quickstart in the readmes.

### the prev readme
As there are still some WIP changes to the main Readme, it has been "backed-up" into the `docs/` folder, and can be removed once a re-sync with the `docs/` content has occurred (I will attempt that ASAP after merging this).

### Other Information

* This should go a fair way to resolving #438 which was one of the reasons I started contributing to asdf-vm in the first place. 

### Link to the doc site from repo header

I am unable to edit the repo topics, or the repo website link in the repo header so  could you, @HashNuke add https://asdf-vm.github.io/asdf/#/ to that. It would be greatly appreciated.


jthegedus:
@Stratus3D Any thoughts on this? I would like to get the okay from most the maintainers before making such a drastic change, and I would like to make it soon before further PRs to the existing readme

README.md
```diff
@@ -1,374 +1,29 @@
...
-[plugins]: https://github.com/asdf-vm/asdf-plugins
-[plugin_api]: https://github.com/asdf-vm/asdf/blob/master/docs/plugins-create.md
 
-## INSTALLATION
+## Why use asdf-vm?
 
-If you're on macOS you can [Install using Homebrew](https://github.com/asdf-vm/asdf#homebrew-on-macos)
+- single CLI for multiple languages
+- consistent commands to manage all your languages
+- single `.tool-versions` config file per project
```
Stratus3D: Might also be nice to mention the support for other version files like `.ruby-version`, `.node-version`, etc...

README.md
```diff
@@ -1,374 +1,29 @@
-# asdf [![Build Status](https://travis-ci.org/asdf-vm/asdf.svg?branch=master)](https://travis-ci.org/asdf-vm/asdf)
+# asdf-vm [![Build Status](https://travis-ci.org/asdf-vm/asdf.svg?branch=master)](https://travis-ci.org/asdf-vm/asdf)
 
-### _extendable version manager_
```
Stratus3D: I think we could keep both these lines if we wanted to.

```
### _extendable version manager_

**Manage multiple runtimes with a single CLI tool**
```

Stratus3D:
@jthegedus and thanks for keeping the old readme checked into the repo, I'm sure it will come in handy when we are reviewing future documentation PRs and issues.

jthegedus:
@Stratus3D I'll add your suggestions, then are we happy enough for me to merge this today?

jthegedus:
I agree with the feedback is all 😄 Awesome, I'll get this in today and
then work on migrating existing readme PRs and resync any readme changes
that occurred while making this. Nearly there!

PR #445, opened 2019-01-28
update README:  asdf cannot be installed via homebrew on macOS

### LINKED ISSUE #577: Is there a way to use asdf as system global?
I wonder Is there a way to use asdf as global for every user?
to not duplicate plugins and save disks space.

unenglishable:
see [issue 577](https://github.com/thoughtbot/dotfiles/issues/577), opened by @dcalhoun

resolves #577 by changing readme to specify that asdf can't be installed via homebrew

further comments:

the documentation could also use the specific language `shouldn't` instead of `can't`
OR
the line can be removed altogether

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Supporting a new language is as simple as [this plugin API][plugin_api].
 
 ## INSTALLATION
 
-If you're on macOS you can [Install using Homebrew](https://github.com/asdf-vm/asdf#homebrew-on-macos)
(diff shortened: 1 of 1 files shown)


dcalhoun:
This likely isn't the appropriate resolution to thoughtbot/dotfiles#577. To my knowledge, installing `asdf` via Homebrew works fine. thoughtbot's dotfiles merely do not support the location at which Homebrew installs packages.

unenglishable:
@dcalhoun whoa, i didn't realize that it was an issue on thoughtbot/dotfiles!  that's a big whoops 😅 

my mistake, i'll close this!

Issue #449, opened 2019-01-30
duplicated paths for fish 3.0


tkossak:
There were some breaking changes in Fish 3.0.0, that result in asdf adding duplicated (and broken) paths for nested fish shells. This happens on both asdf 0.6.3 and latest master.

**How to reproduce:**
Start new fish >= v3.0.0 shell and write:
```fish
$ echo $fish_user_paths
/home/kossak/.asdf/bin /home/kossak/.asdf/shims
$ fish
$ echo $fish_user_paths
/home/kossak/.asdf/bin /home/kossak/.asdf/shims /home/kossak/.asdf/bin /home/kossak/.asdf/shims
```

**Result:**
In last output line variable `$fish_user_paths` (in result `$PATH` too!) contains duplicated directories (2 bins and 2 shims)

**Expected Result**
Variable `$fish_user_paths` should have only 2 different directories for asdf (like before starting second fish shell).

The asdf code adding paths is:
https://github.com/asdf-vm/asdf/blob/65aa5abe4e9858efdf798bc44f645b2b59f134ef/asdf.fish#L11-L15

The reason for duplicated paths is fish doesn't treat `$fish_user_paths` as array variable by default, so the check if directory already exists in it fails. Additionally, based on information I managed to gather from https://github.com/fish-shell/fish-shell/issues/5594#issuecomment-458889967, `$fish_user_paths` shouldn't be actually a global (nor exported) variable, but unviersal one (also that's how it is treated in [fish tutorial](http://fishshell.com/docs/current/tutorial.html#tut_path)). Is there any reason asdf makes it global? We can either make it universal one (set once, during installation, not in `config.fish`) or if for some reason we want it global - at least not export it (so the paths won't be duplicated). What do you think?

Stratus3D:
This seems related, not sure if it would fix this issue though: https://github.com/asdf-vm/asdf/pull/455

tkossak:
> This seems related, not sure if it would fix this issue though: #455

For me it doesn't fix mangled paths in fish 3.0.0. Using this new `asdf.fish` results in:

First (not nested) shell:
```fish
$ for p in $PATH; echo $p; end                                      
/home/kossak/.poetry/bin
/home/kossak/.local/bin
/home/kossak/.dotfiles.local/bin
/home/kossak/.dotfiles/bin
/home/kossak/.asdf/bin
/home/kossak/.asdf/shims
/usr/local/bin
/usr/bin
/bin
/usr/local/sbin
[...]

PR #455, opened 2019-02-03
Update fish_user_paths only if it is defined

pcr910303:
# Summary

fish_user_paths variable is prepended to $PATH automatically. Some people set fish_user_paths, while others set PATH.
It will be better to test if fish_user_paths is set, and prepend it if it is, while just prepending to PATH otherwise.

## Other Information

[setting $PATH in fish shell](http://fishshell.com/docs/current/tutorial.html#tut_path)

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -10,10 +10,6 @@ set -l asdf_bin_dirs $ASDF_DIR/bin $ASDF_DIR/shims $asdf_data_dir/shims
 
 for x in $asdf_bin_dirs
   if begin not contains $x $PATH; and test -d $x; end
-    if set -q fish_user_paths
-      set -gx fish_user_paths $fish_user_paths $x
-    else
-      set PATH $PATH $x
-    end
+    set PATH $x $PATH
   end
 end
--- a/asdf.fish
+++ b/asdf.fish
@@ -9,7 +9,11 @@ set -l asdf_data_dir (
 set -l asdf_bin_dirs $ASDF_DIR/bin $ASDF_DIR/shims $asdf_data_dir/shims
 
 for x in $asdf_bin_dirs
-  if begin not contains $x $fish_user_paths; and test -d $x; end
-    set -gx fish_user_paths $fish_user_paths $x
+  if begin not contains $x $PATH; and test -d $x; end
+    if set -q fish_user_paths
+      set -gx fish_user_paths $fish_user_paths $x
+    else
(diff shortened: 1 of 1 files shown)


tkossak:
My few cents:
Not sure why people would add it manually to `$PATH`, the [asdf docs](https://asdf-vm.com/#/core-manage-asdf-vm) state that to install asdf in fish, we have to run only:

```
echo 'source ~/.asdf/asdf.fish' >> ~/.config/fish/config.fish
mkdir -p ~/.config/fish/completions; and cp ~/.asdf/completions/asdf.fish ~/.config/fish/completions
```

In my opinion - in order to support latest fish 3.0.0 -  we should switch to using universal `$fish_user_paths` variable (as it is inteded to be used according to fish developers) and not global, nor exported one. See https://github.com/asdf-vm/asdf/issues/449 and https://github.com/fish-shell/fish-shell/issues/5594#issuecomment-458889967.

pcr910303:
Oh, I was also curious why `$fish_users_path` was a `-gx` variable instead of a universal one.
The reason why not to use that is because that variable is intended for user use, not for a config file that runs every launch.
Universal variables are saved in fish’s internal file when set, which means that setting in a config file can be error prone, does not reset in another terminal session,  and can be appendend multiple times when configured strangely.
Appending to `$PATH` is right in a config file that runs every session launch, IMHO.

pcr910303:
@tkossak 
According to the [fish-article](http://mvolkmann.github.io/fish-article/),

> The universal variable fish_user_paths is another list of paths that is automatically prepended to PATH. Universal variables are described later. Setting this modifies PATH in a way that affects all current and future sessions without having to modify config.fish and source it again. However, a benefit of changing PATH in config.fish is that the file can easily be copied to other machines to propagate the change.
fish_user_paths should only be set in a terminal, not in config.fish.

It really shouldn't modify a universal variable intended for users.

tkossak:
@pcr910303 Personally I did set universal `fish_user_paths` once from terminal and then manually commented out unnecessary lines of the `asdf.fish` file until this situation clarifies.
Although [solution with universal fish_user_paths in script](https://github.com/asdf-vm/asdf/pull/479) works for me, you're right that in scripts we shouldn't use `fish_user_paths` at all. Your PR with appending to `PATH` variable works in my fish v3.0.2 perfectly fine.

danhper:
Looks good, thanks!

PR #466, opened 2019-02-13
Fix "Edit on Github" link for changelog

Stratus3D:
Currently the URL used by the "Edit on Github" link is `https://github.com/asdf-vm/asdf/edit/master/docs/https://raw.githubusercontent.com/asdf-vm/asdf/master/CHANGELOG.md`. Not sure if this change will fix it, but it will correct the URL used when the link is clicked.

Note this is still a problem for the "Edit on Github" link on the plugins page.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -41,7 +41,7 @@
           ".*?/plugins-all":
             "https://raw.githubusercontent.com/asdf-vm/asdf-plugins/master/README.md",
           ".*?/changelog":
-            "/CHANGELOG.md",
+            "../CHANGELOG.md",
         },
         markdown: {
           smartypants: true,
--- a/docs/index.html
+++ b/docs/index.html
@@ -41,7 +41,7 @@
           ".*?/plugins-all":
             "https://raw.githubusercontent.com/asdf-vm/asdf-plugins/master/README.md",
           ".*?/changelog":
-            "https://raw.githubusercontent.com/asdf-vm/asdf/master/CHANGELOG.md",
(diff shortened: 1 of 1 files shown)


jthegedus:
This change will require us to either move or duplicate the `CHANGELOG.md` file to `docs/` as the edit URL after this change would be `https://github.com/asdf-vm/asdf/edit/master/docs/CHANGELOG.md`.

My solution to this issue was to re-write the edit-on`github` plugin as it doesn't account for route `alias`es. The Docsify site has it's own custom plugin with support for these issues, seen here https://github.com/docsifyjs/docsify/blob/cb53106fa7f8a6fa904c1b05db1923befc26518e/docs/index.html#L63

If the Changelog and All-Plugins page edit buttons are needed to be fixed immediately **I would opt for a custom plugin at this time**.

When I get around to writing a more flexible `edit-on-github` plugin we can migrate

Stratus3D:
Oops... I meant for that URL to be `../CHANGELOG.md`. Which should become `https://github.com/asdf-vm/asdf/edit/master/docs/../CHANGELOG.md` which will redirect to `https://github.com/asdf-vm/asdf/edit/master/CHANGELOG.md`. Do you think that would work @jthegedus ?

Either way there is no hurry on this. Just something I noticed today.

jthegedus:
As a link to the correct `md` file it will work, I am not sure that Docsify will register that correctly and render the file as expected.

Stratus3D:
Closing this for now then. I'll create a new issue to track this.

PR #471, opened 2019-02-19
Add a note about Spack installation

whitfin:
This PR adds documentation to cover that components can be installed via [Spack](https://spack.io/) on macOS. There's really no difference in the package names (and so on), but this serves to show that there are more options than just Homebrew! 

As a side note, I do not know if the website is automatically generated from GitHub. If so, it might be good to wait on this PR for a little while. I recently had the final specification for `libyaml` merged into Spack - I'm not entirely sure when that becomes more broadly available.

### DIFF
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -75,10 +75,18 @@ tab will usually do it.)
 
 #### ** macOS **
 
-Install these via homebrew:
+Installation via Homebrew:
 
 ```shell
 brew install \
+  coreutils automake autoconf openssl \
+  libyaml readline libxslt libtool unixodbc
+```
+
+Installation via Spack:
+
+```shell
+spack install \
   coreutils automake autoconf openssl \
   libyaml readline libxslt libtool unixodbc
 ```


JuPlutonic:
[Spack](https://spack.io/) and [Homebrew](https://brew.sh/) are now  can be put to `#### ** macOS and Linux**` section.
We may show here what in common sense these packagers are the same.

jthegedus:
@whitfin Thanks for contributing!

This looks good to me. Yes, the site is auto-generated from the `docs/` folder so once this is merged it will be live. When you know `libyaml` is good to go, ping me and I will merge.

@JuPlutonic I'm not a fan of combining instructions for platforms in the docs if we can avoid it. The nature of [`docsify-tabs`](https://jhildenbiddle.github.io/docsify-tabs/#/?id=sync) means that if a user selects a tab, and that tab name is used elsewhere, then all tabs are changed across the site, and combining platforms breaks this useful feature.

whitfin:
@jthegedus ok, it looks like it's good now! The `libyaml` spec was merged, and I can see it when I follow the fresh setup guide of Spack, so I think that's sufficient.

jthegedus:
Thanks for the contribution @whitfin :tada:

PR #472, opened 2019-02-21
asdf.sh: support mksh

fallwith:
The mksh shell (and perhaps others too) will not be able to determine a
sourced script's path by looking at either `${BASH_SOURCE[0]}` or `$0`.

By capturing the value of `$_` as the very first thing the script does,
`asdf.sh` can use that as the input for `dirname` in order to determine
a value to use for `$ASDF_DIR`.

I have tested this change with mksh to confirm that it works, and have
also tested it with both bash and zsh to verify that neither of those
stop working.

This resolves the only compatibility issue I've ran into with asdf,
caused by mksh sourcing a script. All subsequent asdf usage involves an
asdf executable doing the sourcing, so it's bash to bash at that point.

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+# For Korn shells (ksh, mksh, etc.), capture $_ (the final parameter passed to
+# the last command) straightaway, as it will contain the path to this script.
+# For Bash, ${BASH_SOURCE[0]} will be used to obtain this script's path.
+# For Zsh and others, $0 (the path to the shell or script) will be used.
 _under="$_"
 if [[ "$_under" == *".sh" ]]; then
   current_script_path="$_under"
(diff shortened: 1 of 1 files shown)


Stratus3D:
Changes look good to me. What do you think @danhper ?

Stratus3D:
Actually, it might be good to add a comment to the code explaining that the conditional is need for mksh.

danhper:
@Stratus3D I agree that a comment would help.
Maybe adding what `$_` is supposed to return in the comment would be nice.
For what I have tried (when using either `source test.sh` or `. ./test.sh`)

* bash: `_filedir`
* sh: `/bin/sh`
* ksh: `test.sh`
* mksh : `test.sh`

Once we have the comment I think we can merge!

fallwith:
Thank you, @Stratus3D and @danhper. I added a comment with 2743bcb for your consideration.

danhper:
Great, thanks!

PR #479, opened 2019-03-08
Set fish_user_paths as universal variable

dtcristo:
# Summary

For fish shell users, `$fish_user_paths` is incorrectly being set as a global variable and will clobber anyone setting `$fish_user_paths` the way recommended by the [fish documentation](https://fishshell.com/docs/current/tutorial.html#tut_path) (as a universal variable).

This change correctly sets `$fish_user_paths` as a universal variable. See more details [here](https://github.com/asdf-vm/asdf/issues/449#issuecomment-465550336).

Fixes: #449


danhper:
Thanks for the PR. As discussed in #455, we went with `PATH` rather than `fish_user_paths`.

Issue #378, opened 2018-10-19
enhancement: Add "shell" command


sr105:
`pyenv` has "shell", "local", and "global" commands. The "shell" command sets or reports the currently set tool version via environment variable. It would be nice if `asdf` had the "shell" command to to complement Enhancement #49. 

```
Usage: pyenv shell <version>...
       pyenv shell -
       pyenv shell --unset

Sets a shell-specific Python version by setting the `PYENV_VERSION'
environment variable in your shell. This version overrides local
application-specific versions and the global version.

<version> should be a string matching a Python version known to pyenv.
The special version string `system' will use your default system Python.
Run `pyenv versions' for a list of available Python versions.

When `-` is passed instead of the version string, the previously set
version will be restored. With `--unset`, the `PYENV_VERSION`
environment variable gets unset, restoring the environment to the
state before the first `pyenv shell` call.
```

PR #480, opened 2019-03-09
Add "shell" subcommand for setting versions in shell session

jonmast:
# Summary
Add a "shell" command similar to the existing "global" and "local" commands, which sets the version in an environment variable instead of writing it to a file. This was inspired by the similar functionality in rbenv.

It works by adding a wrapper function for the asdf command. This forwards to a "sh-shell" command that returns the exports as shell code which is then evaled by the wrapper. This is a little gross, but we need to run the code in the shell context in order to set variables, and it matches rbenv's approach.

Fixes: #378

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -13,3 +13,18 @@ for x in $asdf_bin_dirs
     set -gx fish_user_paths $fish_user_paths $x
   end
 end
+
+# Add function wrapper so we can export variables
+function asdf
+  set command $argv[1]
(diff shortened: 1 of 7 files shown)


danhper:
Thanks for helping out!
Both the functionality and the implementation look good to me.
@asdf-vm/core What do you think?

danhper:
I just saw that @Stratus3D said he liked the idea in #378 so let's merge.
Thanks again for contributing!

Issue #482, opened 2019-03-14
Multiple versions broken


danhper:
Hi guys,

Sorry I did not have much time to help recently.
It seems that with the latest version we lost the ability to set multiple versions for the same plugin.
I've been relying on this behavior a lot and I think it's a must have for Python.
I'll try to look into the changes to see how this can be restored but I'll start by opening an issue.    

### Steps to reproduce

```
asdf global python 3.7.2 2.7.15
python2 --version
```

### Expected behavior

Print `python 2.7.15`

### Actual behavior

Tells that no version is set for Python2

### Environment

**OS**: Not relevant for this issue

**asdf version**: Anything after 9b27848d07b86a7fe5c5294e89bf7f3674ea50f2

Issue #470, opened 2019-02-18
asdf is not using version from local legacy file


pmk1c:
### Steps to reproduce

Install asdf 0.6.3
Run `asdf plugin-add ruby`
Create a directory with a `.ruby-version` file and a ruby version VERSION specified
Run `asdf install ruby VERSION`

### Expected behavior

`ruby -v` should give VERSION

### Actual behavior

`ruby -v` gives version from default `~/.tool-versions`

### Environment

**OS**: Fedora

**asdf version**: 0.6.3

The problem disappears when I downgrade to asdf 0.6.2 and reshim ruby.
Since I can fix the problem by downgrading, I thought this might be a asdf problem, but maybe it's related to `asdf-ruby`.

rickclare:
This is also occurring for me, after upgrading to **asdf v0.7.0** (via homebrew).  It doesn't look like it's just applicable to `asdf-ruby`, as **asdf** also appears to be ignoring my local `.nvmrc` file 

```sh
$ echo $SHELL
/usr/local/bin/fish

$ fish --version
fish, version 3.0.2

$ asdf --version
v0.7.0

$ cat ~/.tool-versions
elixir 1.8.1-otp-21
nodejs 11.10.0
postgres 11.2
redis 3.2.12
ruby 2.6.1
erlang 21.2.6

$ cd ~/Dev/my_app

$ cat ~/.asdfrc
legacy_version_file = yes

$ cat ./.ruby-version
2.3.7

$ cat ./.nvmrc
6.11.5

$ asdf current 
elixir         1.8.1-otp-21 (set by /Users/rick/.tool-versions)
erlang         21.2.6   (set by /Users/rick/.tool-versions)
nodejs         6.11.5   (set by /Users/rick/Dev/my_app/.nvmrc)
postgres       11.2     (set by /Users/rick/.tool-versions)
redis          3.2.12   (set by /Users/rick/.tool-versions)
ruby           2.3.7    (set by /Users/rick/Dev/my_app/.ruby-version)

$ /usr/bin/which ruby
/Users/rick/.asdf/shims/ruby

$ /usr/bin/which node
/Users/rick/.asdf/shims/node

$ ruby -v
ruby 2.6.1p33 (2019-01-30 revision 66950) [x86_64-darwin17]

$ node --version
v11.10.0
```

bowmanmike:
[...]

PR #484, opened 2019-03-14
Fix version handling

danhper:
# Summary

Some regressions had been introduced in version handling, the most important one being the support for legacy files.

Support for multiple versions of the same plugin had also changed, making it incompatible with `asdf current`, `asdf local` and `asdf global` commands.
Up to 9b27848d07b86a7fe5c5294e89bf7f3674ea50f2, multiple versions could be set in `.tool-versions` by separating versions with a space. After 9b27848d07b86a7fe5c5294e89bf7f3674ea50f2, the same could be achieved by adding multiple lines with the same plugin name.
However, this plays badly with `asdf local` and `asdf global` as there is no easy way to set multiple versions, while it is trivial with the single line approach.
This PR reverts the behavior to use space separated versions.

I also added a regression test for legacy files and some documentation for using multiple versions.

Please take a look when you have a moment @vic @Stratus3D 

Fixes: #482, #470

### DIFF
--- a/test/shim_exec.bats
+++ b/test/shim_exec.bats
@@ -116,9 +116,19 @@ teardown() {
 @test "shim exec should execute first plugin that is installed and set" {
   run asdf install dummy 3.0
 
-  echo "dummy 1.0" > $PROJECT_DIR/.tool-versions
-  echo "dummy 3.0" >> $PROJECT_DIR/.tool-versions
-  echo "dummy 2.0" >> $PROJECT_DIR/.tool-versions
+  echo "dummy 1.0 3.0 2.0" > $PROJECT_DIR/.tool-versions
+
+  run $ASDF_DIR/shims/dummy world hello
+  [ "$status" -eq 0 ]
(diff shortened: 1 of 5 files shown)


test/shim_exec.bats
```diff
@@ -116,9 +116,7 @@ teardown() {
 @test "shim exec should execute first plugin that is installed and set" {
   run asdf install dummy 3.0
 
-  echo "dummy 1.0" > $PROJECT_DIR/.tool-versions
-  echo "dummy 3.0" >> $PROJECT_DIR/.tool-versions
-  echo "dummy 2.0" >> $PROJECT_DIR/.tool-versions
```
Stratus3D: What is the expected behavior for scenarios like this? Even though this isn't supported it might be good to have a test that shows only the first version is used and the others are ignored.

danhper: Sounds good, I'll add a test for this and merge.

PR #487, opened 2019-03-19
Fix select version when multiple versions available

danhper:
With the previous version, the following case would fail.
It would use python 2.7.15 when running pip
instead of version 3.7.2.

Shim for `pip`

```bash
exec /home/daniel/.asdf/bin/asdf exec "pip" "$@"
```

`.tool-versions`:

```
python 3.7.2 2.7.15 system
```

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -603,24 +603,30 @@ with_shim_executable() {
   }
 
   select_version() {
-    # First, we get the all the plugin/version pairs where the
+    # First, we get the all the plugins where the
     # current shim is available.
-    # Then, we iterate on these and check if the version of the current
-    # plugin is the one currently set.
+    # Then, we iterate on all versions set for each plugin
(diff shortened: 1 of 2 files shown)


lib/utils.sh
```diff
@@ -603,24 +603,30 @@ with_shim_executable() {
...
     IFS=$'\n' read -rd '' -a shim_versions <<< "$(get_shim_versions)"
 
+    local plugins
+    plugins=()
+
     for plugin_and_version in "${shim_versions[@]}"; do
       local plugin_name
       local plugin_shim_version
       IFS=' ' read -r plugin_name plugin_shim_version <<< "$plugin_and_version"
+      if ! [[ " ${plugins[*]} " == *" $plugin_name "* ]]; then
```
Stratus3D: Maybe prefix `plugin_shim_version` with an underscore since it's not used?

Issue #477, opened 2019-03-07
Where command does not print a helpful message when system version is set


Stratus3D:
Discovered `~/.asdf` was a git repo, checked out tag `v0.7.0`, ran `asdf reshim`.

```
~$ asdf which node
/usr/bin/node
~$ asdf where nodejs
Version not installed
```

The `where` could still be improved, but the `which` command appears to be fixed.

_Originally posted by @mclark-newvistas in https://github.com/asdf-vm/asdf-nodejs/issues/122#issuecomment-469817417_

PR #493, opened 2019-03-23
Improve output of where command when system version is set

Stratus3D:
# Summary

Update `where` command so `System version is selected` is printed instead of `Version not installed` when the system version is selected for the plugin.

Fixes: #477


danhper:
Looks good, thanks!

PR #496, opened 2019-03-24
Add Homebrew instructions

dboune:
# Summary

Added installation, shell setup, and upgrade instructions for use with Homebrew.
Added a brief item for uninstall related to package managers.

Fixes: #425

## Other Information

I decided to not make the `Add to your Shell` portion a tab item of its own, instead keeping it under `Bash on macOS`. I followed nomenclature found elsewhere in the document to clarify variant install situations.

It's worth noting that Homebrew can be used with Windows 10 with Windows Subsystem for Linux, though I'm not sure if that is applicable to `asdf` at present.

### DIFF
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -1,5 +1,9 @@
 ## Install asdf-vm
 
+<!-- tabs:start -->
+
+#### ** Git **
+
 Clone only the latest branch:
 
 ```shell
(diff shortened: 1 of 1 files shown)


jthegedus:
Hmm, seems this will get out of hand once we add windows with wsl instructions. I'll explore select lists over tabs this coming weekend.

dboune:
Sounds good. There really are so many ways and different systems, and if ‘asdf’ isn’t strongly opinionated in how it is done, then that will be for the best.

I’ll leave this here for now and you can scrape it and/or scrap it when ready.

jthegedus:
A lot of people have been asking for this so getting it in now is more important than having a perfect solution now imo

Issue #489, opened 2019-03-22
Comments in .tool-versions file


axelson:
### Steps to reproduce

With a `.tool-versions` like this:
```
# Note: Update Dockerfile and elixir_buildpack.config as well
elixir 1.8.1-otp-21
erlang 21.2.4
```

### Expected behavior

The specified elixir and erlang versions should be installed and lines beginning with `#` should be ignored.

### Actual behavior

Running `asdf install` fails
```
$ asdf install
No such plugin: #
```

But `asdf current` works:
```
jason@Jasons-MacBook-Pro-2 ~/n/messages> asdf current
elixir         1.8.1-otp-21(set by /Users/jason/my_app/.tool-versions)
erlang         21.2.4  (set by /Users/jason/my_app/.tool-versions)
```

### Environment

**OS**: 10.13.6

**asdf version**: `v0.7.0`

danhper:
I think this is a good idea.
If someone would like to tackle this one, it would be great.

PR #498, opened 2019-03-25
Make install command ignore comments

Stratus3D:
# Summary

Make install command ignore comments. Skip lines matching the pattern `^#.*`, e.g. any line starting with `#`.

Fixes: #489

### DIFF
--- a/lib/commands/install.sh
+++ b/lib/commands/install.sh
@@ -40,19 +40,17 @@ install_local_tool_versions() {
   local asdf_versions_path
   asdf_versions_path=$(find_tool_versions)
   if [ -f "${asdf_versions_path}" ]; then
-    while IFS= read -r tool_line || [ -n "$tool_line" ]; do
-      if ! (echo "$tool_line" | grep "^#.*" > /dev/null); then
-        IFS=' ' read -r -a tool_info <<< "$tool_line"
-        local tool_name
-        tool_name=$(echo "${tool_info[0]}" | xargs)
-        local tool_version
-        tool_version=$(echo "${tool_info[1]}" | xargs)
+    while IFS= read -r tool_line; do
+      IFS=' ' read -r -a tool_info <<< "$tool_line"
+      local tool_name
+      tool_name=$(echo "${tool_info[0]}" | xargs)
+      local tool_version
+      tool_version=$(echo "${tool_info[1]}" | xargs)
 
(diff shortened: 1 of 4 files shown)


lib/commands/install.sh
```diff
@@ -41,14 +41,16 @@ install_local_tool_versions() {
   asdf_versions_path=$(find_tool_versions)
   if [ -f "${asdf_versions_path}" ]; then
     while IFS= read -r tool_line || [ -n "$tool_line" ]; do
-      IFS=' ' read -r -a tool_info <<< "$tool_line"
-      local tool_name
-      tool_name=$(echo "${tool_info[0]}" | xargs)
-      local tool_version
-      tool_version=$(echo "${tool_info[1]}" | xargs)
+      if ! (echo "$tool_line" | grep "^#.*" > /dev/null); then
```
danhper: I think I would probably rather go with something like

```shell
tool_line="$(echo "$tool_line" | sed -r -e 's/(\#.*)//')"
```

and leave the decision of installing or not to the condition we already have: 

```
if ! [[ -z "$tool_name" || -z "$tool_version" ]];
```

This would allow inline comments:

```
python 3.5.2 2.7.15  # fallback to py2 if command not found in py3
nodejs 8.12.0  # anything above 8 should work
``` 

What do you think?

Stratus3D: Good idea. I've updated the code (and added unit tests) so comments are where ever they appear - on their own line, or at the end of a version line.

danhper:
Thanks! Looks good but it seems to be failing on mac, I'm not sure why.

lib/utils.sh
```diff
@@ -574,6 +574,19 @@ shim_plugin_versions() {
   fi
 }
 
+strip_tool_version_comments() {
+  local tool_version_path="$1"
+
+  while IFS= read -r tool_line || [ -n "$tool_line" ]; do
+    # Remove whitespace before pound sign, the pound sign, and everything after it
+    new_line="$(echo "$tool_line" | cut -f1 -d"#" | xargs)"
```
Stratus3D: Today I learned `xargs` can be used to remove leading and trailing whitespace`.

Stratus3D:
@danhper just fixed the build on OSX. At first I was using sed with extended regex (`-r`) and then switched to `xargs` for removing whitespace. I settled on using `cut` to remove the comments, and some simple sed regex to remove trailing whitespace. The simpler regex works on Linux and OSX.

PR #504, opened 2019-03-31
Fix current when .tool-versions contains comments

danhper:
#498 did not support comments when using `asdf current`.
This should fix the issue.


Stratus3D:
Thanks @danhper ! I knew there was another place I needed to call that function!

PR #506, opened 2019-04-05
Always prepend asdf_bin_dirs to PATH

jechol:
# Summary

Always prepend asdf_bin_dirs to PATH.

Fixes: 

When launch new fish shell in tmux, /etc/paths are prepended so that asdf_bin_dirs are after system dirs.
This commit brings asdf_bin_dirs before system dirs.


danhper:
Hi, thanks for the PR.
I am fine with prepending directories to the path but this PR breaks support for Fish < 2.3.
Fish 2.2 being the default on Ubuntu Xenial, I want to keep supporting it for now.
Could you please implement this without using `string`, which only appeared in 2.3, and keep the `test -d` too, to avoid issues with older versions of fish?
Thank you.

jechol:
@danhper Replaced `string` with `printf` and `grep`.

Some users alias `xargs` with `xargs -0`, so I used `command` to reset it.

Stratus3D:
@jechol we've now got some unit tests for this code merged into master. Can you rebase your branch onto master to make sure the tests still pass?

danhper:
Looks good, thanks a lot!

Issue #286, opened 2018-02-13
Add tests for code in asdf.sh


Stratus3D:
Our builds don't catch breaking changes in asdf.sh. We should be evalating this code during the build so we can have some assurance it's working properly.

#256

Stratus3D:
Also need to add tests for behavior of shims so we can assert their exit codes and output.

danhper:
I think we are now mostly testing for this. Let's open a new issue if at some point we think it is not enough.

Stratus3D:
I do not believe the code in https://github.com/asdf-vm/asdf/blob/master/asdf.sh is tested in any of our unit tests. We need to write unit tests for the code in `asdf.sh` and `asdf.fish`. Specifically, we need to test:

* `PATH` is set properly
* Any other environment variables are set properly
* Functions like `asdf` are working as expected (e.g. providing the commands we've defined)

I think this is worth doing as this code for this has been broken on several occasions. I will create a PR for this as soon as I can.

Stratus3D:
I still need to write tests for fish.

PR #508, opened 2019-04-06
Add tests for code in asdf.fish

Stratus3D:
# Summary

Write unit tests for code in asdf.fish.

Fixes: #286

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,7 @@ script:
   - ./lint.sh
 
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install shellcheck; fi
+  - ./prepare-travis.sh
 
 before_script:
   - git clone https://github.com/sstephenson/bats.git /tmp/bats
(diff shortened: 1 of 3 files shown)


Stratus3D:
Looks like Travis CI doesn't have fish shell installed.

danhper:
@Stratus3D  Here is the script I use to install fish on Travis:
https://github.com/danhper/fundle/blob/master/prepare-ci.sh
Let me know if you would like me to help.

Stratus3D:
Thanks @danhper , I created a script for setting up Travis and got all but one of the asdf.fish tests passing on Travis. They all passed for me locally, and I tried several things and eventually gave up on getting it to pass on Travis, so I commented it out. The other three tests passed on Travis.

danhper:
Thanks a lot for working on this. Looks good to me!

I found the issue. It seems that `status -f`, which is used to set `ASDF_DIR`, is very naive and returns the current filename as given to fish. This means that if we call `source asdf.fish`, `status -f` will return `./asdf.fish`. Therefore, `ASDF_DIR` will become `.`. It then appends `./bin` to the path and the `grep` fails.
I believe it works in your environment because `$HOME/.asdf/shims` exist. `ASDF_DATA_DIR` will default to `$HOME/.asdf` and `$HOME/.asdf/shims` will therefore be added to the path, making `grep` succeed.

I just pushed the fix to this branch. Please merge if it looks good.

Stratus3D:
Awesome! Thanks for fixing that test.

PR #510, opened 2019-04-09
Add more plugin dependencies 

stefansedich:
# Summary

I was doing a test of getting up and running using asdf on a fresh Ubuntu install it became aparant that curl and unzip were required for some plugins to install, it makes sense to me to add these to the base dependencies list to give a better getting started experience.

### DIFF
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -108,7 +108,8 @@ Installation via Homebrew:
 ```shell
 brew install \
   coreutils automake autoconf openssl \
-  libyaml readline libxslt libtool unixodbc unzip curl
+  libyaml readline libxslt libtool unixodbc \
+  unzip curl
 ```
 
 Installation via Spack:
@@ -116,7 +117,8 @@ Installation via Spack:
 ```shell
 spack install \
   coreutils automake autoconf openssl \
-  libyaml readline libxslt libtool unixodbc unzip curl
+  libyaml readline libxslt libtool unixodbc \
+  unzip curl
 ```
 
 #### ** Ubuntu **
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -108,7 +108,7 @@ Installation via Homebrew:
 ```shell
 brew install \
   coreutils automake autoconf openssl \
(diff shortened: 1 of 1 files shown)


jthegedus:
Thanks for the addition. While I will add these as they are common to many tools, I feel this isn't the exact place plugin deps should be.

I've some ideas for improving the experience with plugin documentation particularly around dependencies, but they're still ideas.

Do you have any thoughts for plugin authors to improve the experience?

For instance, do you think we should have all plugin authors test their install instructions against systems that:
* are fresh OS installs
* asdf-vm installed (with it's deps)
* no other plugins

stefansedich:
Thanks @jthegedus I see your point and agree this is not sustainable to keep adding random tools plugin X needs to work.

I will have to have a think about it a little bit and get back to you :).

jthegedus:
Thanks @stefansedich I look forward to it 😄

Stratus3D:
I think we should probably be going the other direction here. We should not be listing packages as dependencies of asdf if they are not truly packages it depends on. `curl` and `unzip` are currently not required. But neither is `unixodbc` I don't think. I think the asdf docs should only list actual dependencies for asdf core. For many of us it doesn't matter, but if someone is using asdf on a low power machine, or in a build, we don't want them installing unnecessary packages. For example, my Lua plugin could be used on very low power machine, and all it requires is a C compiler. Ideally all the user should be instructed to do is install asdf, asdf-lua, and the C compiler.

Plugin READMEs should already be specifying all the dependencies they require, and from what I've seen most of them are. (Please create issues on the plugin repos if you see otherwise).

stefansedich:
@Stratus3D none of the ones in question had any indication of dependencies, but in this case it seems to make sense to me at least that unzip and curl would be pretty widely used and should be part of the common base dependencies to give most people a good initial experience.

@jthegedus what I am trying to achieve is a simple "just install asdf and run make-magic-happen.sh" with the least amount of pain. At the moment while you can play the install dependencies until it works game I feel it could be better.

I wonder if plugins can be smarter about checking that the tools they need are actually installed or not and provide better feedback to the user that they need to install X, Y and Z, would be nice to do it for them or provide the commands but obviously this is not feasible with all the instalation options out there.

PR #514, opened 2019-04-16
Run source after .bashrc update in Linux bash

marisradu:
After you update .bashrc in Linux, you either need to run source or re-open the terminal. Added the source command in this install recommendation doc.


danhper:
We already mention

> Restart your shell so that PATH changes take effect

sourcing `.bashrc` could have unwanted side effects so I would rather stick with the current version.

Stratus3D:
Closing this because I don't think this change is needed.

PR #518, opened 2019-05-01
update setup documentation for zsh and homebrew

FlorinSenoner:
Minor update the setup documentation for zsh and homebrew users. Hope this makes it easier for some people in the future to add the correct lines to their `~/.zshrc` file

### DIFF
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -66,9 +66,23 @@ If you are using a framework, such as oh-my-zsh, use these lines. (Be sure
 that if you make future changes to .zshrc these lines remain _below_ the line
 where you source your framework.)
 
+Installation via **Git**:
+
 ```shell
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR! My only concern here is that we've got quite a bit of duplication. I wonder if we could reorganize things here so we don't have to repeat the paragraph on Homebrew completions.

olets:
Fwiw I like the duplication. If it was code it'd be nice to condense it more, but a given reader is going to be in one tab or the other

jthegedus:
It might be time for us to split out the shared contents of the tabs into their own .md files and use the [Docsify include feature](https://docsify.js.org/#/embed-files?id=embed-files). Just a suggestion. It can get unwieldy as well, but probably less so than the duplication

danhper:
Sorry, I just realized this PR was opened. I merged #537, which was doing mostly the same thing, so I will close this one.
Thanks again!

Issue #160, opened 2017-02-13
Use `set -o nounset -o pipefail -o errexit` on top/in all scripts


ypid:
Ref: https://news.ycombinator.com/item?id=10736584

Stratus3D:
I've been using the "Unofficial Bash Strict Mode" (http://redsymbol.net/articles/unofficial-bash-strict-mode/) in my personal Bash scripts, which is very similar to this. For an example see this: https://github.com/Stratus3D/dotfiles/blob/master/scripts/tools/epmd_port_forwarder

@HashNuke @tuvistavie @vic what do you guys think? This can have a pretty significant effect on the way we write Bash script so it's important everyone is on board with this. I'm fine either way. We've got a good test suite so this isn't critical.

tuvistavie:
I do not have any strong opinion about this, so if you want to use these options it I am fine with it too.

vic:
Now that we have #223 merged, I'm +1 on also adding the "bash script mode"  to all our bash files.

```
set -euo pipefail
IFS=$'\t\n' # Stricter IFS settings
```

I'll try adding this and test everything goes as expected.

ypid:
```Shell
set -euo pipefail
```

> I used to do that but the long versions are more understandable by other people who will look at the script.

Ref: https://news.ycombinator.com/item?id=10738276

I would also prefer the long version. Why go with the short one?

Stratus3D:
@ypid that's a good question. I typically use the long version in my personal scripts, but I'm not sure there was a reason behind using the short version in asdf. We could add the details on strict mode to the document contributing rather than switch to the long version of strict mode. One nice this with the short version is that it's only two lines, so it hardly takes up any space at the top of the file.

ypid:
> One nice this with the short version is that it's only two lines, so it hardly takes up any space at the top of the file.

My previous quote was incomplete. Both versions fit into two lines:

```Shell
set -o nounset -o pipefail -o errexit
IFS=$'\t\n' # Stricter IFS settings
```

vs.

```Shell
set -euo pipefail
IFS=$'\t\n' # Stricter IFS settings
```

Note that I have not yet checked out the `IFS` setting in depth nor used it, but I have used `set -o nounset -o pipefail -o errexit` since about a year in Bash scripts and find it useful.

vic:
Yes, using the long version is better, thanks for pointing that.

peterwwillis:
[...]

PR #520, opened 2019-05-11
WIP: Bash Strict Mode

Stratus3D:
# Summary

Add Bash strict mode to all Bash script that are not sourced directly by the user.

We can't add Bash script mode to any files the user may source, or may get sourced by any other scripts, as the strict mode settings will leak into other contexts. I've had to make some changes to a few of the tests that emulate a user shell to get around this in the tests.

Fixes: #160, #247

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+set -o nounset -o pipefail -o errexit
+IFS=$'\t\n' # Stricter IFS settings
 
 # shellcheck source=lib/utils.sh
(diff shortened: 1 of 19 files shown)


Stratus3D:
I've got one failing test locally, but nearly all the tests fail on Travis.CI and I'm not sure why.

lastthyme (approved):

Stratus3D:
I still want to add Bash Strict Mode to asdf, but this PR is way too old to use, and https://github.com/asdf-vm/asdf/pull/817 takes a more incremental approach. All further discussion should be done on https://github.com/asdf-vm/asdf/pull/817.

PR #521, opened 2019-05-12
Switch to latest version of BATS

Stratus3D:
We need to change the BATS we use to get around this issue:
https://github.com/sstephenson/bats/issues/171


Stratus3D:
This new fork is more active, and I haven't heard back from https://github.com/sstephenson.

danhper:
Sounds good, thanks!

PR #524, opened 2019-05-17
fix: check shims for the full version

### LINKED ISSUE #517: No version set for command ... - still happening in v0.7.1
Still having the "no version set" issue (originally reported in https://github.com/asdf-vm/asdf/issues/488) with v0.7.1 even after run `asdf reshim`

### Steps to reproduce

my .tool-versions is

```
erlang 21.0.6
elixir path:/home/eksperimental/elixir/
```

```
$ cd /home/eksperimental/elixir/
$ iex
asdf: No version set for command iex
you might want to add one of the following in your .tool-versions file:

elixir 1.4.4
elixir 1.7.4
elixir 1.8.0
elixir 1.8.1
```

### Expected behavior

to run `/home/eksperimental/elixir/bin/iex`

### Actual behavior

```
asdf: No version set for command iex
```

### Environment

**OS**: Ubuntu GNU/Linux 16.04 32 bits

**asdf version**: v0.7.1

paulswartz:
The current behavior in shims is to check if there's an existing shim for
which the shim we're currently checking is a prefix. For example, if the shim
has

    # asdf-plugin: erlang 21.2.6

Then adding a shim for `erlang 21.2` will fail.

This updates the `grep` check to match the end of the line as well so we
always get a full version check.

Fixes #517

### DIFF
--- a/lib/commands/reshim.sh
+++ b/lib/commands/reshim.sh
@@ -65,7 +65,7 @@ write_shim_script() {
   shim_path="$(asdf_data_dir)/shims/$executable_name"
 
   if [ -f "$shim_path" ]; then
-    if ! grep "# asdf-plugin: ${plugin_name} ${version}" "$shim_path" >/dev/null; then
+    if ! grep -x "# asdf-plugin: ${plugin_name} ${version}" "$shim_path" >/dev/null; then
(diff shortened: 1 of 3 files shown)


danhper:
Good catch! Thanks for the PR.

Issue #269, opened 2018-01-05
`asdf install` should read versions from legacy version files


Stratus3D:
When determining what versions need to be installed `asdf install` should pay attention to legacy version files on the path when it is configured to do so. Right now the command disregards any legacy files that maybe in the search path.

We could optionally add a flag here for reading legacy version files, but I don't think it's necessary since the user has to manually configure asdf to read legacy version files anyway. And all the other commands do read legacy version files by default when configured to do so.

samneely:
I'm new to `asdf` and noticed this as well. I'm really hoping that this becomes a feature with `asdf install`. Is this something that could be added in the near-ish future?

PR #533, opened 2019-05-31
Change `asdf install` to loop through available plugins and check legacy version files

leonid-shevtsov:
I set out with the goal to make `asdf install` install plugin versions from legacy version files (like `.nvmrc`).

However, since `asdf install` normally loops through the `.tool-versions` file, and wouldn't do anything at all if the `.tool-versions` file doesn't exist, I had to change the general approach to determining what to install.

So now, instead of looping through the `.tool-versions` file, **the command will loop through all available plugins, and ensure that whatever version is selected for the plugin and in whichever way, this version is installed.**

(One downside is it might try to install tools irrelevant to my current project - for example: if I have the elixir plugin, and the nodejs plugin, and I'm only using the latter in my current project, but I don't have the globally selected elixir version installed yet - it will go and install elixir, too. But that would only happen if I have set a global elixir version and haven't installed it yet.)

Fixes: #269

### DIFF
--- a/lib/commands/install.sh
+++ b/lib/commands/install.sh
@@ -37,21 +37,35 @@ get_concurrency() {
 }
 
 install_local_tool_versions() {
-  local asdf_versions_path
-  asdf_versions_path=$(find_tool_versions)
-  if [ -f "${asdf_versions_path}" ]; then
-    while IFS= read -r tool_line; do
-      IFS=' ' read -r -a tool_info <<< "$tool_line"
-      local tool_name
-      tool_name=$(echo "${tool_info[0]}" | xargs)
-      local tool_version
-      tool_version=$(echo "${tool_info[1]}" | xargs)
+  local plugins_path
+  plugins_path=$(get_plugin_path)
 
-      if ! [[ -z "$tool_name" || -z "$tool_version" ]]; then
-        install_tool_version "$tool_name" "$tool_version"
+  local search_path
+  search_path=$(pwd)
+
+  local some_tools_installed
+
+  if ls "$plugins_path" &> /dev/null; then
+    for plugin_path in "$plugins_path"/* ; do
+      local plugin_name
+      plugin_name=$(basename "$plugin_path")
+
+      local plugin_version_and_path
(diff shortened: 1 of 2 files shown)


lib/commands/install.sh
```diff
@@ -37,21 +37,35 @@ get_concurrency() {
...
-        install_tool_version "$tool_name" "$tool_version"
+  local plugins_path
+  plugins_path=$(get_plugin_path)
+
+  local search_path
+  search_path=$(pwd)
+
+  local some_tools_installed
+
+  if ls "$plugins_path" &> /dev/null; then
```
Stratus3D: If you are just wanting to check if this directory exists, `[ -d "$plugins_path" ]` is going to work on more systems. If you are wanting to check for subdirectories I think you will need `ls "$plugins_path"/*/`.

Stratus3D:
Thanks for the PR! This is a nice improvement.

I noticed the build is failing on Linux and OSX due to a failing test.

```
not ok 37 install_command skips comments in .tool-versions file
# (in test file test/install_command.bats, line 161)
#   `[ "$status" -eq 0 ]' failed
```

This might be caused by the `exit 1` you added on line 65. It would also be nice to have a new test to test that the `install` command installs versions from legacy files.

lib/commands/install.sh
leonid-shevtsov: I've taken this code as-is from 
https://github.com/asdf-vm/asdf/blob/master/lib/commands/plugin-list.sh#L11, so perhaps it's better to change it in both places?

Stratus3D: Yes, but I suppose I can merge these changes as they are. We always update this later.

Stratus3D:
Thanks for the updates. It looks like the build is failing, this time for a different test:

```
not ok 37 install command without arguments installs versions from legacy files
# (in test file test/install_command.bats, line 161)
#   `[ "$status" -eq 0 ]' failed
```

Stratus3D:
Thanks for the PR @leonid-shevtsov! I added one commit to your branch to fix the test and then merged it (https://github.com/asdf-vm/asdf/pull/539).

PR #537, opened 2019-06-06
Add Homebrew instructions for ZSH

HeroicEric:
# Summary

Adds instructions for adding asdf to your shell when using zsh and Homebrew.

This was mostly copied from https://github.com/asdf-vm/asdf/pull/496

I believe similar instructions should also be added for fish but I don't use it and cannot test it at the moment.


skatkov:
This was helpful (I managed to test this), I vouch that this should be merged ! ;-)

tylerbarker:
Worth noting there is a typo in this commit on line 85 in which `.zshrc` is spelt `zshrc.`

HeroicEric:
thanks @tylerbarker, fixed now

danhper:
Looks good, thanks a lot!

PR #71, opened 2016-06-29
zsh: remove _asdf completion.

dbernheisel:
Should help with #68.

PR #544, opened 2019-07-01
zsh-native asdf completion setup

philpennock:
# Summary

Add a zsh-native completion system function for `asdf`.

I too often hit snags with bash completions in zsh, so I bias strongly towards zsh-native completions.

This needs to end up in `$fpath` with a filename of `_asdf` but it seemed unfriendly to use that name in this repo, so call it `_asdf.zsh` and document a symlink.  Mark the file as executable to be compatible with one convention of "personal functions should only autoload if executable" (but that's probably so historic that very few people still use it; I'm one of them, though).


danhper:
Hi, thanks for the PR!
I do not use zsh so I cannot really comment on this.
@asdf-vm/core Anybody using zsh?

HashNuke:
I use zsh as my shell. Not the completions though. Will try it out and post if any issues.

nomasprime:
Looks good…why isn't it merged already? 😄

philpennock:
Since somebody nudged on this ticket, I just took a look and updated the PR with a small bug-fix from my local install.

jthegedus:
@nomasprime We need more people to test and validate.

I will give this a final one this coming weekend 👍

nomasprime:
@philpennock @jthegedus sweet 👌 Thanks, guys.

jthegedus:
Hey @philpennock I got this to work nicely, but would like a little more explanation in the docs addition about the `fpath` modification. I've no idea how `fpath` works and it seemed off to symlink this file to a random item in the `fpath` list.
EG: My system's `fpath` is:

```
/home/jthegedus/.oh-my-zsh/custom/plugins/zsh-syntax-highlighting
/home/jthegedus/.oh-my-zsh/plugins/pip
/home/jthegedus/.oh-my-zsh/plugins/python
/home/jthegedus/.oh-my-zsh/plugins/npx
/home/jthegedus/.oh-my-zsh/plugins/npm
/home/jthegedus/.oh-my-zsh/plugins/node
/home/jthegedus/.oh-my-zsh/plugins/go
/home/jthegedus/.oh-my-zsh/plugins/git
/home/jthegedus/.oh-my-zsh/plugins/gcloud
/home/jthegedus/.oh-my-zsh/plugins/docker
/home/jthegedus/.oh-my-zsh/functions
/home/jthegedus/.oh-my-zsh/completions
/usr/local/share/zsh/site-functions
/usr/share/zsh/vendor-functions
/usr/share/zsh/vendor-completions
/usr/share/zsh/functions/Calendar
/usr/share/zsh/functions/Chpwd
/usr/share/zsh/functions/Completion
/usr/share/zsh/functions/Completion/AIX
/usr/share/zsh/functions/Completion/BSD
/usr/share/zsh/functions/Completion/Base
/usr/share/zsh/functions/Completion/Cygwin
/usr/share/zsh/functions/Completion/Darwin
/usr/share/zsh/functions/Completion/Debian
/usr/share/zsh/functions/Completion/Linux
/usr/share/zsh/functions/Completion/Mandriva
/usr/share/zsh/functions/Completion/Redhat
/usr/share/zsh/functions/Completion/Solaris
/usr/share/zsh/functions/Completion/Unix
/usr/share/zsh/functions/Completion/X
/usr/share/zsh/functions/Completion/Zsh
/usr/share/zsh/functions/Completion/openSUSE
/usr/share/zsh/functions/Exceptions
/usr/share/zsh/functions/MIME
/usr/share/zsh/functions/Math
/usr/share/zsh/functions/Misc
/usr/share/zsh/functions/Newuser
/usr/share/zsh/functions/Prompts
/usr/share/zsh/functions/TCP
/usr/share/zsh/functions/VCS_Info
/usr/share/zsh/functions/VCS_Info/Backends
/usr/share/zsh/functions/Zftp
/usr/share/zsh/functions/Zle
```

Running `ln -s ~/.asdf/completions/_asdf.zsh ${fpath[1]}/_asdf` adds the symlink in 
`/home/jthegedus/.oh-my-zsh/custom/plugins/zsh-syntax-highlighting/_asdf`. It works, but seems odd and I've no context on how this is supposed to work.

Surprisingly Googling `fpath` wasn't very helpful. Please share you knowledge with me 😄

philpennock:
`fpath` is the lower-case array-tied form of `FPATH`, just as `path` is an array which updates and is updated by `PATH`.  `FPATH` is the function load path, where autoload looks for function definitions.  This is documented in man-page `zshparam(1)`.

My assumption was that the first item in fpath would be a place for users to drop their own functions.  I overly tersely documented that assumption with:
>If you have a function path set up and have a place to put a custom completion file,

Perhaps:
> zsh loads functions from `$FPATH`/`$fpath` and this completion should be symlinked to be available as filename `_asdf` in a directory in your `FPATH`; the variety of plugin module systems available for zsh means that we can't give solid advice here; if packaging _asdf_ then "wherever your packaging system guidelines say to place custom completions".  If installing for personal use, `print -l $fpath` and look for a suitable location.  Assuming that the very first directory is the right place, then run:

jthegedus:
@philpennock I just found that we used to have `_asdf` and automatically added the `.asdf/completions` dir to `$fpath` - https://github.com/asdf-vm/asdf/pull/71

I personally prefer the solution of adding the `.asdf/completions` as then there's no ambiguity of where the `_asdf` file should live and we automatically take care of the setup as part of the install process.

The issues around https://github.com/asdf-vm/asdf/issues/68 seem to have been around maintenance effort of the auto-completion. @danhper , given this project now has a larger community, what do you think about this now?

philpennock:
Prepending fpath seems sane to me.

jthegedus:
Will wait for feedback from other asdf members. @danhper @HashNuke @Stratus3D

asdf.sh
```diff
@@ -33,6 +33,7 @@ PATH="${ASDF_USER_SHIMS}:$PATH"
 . "${ASDF_DIR}/lib/asdf.sh"
 
 if [ -n "${ZSH_VERSION:-}" ]; then
-  autoload -U bashcompinit
-  bashcompinit
+  fpath=(${ASDF_DIR}/completions $fpath)
+  autoload -Uz compinit
+  compinit
```
jthegedus: Use `compinit` as per latest instructions in README.

in #69 we changed to `bashcompinit` for zsh users in this script, but then in #270 instructed users to use `compinit` in the README. We will go with the zsh communities latest suggestion unless someone raises an issue with this change?

jthegedus:
@philpennock Sorry for co-opting your branch. I made the following changes:

- add `${HOME}/.asdf/completions` to `fpath` as was in #71 
- swapped `bashcompinit` for `compinit` as we were telling people to use the latter anyway
- simplified the documentation as we no longer need to write edge cases for including completions

Concerns I have:

- Homebrew autocompletions conflicting with our new found ZSH completions

philpennock:
Rebased onto master.
Implemented support for git-style space-separated commands.
Added support for `asdf latest` and `latest` as a version specifier.

Stratus3D:
Thanks @philpennock ! Looks good to me.

asdf.sh
jthegedus: This was a bad decision. See #674 & fix in #675

PR #546, opened 2019-07-06
add quite flag to git clone

spencerdcarlson:
# Summary
We don't need all that noise 👂, it's getting kind of loud up in here.  😉
We will still see error messages if git-clone fails, but we don't have to see all the logs of everything getting cloned.

## Other Information

information here.


danhper:
Thanks for the PR!

PR #553, opened 2019-07-22
Docs: Homebrew setup: do not add brew commands to the shell profile

olets:
# Summary

The recommended Homebrew setup required running two `brew --prefix` commands
to the shell profile. `brew` commands can be slow, so this introduced a notable
lag when sourcing the shell profile.

Homebrew paths do not change, so we can determine the necessary paths one time
during setup and use static paths in the shell profile.

Also changes one `` ```bash`` to `` ```shell`` in the ZSH documentation, to match the rest of the ZSH documentation.


danhper:
Indeed, this is better. I'll change it for fish too, thanks!

Issue #509, opened 2019-04-09
plugin-test can't be run with asdf installed via brew


quiqua:
### Steps to reproduce

- Install via homebrew on mac os x: `brew install asdf`
- Add `source /usr/local/opt/asdf/asdf.sh` to `.zshrc` file
- Clone https://github.com/skotchpine/asdf-java 
- Run `asdf plugin-test java . 'java -version'

### Expected behavior

The command runs without errors

### Actual behavior

When installing via homebrew on mac os x: `brew install asdf` and adding `source /usr/local/opt/asdf/asdf.sh` to my `.zshrc` file, I am unable to run `asdf plugin-test` properly.

I am currently working on a fix for https://github.com/skotchpine/asdf-java and cloned that repo locally.
When testing I get the following output:

```
asdf plugin-test java . 'java -version'
fatal: repository '/usr/local/opt/asdf/.git' does not exist
/usr/local/opt/asdf/lib/commands/plugin-test.sh: line 49: /var/folders/5d/2m5v9bpx5fd6rc_fbfz1ffk80000gn/T/asdf.XXXX.78cJ5jmW/asdf.sh: No such file or directory
Cloning into '/var/folders/5d/2m5v9bpx5fd6rc_fbfz1ffk80000gn/T/asdf.XXXX.78cJ5jmW/plugins/java'...
done.
```

The errors here are I assume are due to not having a git repo available at that path:
```
➜ cd /usr/local/opt/asdf

local/opt/asdf
➜ ls -la
total 112
drwxr-xr-x  24 marcel  admin   768 Apr  5 09:39 .
drwxr-xr-x   3 marcel  admin    96 Apr  5 09:39 ..
drwxr-xr-x   3 marcel  admin    96 Apr  5 09:39 .brew
-rw-r--r--   1 marcel  admin  7998 Mar 31 22:15 CHANGELOG.md
-rw-r--r--   1 marcel  admin  1686 Mar 31 22:15 CONTRIBUTING.md
-rw-r--r--   1 marcel  admin   917 Apr  5 09:39 INSTALL_RECEIPT.json
-rw-r--r--   1 marcel  admin  1082 Mar 31 22:15 LICENSE
-rw-r--r--   1 marcel  admin  1574 Mar 31 22:15 README.md
-rw-r--r--   1 marcel  admin     7 Mar 31 22:15 VERSION
-rw-r--r--   1 marcel  admin  3027 Mar 31 22:15 Vagrantfile
[...]

PR #556, opened 2019-07-31
Fix for asdf plugin-test when installed with brew, and a minor logic fix to a warning

klundberg:
# Summary

This makes 2 changes:
1. Provides an alternative way for `asdf plugin-test` to set up its environment, by copying the contents of $ASDF_DIR to $TEST_DIR instead of doing a git clone. This fixes `plugin-test` for homebrew installations
2. Makes a small logic fix to silence a warning when GITHUB_API_TOKEN or OAUTH_TOKEN are present, but not both (before both needed to be set before the warning would disappear, which seems incorrect).

Fixes: #509

## Other Information

This may be an inadequate solution if someone installs asdf such that $ASDF_DIR is equal to $ASDF_DATA_DIR, since all installed plugins and packages will also be copied which could be wasteful of time and resources. I don't know if that's typical or not of a pure git installation as I haven't done that myself, though.

### DIFF
--- a/lib/commands/plugin-test.sh
+++ b/lib/commands/plugin-test.sh
@@ -39,7 +39,9 @@ plugin_test_command() {
     fi
 
     TEST_DIR=$(mktemp -dt asdf.XXXX)
-    cp -R "$ASDF_DIR/" "$TEST_DIR"
+    cp -R "$ASDF_DIR/bin" "$TEST_DIR"
+    cp -R "$ASDF_DIR/lib" "$TEST_DIR"
+    cp -R "$ASDF_DIR/asdf.sh" "$TEST_DIR"
 
     plugin_test() {
         export ASDF_DIR=$TEST_DIR
--- a/lib/commands/plugin-test.sh
+++ b/lib/commands/plugin-test.sh
@@ -39,7 +39,7 @@ plugin_test_command() {
     fi
 
     TEST_DIR=$(mktemp -dt asdf.XXXX)
-    git clone "$ASDF_DIR/.git" "$TEST_DIR"
+    cp -R "$ASDF_DIR/" "$TEST_DIR"
 
     plugin_test() {
         export ASDF_DIR=$TEST_DIR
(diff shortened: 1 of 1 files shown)


lib/commands/plugin-test.sh
```diff
@@ -39,7 +39,7 @@ plugin_test_command() {
     fi
 
     TEST_DIR=$(mktemp -dt asdf.XXXX)
-    git clone "$ASDF_DIR/.git" "$TEST_DIR"
+    cp -R "$ASDF_DIR/" "$TEST_DIR"
```
Stratus3D: These two commands are not equivalent Git clone only copies over the files that are checked in. It is has ignore rules for directories like `installs`, `plugins`, and `shims`.

Stratus3D: Running this locally it tries to copy everything in `~/.asdf`, which is a lot of install and plugin data.

Stratus3D: I'd suggest copying over just the things we need. `bin`, `lib`, `test`, etc...

klundberg: Yup, that was my fear. brew installs don't have $ASDF_DIR and $ASDF_DATA_DIR in the same spot. I'll make that change when I have the chance.

lib/commands/plugin-test.sh
```diff
@@ -39,7 +39,9 @@ plugin_test_command() {
     fi
 
     TEST_DIR=$(mktemp -dt asdf.XXXX)
-    git clone "$ASDF_DIR/.git" "$TEST_DIR"
+    cp -R "$ASDF_DIR/bin" "$TEST_DIR"
+    cp -R "$ASDF_DIR/lib" "$TEST_DIR"
+    cp -R "$ASDF_DIR/asdf.sh" "$TEST_DIR"
```
klundberg: @Stratus3D This seems to me to be enough to copy over, at least in my rudimentary testing of a couple things, did I miss anything?

Stratus3D: Looks good to me.

Stratus3D:
Thanks for the PR!

PR #561, opened 2019-08-15
Fix PATH construction on fish

lalloni:
When prepending asdf_bin_dirs to $PATH on asdf startup script for fish shell
there was a problem when PATH elements contained blank characters.

This patch gets it working again and also improves on simplicity and
performance by not spawning external processes for simple strings
handling stuff.

# Summary

## Other Information

information here.

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -10,7 +10,7 @@ set -l asdf_bin_dirs $ASDF_DIR/bin $ASDF_DIR/shims $asdf_data_dir/shims
 
 for x in $asdf_bin_dirs
   if test -d $x
-    set PATH $x (echo $PATH | command xargs printf '%s\n' | command grep -v $x)
+    set PATH $x (string match -v $x $PATH)
   end
 end
 


Stratus3D:
Thanks for the PR @lalloni. @danhper can you test these changes out when you get a chance? I don't use fish shell.

Stratus3D:
Actually this seems very similar to #551.

lalloni:
@Stratus3D you're right. I just hit the problem, solved it and sent a PR without checking the issue tracker, sorry for that.
I guess this also means the problem is somewhat prevalent.
Although this patch is much simpler and smaller.
Is still a fish 2.2 support a requirement? (2.3.0 came out on May 2016!).

danhper:
@lalloni Hi, and thanks for the PR.

Yes, Ubuntu 16.04, which has not reached EOL yet, uses fish 2.2 by default, so I would rather keep supporting fish 2.2, especially given the cost is really not that high.
Let's merge #551 for now.

PR #563, opened 2019-08-17
Add --unset option to shell subcommand

sakuro:
# Summary

Issue #378 proposes the `shell` subcommand and PR #480 implements it.

But #378 also mentions about its `--unset` option to unset ASDF_WHATEVER_VERSION variable which is missing from PR #480.

This PR adds support of `--unset`.

## Example

before
```shell
$ asdf current ruby
2.7.0-dev (set by /Users/sakuro/.tool-versions)
$ asdf shell ruby 2.6.3
$ asdf current ruby
2.6.3    (set by ASDF_RUBY_VERSION environment variable)
$ asdf shell ruby --unset
version --unset is not installed for ruby
```

after
```shell
$ asdf shell ruby
Usage: asdf shell <name> <version> | --unset
$ asdf current ruby
2.7.0-dev (set by /Users/sakuro/.tool-versions)
$ asdf shell ruby 2.6.3
$ asdf current ruby
2.6.3    (set by ASDF_RUBY_VERSION environment variable)
$ asdf shell ruby --unset
$ asdf current ruby
2.7.0-dev (set by /Users/sakuro/.tool-versions)
```

## Other Information

I've moved the assignment to `version_env_var` up because `--unset` needs the variable name to unset.

### DIFF
--- a/lib/commands/version_commands.sh
+++ b/lib/commands/version_commands.sh
@@ -81,7 +81,7 @@ global_command() {
 # Output from this command must be executable shell code
 shell_command() {
   if [ "$#" -lt "2" ]; then
-    echo "Usage: asdf shell <name> <version>" >&2
+    echo "Usage: asdf shell <name> <version> | --unset" >&2
     echo 'false'
     exit 1
   fi
@@ -89,14 +89,18 @@ shell_command() {
   local plugin=$1
   local version=$2
 
+  local upcase_name
+  upcase_name=$(echo "$plugin" | tr '[:lower:]-' '[:upper:]_')
+  local version_env_var="ASDF_${upcase_name}_VERSION"
+
+  if [ "$version" = "--unset" ]; then
+    echo "unset $version_env_var"
+    exit 0
+  fi
   if ! (check_if_version_exists "$plugin" "$version"); then
     echo 'false'
     exit 1
   fi
-
-  local upcase_name
-  upcase_name=$(echo "$plugin" | tr '[:lower:]-' '[:upper:]_')
-  local version_env_var="ASDF_${upcase_name}_VERSION"
 
   case $ASDF_SHELL in
     fish )
--- a/lib/commands/version_commands.sh
+++ b/lib/commands/version_commands.sh
@@ -81,7 +81,7 @@ global_command() {
 # Output from this command must be executable shell code
 shell_command() {
   if [ "$#" -lt "2" ]; then
-    echo "Usage: asdf shell <name> <version> | --unset" >&2
+    echo "Usage: asdf shell <name> {<version>|--unset}" >&2
     echo 'false'
     exit 1
   fi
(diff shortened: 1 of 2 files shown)


lib/commands/version_commands.sh
```diff
@@ -81,23 +81,27 @@ global_command() {
 # Output from this command must be executable shell code
 shell_command() {
   if [ "$#" -lt "2" ]; then
-    echo "Usage: asdf shell <name> <version>" >&2
+    echo "Usage: asdf shell <name> <version> | --unset" >&2
```
Stratus3D: I'm trying to remember what syntax is typically used for mutually exclusive parameters. I think this might be clearer: `asdf shell <name> (<version>|--unset)`.

test/version_commands.bats
```diff
@@ -238,6 +238,15 @@ teardown() {
   unset ASDF_DUMMY_VERSION
 }
 
+@test "shell wrapper function with --unset should unset ENV var" {
+  source $(dirname "$BATS_TEST_DIRNAME")/asdf.sh
+  asdf shell "dummy" "1.1.0"
+  [ $(echo $ASDF_DUMMY_VERSION) = "1.1.0" ]
+  asdf shell "dummy" "--unset"
```
Stratus3D: Just curious, why the quotes around these arguments? Flags are typically not quoted.

Stratus3D:
Thanks for the PR, and apologies for the late review. This looks like a nice feature to have. I will merge if you can address that first comment of mine. Thanks!

lib/commands/version_commands.sh
sakuro: Just see if removing spaces is sufficient

```
asdf shell                                                                            
Usage: asdf shell <name> <version>|--unset
```

Do you think surrounding parenthesis needed?

test/version_commands.bats
sakuro: Ah! I've copied and edited the line above:) I'll remove quotes from `--unset`

lib/commands/version_commands.sh
Stratus3D: Well, now that I'm looking into it, it seems like something is needed, but maybe `{<version>|--unset}` is the correct syntax. Here is what I found:

IBM: `{...|...}` - https://www.ibm.com/support/knowledgecenter/en/STAV45/com.ibm.sonas.doc/sonas_manpages_syntax.html
Microsoft: `(...|...)` - https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/command-line-syntax-key
docopt: `(...|...)` - http://docopt.org/

sakuro: Thank you! I'll adopt `{...|...}`

sakuro: done in b26f566

test/version_commands.bats
sakuro: done in bc309c0

Stratus3D:
Thanks!

Issue #565, opened 2019-08-29
Crash when running global node command


ushu:
### Steps to reproduce

install Node.js (in my tests 12.4.0):

```sh
$ asdf current
nodejs         12.4.0   (set by /Users/ushu/.tool-versions)
```

Then globally install `standard-version`:

```sh
$ npm install -f standard-version
# or with yarn (crashes the same)
$ yarn global add standard-version
```

### Expected behavior

It should run the executable, eg:

```sh
$ standard-version 
✔ outputting changes to CHANGELOG.md
✔ committing CHANGELOG.md
✔ tagging release v0.7.5
```

### Actual behavior

It crashes at line 690 of utils.sh:

```sh
$ standard-version 
/usr/local/opt/asdf/lib/utils.sh: ligne 690: executable_path: 12.4.0 : erreur de syntaxe : opérateur arithmétique non valable (le symbole erroné est « .4.0 »)
No standard-version executable found for nodejs 12.4.0
```
(sorry my shell is in French...)

From a few test, I could determine it is due to this line: https://github.com/asdf-vm/asdf/blob/master/lib/utils.sh#L690

And more precisely this expression: `"${shim_name}"`.
Since there `shim_name="standard-version"`, it seems that bash interprets the dash (`-`) in the "shim name" as a "minus operator" so it breaks everything...

### Environment

**OS**:

macOS 10.14.5 (18F96h)

```sh
Darwin IAPP-MBP-15-2014.local 18.6.0 Darwin Kernel Version 18.6.0: Thu Mar 14 21:02:45 PDT 2019; root:xnu-4903.260.65~14/RELEASE_X86_64 x86_64
```

**asdf version**:

```sh
$ asdf --version
v0.7.4
```

Stratus3D:
Interesting. We've not encountered any errors like this before.

Stratus3D:
I think I have fixed this in https://github.com/asdf-vm/asdf/pull/589. Can you give that a try?

PR #566, opened 2019-08-29
Fix crash when running shimmed command with a "-"

ushu:
# Summary

When trying to run `standard-version` as a global npm command, crashes happened/

From a few test, I could determine it is due to this line: https://github.com/asdf-vm/asdf/blob/master/lib/utils.sh#L690

And more precisely this expression: `"...:${shim_name}"`.
It seems that bash interprets the dash (`-`) in the "shim name" as a "minus operator" and raises an error.

This patch removes the expressions that causes the crash.

Fixes: #565 

## Other Information

I'm no bash guru, so I might have missed something here...
Please provide guidance 😁

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -687,7 +687,7 @@ with_shim_executable() {
 
       if [ -x "${plugin_path}/bin/exec-path" ]; then
         install_path=$(find_install_path "$plugin_name" "$full_version")
-        executable_path=$(get_custom_executable_path "${plugin_path}" "${install_path}" "${executable_path:${shim_name}}")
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR! I'd like to have unit tests for this fix as it's an edge case that nobody caught until now. I can write a unit test before merging it.

Stratus3D:
I think we may need to keep the default value of shim name for the third argument to that function call. I think the syntax for the variable was just wrong. I've created another PR to address this - https://github.com/asdf-vm/asdf/pull/589.

Stratus3D:
This has been solved by #589

PR #579, opened 2019-10-16
Add plugin-disable and plugin-enable commands

tjarvstrand:
# Summary
With this change it will be possible to disable specific plugins. This can be useful if you sometimes want to run pre-release versions, private forks etc. that are installed outside of asdf

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -103,6 +103,14 @@ case "$1" in
     load_cmd "plugin-test"
     plugin_test_command "${cmd_args[@]}";;
 
+  "plugin-disable")
+    load_cmd "plugin-disable"
+    plugin_disable_command "${cmd_args[@]}";;
+
+  "plugin-enable")
+    load_cmd "plugin-enable" "reshim"
+    plugin_enable_command "${cmd_args[@]}";;
+
   "shim-versions")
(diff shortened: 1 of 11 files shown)


tjarvstrand:
Any opinions on this?

The build is reported as still ongoing even though it actually passed

vic:
> This can be useful if you sometimes want to run pre-release versions, private forks etc. that are installed outside of asdf

Not sure I understand what's the purpose of _disabling_ a plugin. Could you provide a bit more of context on it's use case?.

danhper:
Sorry for the delay.
I am not exactly sure I understand the use case either.
To bypass asdf, we already have the `system` version, which can be set either for the current shell, directory, or globally, as any other version.
I would assume it is enough to test any sort of pre-release or fork.
@tjarvstrand  Could you show us a use case that is not solvable using what asdf currently offers please? Thanks!

tjarvstrand:
Ah, I hadn't seen the `system` version. Is that documented anywhere?

That solves my precise use case so I'll close this.

Stratus3D:
Thanks for this PR @tjarvstrand! I think closing this was the right choice though, as this feature is kind of redundant with the versioning features we already have.

Issue #582, opened 2019-10-22
Updating plugins: cd: /root/.asdf/plugins/*: No such file or directory when 


android10:
### Steps to reproduce

1 - Install asdf, let's say under a regular user: fernando
2 - In Arch Linux or other distros it is possible to run hooks after upgrading the system, in this case with pacman which runs as root, so when executing sudo asdf plugin-update --all
3 - Plugin list cannot be updated. 

### Expected behavior

```bash
Updating java...
Already up to date.
Updating nodejs...
Already up to date.
Updating python...
Already up to date.
Updating ruby...
Already up to date.
Updating sbt...
Already up to date.
Updating scala...
Already up to date.
```

### Actual behavior

An error ocurrs:
```bash
Updating *...
/home/fernando/.asdf/lib/commands/plugin-update.sh: line 11: 
cd: /root/.asdf/plugins/*: No such file or directory
```

### Environment

**OS**: Arch Linux

**asdf version**: v.0.7.4

### Pull Request
Here is the solution to the problem: #583

Stratus3D:
If you are running as another user you'll need to set ASDF_DATA_DIR to the correct directory. e.g. `export ASDF_DATA_DIR=/home/otheruser/.asdf`.

PR #583, opened 2019-10-22
asdf_data_dir() returning incorrect path when running as root.

android10:
# Summary

In some Linux distributions it is possible to run hooks after upgrading the system. These hooks, in the case of [pacman](https://wiki.archlinux.org/index.php/pacman), which is the package manager for [Arch Linux](https://www.archlinux.org/), are run as root. 

With that in mind, when plugins are updated, several calls to the ```asdf_data_dir()``` are processed, returning a path containing ```$HOME``` environment variable, which is incorrect since the current user executing the update is different from the one who installed ```asdf``` in the home directory. 
Also at this point ```$ASDF_DATA_DIR``` is empty (for some reason has not been set). 

This fix reuses the existing method: ```asdf_dir()``` in order to retrieve the correct path where the plugins are installed. 

Fixes: #582

## Other Information

For some reason I could not see in the code where the ```$ASDF_DATA_DIR``` is being set. The method ```asdf_dir()``` indeed is setting ```$ASDF_DIR``` as environment variable which is being used but not in the case of ```$ASDF_DATA_DIR```.

Should we also replace the usage of ```$ASDF_DATA_DIR``` with calls to the mentioned ```asdf_data_dir()``` method? 

Maybe in another PR?

```bash
asdf_data_dir(){
  local data_dir="$(asdf_dir)"

  if [ -z "$data_dir" ]; then
    data_dir="$HOME/.asdf"
  fi

  echo "$data_dir"
}
```

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -30,7 +30,7 @@ asdf_data_dir(){
 
   if [ -n "${ASDF_DATA_DIR}" ]; then
     data_dir="${ASDF_DATA_DIR}"
-  elif [ asdf_run_as_root ]; then
+  elif [ "${asdf_run_as_root}" ]; then
     data_dir="$(asdf_dir)"
   else
     data_dir="$HOME/.asdf"
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -27,20 +27,16 @@ asdf_repository_url() {
 
 asdf_data_dir(){
   local data_dir
-
+  
   if [ -n "${ASDF_DATA_DIR}" ]; then
     data_dir="${ASDF_DATA_DIR}"
-  elif [ "${asdf_run_as_root}" ]; then
-    data_dir="$(asdf_dir)"
-  else
+  elif [[ $EUID -ne 0 ]]; then
     data_dir="$HOME/.asdf"
+  else
+    data_dir="$(asdf_dir)"
   fi
 
   echo "$data_dir"
-}
-
-asdf_run_as_root() {
-  [ $(id -u) -eq 0 ]
 }
 
 get_install_path() {
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -26,9 +26,13 @@ asdf_repository_url() {
 }
 
 asdf_data_dir(){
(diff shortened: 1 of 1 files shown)


Stratus3D:
@android10 `ASDF_DATA_DIR` is only set by the user. I don't see any other uses of `ASDF_DATA_DIR` outside of this function, aside from in the plugin test command, which can stay as it is. Can you write tests for this? I'm still not sure I understand what this is suppose to be doing. All our existing tests pass without these changes.

android10:
@Stratus3D I will add some tests to it and here is the basic explanation to the problem. 

Arch Linux users use pacman as a package manager. You can add hooks to it in order to upgrade specific components when a system upgrade happens. For example, in this use case scenario, every time a system upgrade happens, ```asdf``` should be automatically upgrade. 

With that being said, these hooks are run as root and without that check, the update will never happens since the path for trying to find the ```asdf update``` command is wrong. 

Let me also see if I can upload a video to showcase and make it even more clear of what I'm talking about.

android10:
@Stratus3D I just wanted to follow up with this since this was specific to some linux distros like Arch.

The code for this PR is no longer necessary, and actually I tried to revert it but it was previously removed, which is great:

https://github.com/asdf-vm/asdf/blob/7cdfc0fa2a5f0f924a4486a971f5110102f58e54/lib/utils.bash#L39-L49

At least on Arch Linux we have now the opportunity to install/upgrade ```asdf``` via package manager:
- https://aur.archlinux.org/packages/asdf-vm/
- https://aur.archlinux.org/packages/asdf/

It is nice to see progress on this amazing tool and apologize for the noise, just wanted to make sure the project is clean and we use a boy scout approach. 

Cheers!

Stratus3D:
@android10 thanks for following up. I eventually wrote unit tests for this functionality and found one of the clauses to not be necessary, so I ended up removing it myself. If there is a scenario our tests don't cover please submit a test for it and we will get that scenario fixed.

Great news on the archlinux package! I'd love to see well maintained asdf-vm packages for all popular operating systems.

PR #595, opened 2019-11-13
Fix completion suggestions for OSX / fish

jmaroeder:
# Summary

Fixes issue with suggestions when running `fish` on OSX.

## Other Information

OSX `sed` is not GNU `sed`, and doesn't support the `\s` whitespace character sequence. This results in autocompletion suggestions including leading spaces, e.g.,
```
$ asdf local python \ \ 3.8.0
```

Using the POSIX compliant `[[:space:]]` sequence is more compatible, and removes the unnecessary spaces on OSX.

### DIFF
--- a/completions/asdf.fish
+++ b/completions/asdf.fish
@@ -31,7 +31,7 @@ function __fish_asdf_arg_at -a number
 end
 
 function __fish_asdf_list_versions -a plugin
-    asdf list $plugin 2> /dev/null | sed -e 's/^\s*//'
+    asdf list $plugin 2> /dev/null | sed -e 's/^[[:space:]]*//'
 end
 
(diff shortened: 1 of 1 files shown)


Stratus3D:
@jmaroeder looks good to me, but I don't use Fish. @danhper can you test this?

We can ignore the failing appveyor build for now, as we've not yet had a successful build on Windows.

danhper:
@Stratus3D @jmaroeder Sorry for the delay. Thanks for the PR, indeed, I completely forgot that `\s` in sed was a GNU thing.

Stratus3D:
I should look into unit testing our completion scripts. This is one of the few areas in the codebase where we don't have any tests.

PR #597, opened 2019-11-25
New feature: asdf extension commands

vic:
When a plugin provides custom executables or bash-scripts under its `bin/` directory,
   it's now possible to invoke them with:
   
   ```shell
   asdf nodejs import-release-team-keyring

   # instead of people having to know where the executable is installed:
   ~/.asdf/plugins/nodejs/bin/import-release-team-keyring
   ```
   
   As in nodejs case, plugin authors can use this feature to provide utilities,
   or people can create plugins that are just extensions to asdf itself.

   See `docs/plugins-create.md` for more info.

Kudos to @imbsky for the idea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@
 
 Features
 
+* Add support for asdf extension commands.
+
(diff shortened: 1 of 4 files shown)


imbsky:
👏

PR #601, opened 2019-11-27
Subcommand CLI support.

vic:
Users familiar with sub-command aware tools like `git` can now
   use `asdf` commands in the same way. For example:

   `asdf plugin list all` is equivalent to `asdf plugin-list-all`

   This is also the case for plugin extension commands, where the
   plugin name is an asdf main subcommand. ie. Having a `foo` plugin
   you can invoke: `asdf foo bar`

   See `docs/plugins-create.md` for more info.


lib/commands/command-install
```diff
@@ -95,9 +97,8 @@ install_tool_version() {
     local install_type="version"
 
     if [ "${version_info[0]}" = "latest" ]; then
-      load_cmd "latest"
       local version
-      version=$(latest_command "$plugin_name" "${version_info[1]}")
+      version=$(asdf latest "$plugin_name" "${version_info[1]}")
```
Stratus3D: Does it make sense to invoke asdf again in this context? It seems like with the command-finding logic this would be slower than just invoking the `latest_command` function directly.

Stratus3D:
Thanks for all your work on this @vic ! Finally got around to reading the code.

PR #605, opened 2019-12-01
Fix shell compatibility with ZSH

deiga:
`command eval` does not work in ZSH

# Summary

Replace `command eval` with just `eval`


Stratus3D:
The build for this PR was failing and broke the build. I do not think this should have been merged.

vic:
The only failing step was appveyor but that one has never been successfull. 
Screen Shot 2019-12-04 at 12 53 50

PR #606, opened 2019-12-01
Replaces Appveyor with Github Actions

deiga:
# Summary

Removes Appveyor config and adds Windows to Actions ocnfig

## Other Information

information here.


deiga:
This is silly and unnecessary. Why would BASH behave differently on Windows, macOS or Linux?

PR #609, opened 2019-12-03
Clarify instructions when <version> of a <shim_name> is missing.

tejanium:
# Summary

Often time when working on multiple projects and multiple versions, the version that specified in `.tool-versions` or legacy config like `.ruby-version` is not yet installed in the system. When that happens this is what we see

Screenshot 2019-12-03 at 11 47 48

So instead of telling us what version is missing, asdf tell us to change the configuration into one that has been installed in our system.

This PR changes the warning by giving us information about which version is missing, and option to install the missing version or replace existing project to use different version that is available.

Screenshot 2019-12-03 at 14 10 55

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -711,16 +711,23 @@ with_shim_executable() {
   fi
 
   (
-    local version
-    version=$(get_preset_version_for "$shim_name")
-    echo "asdf: No version ${version} installed for command ${shim_name}"
-    echo "Please install the missing version by running"
-    echo ""
-    echo "asdf install ${shim_name} ${version}"
-    echo ""
-    echo "or add one of the following in your .tool-versions file:"
+    local preset_version
+    preset_version=$(get_preset_version_for "$shim_name")
+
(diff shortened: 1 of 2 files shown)


vic:
Nice, thank you!

Could you please fix the failing tests so we can merge?

tejanium:
I see `master` is also failing, I'll see what I can do

vic:
Ignore the appveyor build, just try make sure the ubuntu and macOS tests are green :)

tejanium:
After looking at the test, I realized that we have to handle 2 different scenarios:
1. when `.tool-versions` is not set/empty
https://github.com/asdf-vm/asdf/blob/cc004b8dbb28ec1a980f66bc1480d8143b799f3f/test/shim_exec.bats#L73
I'm keeping the warning as is

2. when `.tool-versions` set version that is not installed 
I've added a new test and new warning for this

Please let me know what you think!

vic:
Thank you! 🎉

PR #610, opened 2019-12-05
Fix errors in list-all command documentation

matt17r:
# Summary

Documentation fix

Fixes: N/A

## Other Information

Need `list-all` rather than `list all`:

```
> asdf list all nodejs
No such plugin: all
```

PR #615, opened 2019-12-10
Change 'list all' to 'list-all'

chanhman:
When I ran `asdf list all mysql ` I got `No such plugin: all`. When I checked out their docs, I noticed they wrote `asdf list-all mysql`, and it worked.


vic:
Dupe of #610

PR #616, opened 2019-12-10
Update core-manage-versions.md

sankalpk:
Updating documentation for deprecated commands.

# Summary

Update documentation for managing language versions. `list all` command does not seem to be supported any more and instead we use the `list-all` command.

In addition `asdf latest` does not seem to be supported as well. I've tested the command against Nodejs and Ruby.

### DIFF
--- a/docs/core-manage-versions.md
+++ b/docs/core-manage-versions.md
@@ -31,29 +31,8 @@ asdf list <name>
 ## List All Available Versions
 
 ```shell
-asdf list all <name>
+asdf list-all <name>
 # asdf list all erlang
-```
-
-Limit versions to those that begin with a given string.
-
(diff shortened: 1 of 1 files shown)


sankalpk:
@vic Any idea why tests are failing?

vic:
Actually `asdf list-all` and `asdf list all` are exactly the same and both should work on latest master, dashes can be used as spaces now. And `asdf latest` was not deprecated.

The `appveyor` build has always been failing, we still have to check what to do on windows to get green tests. All other checks (the good ones) are skipped if you only change documentation (.md files) like in this case.

anyways, I'm closing this since it's not a valid issue, and all commands mentioned (with or without dashes) should work.

PR #622, opened 2019-12-20
Fix issue with preset version warning assuming that the shim name and plugin name are the same

CodeFoodPixels:
# Summary
In #609 a clearer error message was added when you're trying to run a shimmed command, but you don't have a correct version. This PR fixes an issue where that code was assuming that the shim and plugin share a name.

An example of where this is not the case is the "nodejs" plugin, as the shim is installed as "node"

In other areas of the utils file, it works on the basis that there may be multiple plugins that provide for that shim, so I have worked on the same basis

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -578,6 +578,19 @@ shim_plugin_versions() {
   fi
 }
 
+shim_plugins() {
+  local executable_name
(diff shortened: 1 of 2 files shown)


vic:
@CodeFoodPixels thank you !

PR #623, opened 2019-12-21
Ignore shim directory for executable lookups

jonmast:
# Summary

Previously, the shims directory was only excluded for system versions, which meant that `asdf exec` would fall back to a shim if the install didn't have the desired binary. If the shims are up to date this isn't a problem since we check the metadata, but if those were out of sync the command would just infinitely recurse.

I'm not sure what the reason was for the current setup, but I looked through the uses of the function and I'm not seeing anywhere it is required.

Fixes: No issue

## Other Information

I also made a small tweak to the config test to make it pass, it was picking up my user `.asdfrc`.

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -674,10 +674,8 @@ with_shim_executable() {
     plugin_path=$(get_plugin_path "$plugin_name")
 
     run_within_env() {
-      local path=$PATH
-      if [ "system" == "$full_version" ]; then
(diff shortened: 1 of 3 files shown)


danhper:
Thanks for the PR!
I think this is a reminiscence of how shims worked a while ago (they were invoking a different wrapper script).
@vic Could you have a look when you have a moment please? Thanks!

vic:
LGTM

Issue #366, opened 2018-10-01
asdf global has issues with relative symbolic links


liamdawson:
### Steps to reproduce

1. symlink `.tool-versions` from a subdir of `$HOME` (I used stow)
2. `cd` to a directory that is not `$HOME`
3. `asdf global ruby 2.5.1`

### Expected behavior

Definition for `ruby 2.5.1` added to the file symlinked under `$HOME/.tool-versions`

### Actual behavior

```
/home/user/.local/share/asdf/lib/commands/version_commands.sh: line 37: relative/dir/.tool-versions: No such file or directory
```

### Environment

**OS**: Kubuntu 18.04

**asdf version**: v0.5.1

liamdawson:
Notably, running this from `$HOME` works because the relative link works.

Stratus3D:
This looks like something I missed in https://github.com/asdf-vm/asdf/pull/337.

Stratus3D:
It looks like my symlink resolution code only works with absolute symlinks. Relative symlinks do not work.

craigfurman:
I have this same issue in v0.7.6

craigfurman:
@Stratus3D I opened https://github.com/asdf-vm/asdf/pull/625 to fix this, and this appears to work in general, but I'm missing context on why asdf needs to resolve symlinks at all?

Would it not be simpler to let the filesystem do the work, using a symlink path in https://github.com/asdf-vm/asdf/blob/master/lib/commands/version_commands.sh#L40-L45?

alecho:
As a workaround I was able to set `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` to the version in my dot-file directory and it now works as expected.

I also worked on a patch for this and my solution was similar to @craigfurman. The documentation for GNU stow says:

> A relative symlink names a relative path; that is, one not starting from /. The target of a relative symlink is computed starting from the symlink’s own directory. Stow only creates relative symlinks.

Seems like we need to `dirname` the symlink and prepend that to the directory the symlink points to as opposed to `PWD`. When looking to see what POSIX says about this all I could find was that

> If the pathname does not begin with a slash, the predecessor of the first filename of the pathname shall be taken to be the current working directory of the process (such pathnames are referred to as "relative pathnames").

I'm not sure if this is an macOX issue, asdf, or something else.

PR #625, opened 2019-12-25
Fix relative symlink path resolution

craigfurman:
# Summary

I've only just started using asdf, and (similarly to https://github.com/asdf-vm/asdf/issues/366), I have `~/.tool-versions` as a symlink to a config repository of mine. Running `asdf global global <plugin> <version>` fails, apparently due to a bug in relative symlink resolution.

I've modified the relative symlink test, and verified that it fails without my change to utils.sh.

Fixes: https://github.com/asdf-vm/asdf/issues/366

Hope this helps! :slightly_smiling_face:

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -450,12 +450,16 @@ resolve_symlink() {
   # If it is a slash we can assume it's root and absolute. Otherwise we treat it
   # as relative
   case $resolved_path in
-  /*)
-    echo "$resolved_path"
-    ;;
-  *)
-    echo "$PWD/$resolved_path"
-    ;;
+    /*)
+      echo "$resolved_path"
+      ;;
+    *)
+      (
+        # shellcheck disable=SC2164
+        cd "$(dirname "$symlink")"
+        echo "$PWD/$resolved_path"
+      )
+      ;;
   esac
 }
 
--- a/test/utils.bats
+++ b/test/utils.bats
@@ -361,14 +361,15 @@ teardown() {
 }
 
 @test "resolve_symlink converts relative symlink path to the real file path" {
+  mkdir dir
   touch foo
-  ln -s foo bar
+  ln -s ../foo dir/bar
 
-  run resolve_symlink bar
+  run resolve_symlink dir/bar
   [ "$status" -eq 0 ]
   echo $status
-  [ "$output" = $(pwd)/foo ]
-  rm -f foo bar
+  [ "$output" = $(pwd)/dir/../foo ]
+  rm -rf dir foo
 }
 
 @test "strip_tool_version_comments removes lines that only contain comments" {


danhper:
Interesting, I never had the issue but I suppose it depends on what the symlink looks like.
Could you please just send the output of `ls -l ~/.tool-versions`, or anything which would allow use to see which case fails?
Thanks!

craigfurman:
@danhper this should be reproducible when the target of a relative symlink is not resolvable from your current directory. For example:

```
$ ls -l ~/.tool-versions
lrwxrwxrwx 1 craig craig 49 Dec 25 11:59 /home/craig/.tool-versions -> workspace/ansible-home/config/home/.tool-versions
```

If $PWD is not `/home/craig` in the above example, the current symlink resolution code in master will not work.

Out of curiosity, why does asdf need to resolve symlinks anyway? Can it not just open `.tool-versions` files regardless of whether they're symlinks?

craigfurman:
@danhper sorry for the hassle, but have you had a chance to try this out yet?

craigfurman:
@Stratus3D I'm not sure if you've seen this. I'm currently dependent on my fork in order for `asdf global` to resolve relative symlinks correctly.

Now that I look at this again, I'm not sure why asdf needs to resolve symlinks at all? We could trust the filesystem, unless there's a specific reason not to.

aca:
Anyone who can check this?
I can confirm that this fix works. (But why not just use `realpath`?)

jthegedus:
@aca We're trying not to add any further dependencies to the `asdf` core to keep it portable. I'll give this a look soon as it is a small change and should be easily resolvable 👍

craigfurman:
Thanks @jthegedus! I'd almost forgotten about this. I've rebased to fix the conflicts. I'm still not quite sure why the tests are failing, and this might be my ignorance of asdf-core, but I'm also not sure why we don't just trust the filesystem and not care whether a file is a symlink or not (https://github.com/asdf-vm/asdf/pull/625#issuecomment-596193433).

jthegedus:
>... I'm also not sure why we don't just trust the filesystem and not care whether a file is a symlink or not

I am ignorant of this myself.

Stratus3D:
From #366:

> @Stratus3D I opened #625 to fix this, and this appears to work in general, but I'm missing context on why asdf needs to resolve symlinks at all?

It's a little complicated, we use `sed -i` to change versions in `.tool-versions` files in place. `sed -i` doesn't handle symlinks correctly and overwrites the file instead of following the symlink like it should. Some users are using symlinked`.tool-versions` files and our code broke them. We get around this issue by resolving the symlink ourselves mnually and editing to the source file instead of the symlink.

Stratus3D:
> @Stratus3D I'm not sure if you've seen this. I'm currently dependent on my fork in order for `asdf global` to resolve relative symlinks correctly.
> 
> Now that I look at this again, I'm not sure why asdf needs to resolve symlinks at all? We could trust the filesystem, unless there's a specific reason not to.

Sorry for the slow progress on this PR. I don't understand what is wrong with our current symlink resolution code, and I don't understand what your code fixes. Can you provide a concrete test/example of a case where our current code doesn't behave as you expect (ideally a new BATS test, but even a sequence of shell commands and there output would be fine). I see you altered one of our existing tests. Is the test incorrect?

craigfurman:
> sed -i doesn't handle symlinks correctly and overwrites the file instead of following the symlink like it should

Ah, that makes sense, thanks. I'm also using symlinked .tool-versions, hence this PR.

@Stratus3D it's been a while since I looked at this, and I see that I didn't do a great job explaining exactly what it fixes! Basically, the current asdf code (at least as of last December) only works when the shell's working directory is at the same level as the relative symlink's starting point.

e.g.

```
# The symlink in ~ is relative
$ readlink ~/.tool-versions
workspace/ansible-home/config/home/.tool-versions

# doesn't work when working dir is not ~
$ asdf global golang 1.15.2                                                                                                                                                                                                             
/Users/craig/.asdf/lib/commands/version_commands.bash: line 44: /Users/craig/.asdf/workspace/ansible-home/config/home/.tool-versions: No such file or directory

# works when working dir is ~
$ asdf global golang 1.15.2                                                                                                                                                                                                               
```

My use case is that I use GNU stow to create ~/.tool-versions as a symlink to a config repo.

I've changed the test to not be a special case in which the working dir is the same level as the symlink. Does this make sense?

aca:
Same here. I also use stow. If I use set .tool-versions with stow, I can't use asdf global command.

alecho:
I'm seeing the same thing also using gnu stow, macOS, and asdf 8.0.0 (and 0.7.8). I'd be for just trusting the filesystem.

```bash
asdf global elixir 1.11.0-otp-23
/Users/username/.asdf/lib/commands/version_commands.bash: line 47: /Users/username/projects/project/dot-files/asdf/.tool-versions: No such file or directory
```

`No such file or directory` is correct as the file exists at `/Users/username/projects/dot-files/asdf/.tool-versions`

```bash
ll ~/.tool-versions
lrwxr-xr-x  1 username  staff  29 Aug  7 23:53 /Users/username/.tool-versions -> dot-files/asdf/.tool-versions
```

Stratus3D:
Replaced by #815

Issue #530, opened 2019-05-29
Enhancement: asdf plugin-update --all could be parallel


mattysweeps:
Each plugin in `asdf plugin-update --all` could be updated in parallel, making the command much quicker.

Each asdf plugin is a git repository. The same is true for [Plug](https://github.com/junegunn/vim-plug): the vim plugin manager. See [how they update their git repos in parallel](https://raw.githubusercontent.com/junegunn/i/master/vim-plug/40-in-4.gif) which makes for a nice user experience.

Stratus3D:
This seems like a good idea. Only issue we may run into while doing this is handling the output in a sane way. Right now we just send the Git output to STDOUT, but if we've got multiple updates running concurrently we'd get jumbled output. We'll need to figure out how to keep the output sequential while making the updates concurrent. I think it should be easy enough, just capture the output from each update in a variable, once updates are finished print the output from each plugin to STDOUT in order.

PR #626, opened 2019-12-30
Enables parallelization of updating all plugins

deiga:
# Summary

Changes plugin update to use separate function

fixes: #530


lib/commands/command-plugin-update
```diff
@@ -10,16 +10,30 @@ plugin_update_command() {
...
+}
+
+update_plugin() {
+  local plugin_name=$1
+  local plugin_path=$2
+  local gitref=$3
+  logfile=$(mktemp)
+  {
     echo "Updating $plugin_name..."
     (cd "$plugin_path" && git fetch -p -u origin "$gitref:$gitref" && git checkout -f "$gitref")
```
joschi: Would it make sense to pass `-q`/`--quiet` to `git fetch` and `git checkout` to avoid interleaved progress messages from the different commands?

deiga:
@Stratus3D Could someone take a look at this PR?

jthegedus:
@deiga I've not written Bats tests for concurrency before, is it possible to test that they do indeed run concurrently successfully? (no is a fine answer)

Additionally, can you share if `wait` and `&` are shell specific or are they POSIX commands or just OS specific?

Stratus3D:
Changes look good to me! @jthegedus I'm actually in the process of writing tests for the plugin update functionality, since we don't have tests and I recently merged a PR that fixed a plugin update bug.

Stratus3D (approved):

Stratus3D:
Rebased this PR onto latest master.

Stratus3D:
Thanks for the PR @deiga ! Merging now.

PR #627, opened 2019-12-30
Fixes tests in master

deiga:
# Summary

Git tags were not being fetched

### DIFF
--- a/test/update_command.bats
+++ b/test/update_command.bats
@@ -38,8 +38,8 @@ teardown() {
   run asdf update
   [ "$status" -eq 0 ]
   cd $ASDF_DIR
-  echo $(git tag) | grep $tag
-  [ "$status" -eq 0 ]
+  git tag | grep $tag
+  [ "$?" -eq 0 ]
 }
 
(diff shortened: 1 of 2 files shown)


test/update_command.bats
```diff
@@ -34,23 +34,23 @@ teardown() {
 }
 
 @test "asdf update should checkout the latest non-RC tag" {
-  local tag=$(git tag | grep -vi "rc" | head -1)
+  local tag=$(git tag | grep -vi "rc" | tail -1)
```
Stratus3D: Why the switch from head to tail? This shouldn't matter.

Stratus3D:
Thanks for the PR @deiga ! I'm not sure the switch to `tail` was necessary, but either way this fixes the tests.

test/update_command.bats
deiga: @Stratus3D It doesn't really matter. I chose to do this because the test specifies `latest tag`  but `head` returns the first tag. 
I just wanted to ensure the test description matched the contents

PR #630, opened 2019-12-31
Defaults to empty ASDF_DATA_DIR

lalloni:
If ASDF_DATA_DIR variable is undefined it will assign an empty value to
it so bash instances running with set -o nounset can run the script
without error.


lalloni:
Any chance of this being merged?

Stratus3D:
Thanks @lalloni !

PR #637, opened 2020-01-15
Prefer open source distros over closed source distros in build

Stratus3D:
Also checkout specific tag of bats-core instead of using master.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -41,7 +41,7 @@ jobs:
 
       - name: Install bats
         run: |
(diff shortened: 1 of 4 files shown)


vic:
I believe it would be good to test on both, macOS and ubuntu linux since that would pretty much validate that all our code is compatible on both (and that we are ok with macOS having bsd version of tools).

Stratus3D:
@vic we run tests on both, and I'm not changing that. Are you saying we should run the linter script on both OSX and Ubuntu?

vic:
Oh, my bad. Test should definitely run on both, but the lint and format checks only in one of them, Ubuntu is fine. 👍

PR #640, opened 2020-01-25
Fix asdf shell xx --unset for fish

danhper:
`unset` does not exists on fish so `asdf shell xx --unset` was failing. This fixes the issue.


Stratus3D:
Looked good to me!

PR #643, opened 2020-01-28
Fix bash completion for `plugin-add`

mikeroll:
# Summary

`$ asdf plugin-add <TAB>` currently yields this mess:

image

Fixing it to provide a clean list of available but not installed plugins (this is also what fish completion already does and somehow got overlooked in bash version):

image


Stratus3D:
Thanks for the PR!

Issue #634, opened 2020-01-06
Errors when running `bats test` locally


Stratus3D:
### Steps to reproduce

On OSX run:

`bats test`

### Expected behavior

Tests should pass as they were before.

### Actual behavior

Many failing tests are encountered. And some tests hang indefinitely. An example error I captured from a single test:

```
/Users/trevorbrown/.asdf/bin/asdf: line 82: /var/folders/7m/9bq62tmj60zf14d788w0lk8c0000gp/T/asdf.XXXX.ed4MwuLg/home/.asdf/lib/commands/command-help: No such file or directory
```

### Environment

**OS**: 10.15.1
**asdf version**: v0.7.6-bde7ae5

Stratus3D:
@vic do you know what is going on? I've been running `bats test` for years and I'm not sure why I'm seeing so many failures now.

Stratus3D:
The `asdf update is a noop for non-git repos` test hangs indefinitely.

klane:
I am having the same issue, which makes debugging #633 challenging. Manually running `asdf` with my changes seems to work as expected.

Stratus3D:
On tag `v0.7.5` I only get one failure:

```
 ✗ get_config returns default when config file does not exist
   (in test file test/get_asdf_config_value.bats, line 29)
     `[ "$result" = "no" ]' failed
```

On tag `v0.7.6` I get 65 failures before it reaches the `asdf update is a noop for non-git repos` test and it hangs indefinitely.

Stratus3D:
I get the exact same failures and hanging test on my Ubuntu laptop as well.

vic:
Try using https://github.com/bats-core/bats-core (just like our CI is doing) and not https://github.com/sstephenson/bats, the later has not been updated in years.

Stratus3D:
@vic I am using bats-core version 1.1.0. I don't think the bats version is the issue.

Stratus3D:
Do test pass for you @vic ? What OS are you on?

vic:
Yep, all green here. 212 test, 0 failures.

Also green in CI macos catalina / ubuntu.
https://github.com/asdf-vm/asdf/runs/375715344#step:7:219

I'm using Mac OS Catalina.
bats-core 1.2.0-dev (commit 87b16eb8ec381beca7f05d028e4ad7cf3f647a0f)

Stratus3D:
[...]

PR #649, opened 2020-02-04
Misc. Test Improvements

Stratus3D:
# Summary

Unset `ASDF_DIR` before running tests to ensure tests always pass regardless of whether the user already has `ASDF_DIR` set in their shell.

Fixes: #634


Stratus3D:
@jthegedus yes. Appveyor has never passed and is for running the tests on windows. I'm going to try to get around to getting the failing tests working on windows soon.

Issue #210, opened 2017-07-26
Automatically fetch plugins


danhper:
@vic 
> Just wondering, once this gets merged, would it be possible to tell people, to just install asdf and afterwards go directly with asdf list-all elixir or even to asdf install elixir 1.4.5 without having to be explicit about plugin-add elixir ? I mean, once we have the plugins repo, plugin-add is looks to me like only really useful for non-official plugins.

This would probably work as follow:
Check if the plugin is available locally, if not, check if the plugin is in the registry. If the plugin is available in the registry, fetch it before running the command.

danhper:
I will close this issue, as it has been inactive for a long time now.
If there is still interest, please feel free to open a new issue.
Thank you!

jthegedus:
Related https://github.com/asdf-vm/asdf/pull/313

PR #651, opened 2020-02-05
Automatically fetch plug-ins

nomasprime:
# Summary

Found @edwardgalligan's patch but no PR.

Fixes: #210

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -73,6 +73,7 @@ list_installed_versions() {
 
 check_if_plugin_exists() {
   local plugin_name=$1
+  local autoinstall_plugin=$2
 
   # Check if we have a non-empty argument
   if [ -z "${1}" ]; then
@@ -81,8 +82,12 @@ check_if_plugin_exists() {
   fi
 
   if [ ! -d "$(asdf_data_dir)/plugins/$plugin_name" ]; then
-    display_error "No such plugin: $plugin_name"
-    exit 1
+    if [ "$autoinstall_plugin" == "1" ]; then
+        plugin_add_command "$plugin_name"
+    else
+        display_error "No such plugin: $plugin_name"
+        exit 1
+    fi
   fi
 }
 


edwardgalligan:
@nomasprime thanks for creating this. I can't recall exactly why I never PRed—it might be that I intended to do further testing (seems `shfmt` has some objections).

Happy to have it PRed/merged, if it works.

nomasprime:
Thank you for creating it. You did the work 👏 

Just thought I'd get up here so can be completed and merged or someone can adopt it.

vic:
How is this functionality enabled ? On the PR I see no mention of it on the documentation nor tests for enabling it from the command line. Who is passing that `autoinstall_plugin` argument to  the `check_if_plugin_exists` function?

Stratus3D:
Thanks for the PR @nomasprime. I'm not sure I understand the changes in this PR, but I don't think this feature is a good idea. I can see how automatically install plugins from third party repos could be a security risk. As it stands now users must choose to install a plugin before third party code is downloaded and run on their machine.

I personally use the old fashioned `asdf plugin-add <plugin name> <plugin url>` syntax to be explicit about what code I am download (https://github.com/Stratus3D/dotfiles/blob/master/scripts/setup.sh#L125). I do not want tools automatically download and run third party code on my machine without my manual approval.

jthegedus:
@Stratus3D Would a solution that lists all discovered plugins with URLs (`<plugin name> <plugin url>`) and then asks for confirmation from the user be acceptable?

edwardgalligan:
> *I do not want tools automatically download and run third party code on my machine without my manual approval.*

While I understand your concern, this is dependency management. Downloading and running third party code on your machine is what package managers do.

How those package managers expose your options (user prompts during discovery, etc.) is a good topic, but avoiding tools that automatically download dependencies and run them on your machine is avoiding every package manager with dependency management as a feature.

Stratus3D:
> @Stratus3D Would a solution that lists all discovered plugins with URLs (`<plugin name> <plugin url>`) and then asks for confirmation from the user be acceptable?

Yes. That seems like a good idea. That way user is made aware of the plugins being installed.

> Re security: I have been thinking we should [draft a security policy](https://github.com/asdf-vm/asdf/security/policy)

I've created a PR for this - https://github.com/asdf-vm/asdf/pull/660

Stratus3D:
> While I understand your concern, this is dependency management. Downloading and running third party code on your machine is what package managers do.

Yes, but there is a difference doing things automatically without the users permission and keeping the user in the loop and asking them for confirmation when installing new plugins. My only objection is to systems that automatically pull in third party code without informing the user or giving them an option.

And while yes, this is part of what package managers do, asdf is unique in that plugins and versions have a direct impact on the user's shell when asdf is installed. Whereas a node package likely will not affect the commands available in the user's shell, and usually won't override commands provided by the operating system (although that happens sometimes). So I feel like at the operating system level, the risk is a little greater with asdf because it's designed to place shims in front of other executables on the path.

SidOfc:
Don't want to add noise to this convo, but I like this idea as well :)

If a plugin isn't installed wouldn't it be possible to check if `$plugin_name` exists within `asdf plugin list all` and if so, it could perhaps prompt the user with a question along the lines of:

```plain
Plugin [NAME] isn't installed, would you like to install from [URL] (y/N)?
```

Both plugin name and URL exist in the output of `asdf plugin list all`, only tricky bit would be writing a confirm prompt. I do actually have some example code which could be used for confirmation:

```shell
confirm() {
  local prompt="$1 (y/N): "
  while IFS=; read -s -r -n 1 -p "$prompt" choice; do
    prompt=""
    case "$choice" in
      [yY]* ) echo 'yes'; break;;
      "" | [nN]* ) echo 'no'; break;;
    esac
  done
}
```

I played around with some shell commands to figure out how to extract URL from `asdf plugin list all`, basic command I came up with:

```shell
asdf plugin list all | grep ruby | rev | cut -d' ' -f1 | rev
```

This does rely on `grep`, `rev`, and `cut` being installed as well as the URL being the last "column" in the output of `asdf plugin list all` though, not sure what your opinions about that are.

Just putting this out there in case anyone wants to have a go at it, on the other hand it's not too much effort to add the plugin once, but this would make asdf ever so slightly more elegant than it already is :)

jthegedus:
@SidOfc Thanks for your input, it is an interesting idea.

When thinking about the solution here and in https://github.com/asdf-vm/asdf/pull/313 they do almost the same thing except one requires a `.tool-versions` and the other not. Perhaps once we get that in we can address this with little change.

As mentioned in #313, I think the biggest concern with this solution is the dependency on the asdf-plugin repo and the fact that the plugin repo only has one repo per plugin name. I have been toying with asdf-plugin repo ideas that would break this 1:1 model and subsequently break these types of features.

alanbccoe:
I've just found asdf and this is a feature I was looking for in the docs. Running in an enterprise with 200+ git repos we'd clone the plugins we wanted to use for security. It'd be convenient if in each .tool-versions file the git path of the plugin to use could be specified (or in some other file). E.g.:

```
jq 1.2.3 https://github.com/myorg/asdf-jq
... 
```

This'd mean devs could just clone an internal repo they'd never used before and just use it instead of messing around manually installing plugins. 

Also, having a way to define that plugins whose GH pages match a regex (e.g. github.com/myorg/*) are trusted and would automatically be installed (with the ability to suppress prompting to aid CI) would be helpful.

prem-prakash:
@alanbccoe your idea seems very promising

Stratus3D:
@nomasprime thank you for the PR. I am sorry to have left this open for so long. Based on the discussion on this thread and on https://github.com/asdf-vm/asdf/pull/313 I have decided to close this PR as we do not feel comfortable adding this feature.

Issue #559, opened 2019-08-09
asdf global elixir path:<PATH> doesn't work


wojtekmach:
### Steps to reproduce

```
$ git clone git@github.com:elixir-lang/elixir.git
$ asdf global elixir path:$PWD/elixir
sed: 1: "s/^elixir .*$/elixir pa ...": bad flag in substitute command: 'U'
rm: /Users/wojtek/src/dotfiles/asdf/.tool-versions.bak: No such file or directory
```

Furthermore, If I manually add to `~/.tool-versions`:

```
elixir path:/Users/wojtek/elixir
```

then I get:

```
$ iex
asdf: No version set for command iex
you might want to add one of the following in your .tool-versions file:

elixir 1.9.0-rc.0-otp-22
elixir 1.9.0-otp-22
```

### Expected behavior

Setting and using path option should work.

### Environment

**OS**: macOS

**asdf version**: v0.7.2

Stratus3D:
Thanks for reporting this. Definitely a bug.

Stratus3D:
I'll have a fix for this out shortly.

Stratus3D:
I've merged the fix. I'll be tagging a new release of asdf shortly.

wojtekmach:
Thank you! I tried double checking it and got this:

```
cd ~/.asdf
git checkout master
git pull origin master
asdf global elixir path:/Users/wojtek/src/elixir
```

but still see:

```
iex --version
asdf: No version set for command iex
you might want to add one of the following in your .tool-versions file:

elixir 1.9.0-rc.0-otp-22
elixir 1.9.0-otp-22
elixir git
```

let me know if you want me to test anything, maybe I'm missing something how to properly use master.

Stratus3D:
Maybe try `asdf reshim`? It's possible there is more to this issue.

wojtekmach:
Yeah, already tried that, no luck.

Stratus3D:
Yes there is another problem here. Not sure what is going on but it's not working for me either.

zachdaniel:
@Stratus3D any motion on this?

Stratus3D:
@zachdaniel I started digging into this again this morning. Hoping to have a PR out to fix this sometime this afternoon!

Stratus3D:
Fixed!

rm-rf-etc:
What should we do if this is still happening?

rm-rf-etc:
Solution for me was to manually create `~/.tool-versions` and add the content
[...]

PR #654, opened 2020-02-08
Fix Path Versions

Stratus3D:
# Summary

Add back in support for path versions, and write additional unit tests for path versions.

Fixes: #559

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -60,6 +60,8 @@ get_install_path() {
 
   if [ "$install_type" = "version" ]; then
     echo "${install_dir}/${plugin}/${version}"
+  elif [ "$install_type" = "path" ]; then
+    echo $version
   else
     echo "${install_dir}/${plugin}/${install_type}-${version}"
   fi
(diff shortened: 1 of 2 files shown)


lib/utils.sh
```diff
@@ -617,58 +622,66 @@ asdf_run_hook() {
   fi
 }
 
-with_shim_executable() {
-  get_shim_versions() {
-    shim_plugin_versions "${shim_name}"
-    shim_plugin_versions "${shim_name}" | cut -d' ' -f 1 | awk '{print$1" system"}'
-  }
```
Stratus3D: Move these functions outside of the `with_shim_executable` function to make things simpler to read.

lib/utils.sh
```diff
@@ -466,6 +468,9 @@ list_plugin_exec_paths() {
   if [ "${version_info[0]}" = "ref" ]; then
     local install_type="${version_info[0]}"
     local version="${version_info[1]}"
+  elif [ "${version_info[0]}" = "path" ]; then
+    local install_type="${version_info[0]}"
+    local version="${version_info[1]}"
```
Stratus3D: Add back in support for path versions.

PR #656, opened 2020-02-14
Default ZSH_VERSION to empty string.

### LINKED ISSUE #655: Shim-load script fails with `set -o nounset`
### Steps to reproduce

```bash
#!/bin/bash
set -u
. $(brew --prefix asdf)/asdf.sh
```

When using Bash with the `nounset` setting, a.k.a. `set -u`, sourcing the `asdf.sh` activation script fails because `ZSH_VERSION` isn't set. Of course, if this variable isn't set, we aren't running Zsh, so we should probably default it to the empty string.

### Expected behavior

No error.

### Actual behavior

Script fails with `ZSH_VERSION: unbound variable`.

### Environment

**OS**: macOS, Bash

**asdf version**: `master` `HEAD`

toothbrush:
# Summary

This avoids borking when a user has `set -u` (a.k.a. the `nounset` option) enabled.

Fixes #655.

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -32,7 +32,7 @@ PATH="${ASDF_USER_SHIMS}:$PATH"
 # Load the asdf wrapper function
 . "${ASDF_DIR}/lib/asdf.sh"
 
-if [ -n "$ZSH_VERSION" ]; then
+if [ -n "${ZSH_VERSION:-}" ]; then
   autoload -U bashcompinit
   bashcompinit
 fi


jthegedus:
I am just noticing that we have some conflicting usage/guides for ZSH users.

#69 and #270 offer differing information about handling compinit. @toothbrush are you able to unify these?

asdf.sh
```diff
@@ -32,7 +32,7 @@ PATH="${ASDF_USER_SHIMS}:$PATH"
 # Load the asdf wrapper function
 . "${ASDF_DIR}/lib/asdf.sh"
 
-if [ -n "$ZSH_VERSION" ]; then
+if [ -n "${ZSH_VERSION-}" ]; then
```
jthegedus: Can we use `:-` here for consistency with the above code

toothbrush: No problem, done.

PR #660, opened 2020-02-25
Create SECURITY.md file

Stratus3D:
# Summary

Create SECURITY.md file with list of supported versions and secure contact information for reporting vulnerabilities.

Relates to: #651

## Other Information

@asdf-vm/core - if you have secure method of being contacted and want to handle vulnerability reporting please add your username and contact address to the list. It doesn't have to be email. I still do not have PGP setup for my own email address yet.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,9 +1,11 @@
 # Security Policy
 
+**This security policy only applies to asdf core, which is the code contained in this repository. All plugins are the responsibility of their creators and are not covered under this security policy.** 
+
 ## Supported Versions
 
 Typically only the latest version is maintained. Versions are
-released sequencially, with major, minor, and patch versions incremented
+released sequentially, with major, minor, and patch versions incremented
 as necessary. If a vulnerability is found in an old version of asdf,
 updating to the latest version of asdf will be required.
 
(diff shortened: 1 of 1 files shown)


danhper:
Added my info

Stratus3D:
Thanks @danhper ! can you recommend any good resources on setting up PGP for email? I use GPG and pass for password management, but I've not yet done anything with PGP.

SECURITY.md
```diff
@@ -0,0 +1,27 @@
+# Security Policy
+
+## Supported Versions
+
+Typically only the latest version is maintained. Versions are
+released sequencially, with major, minor, and patch versions incremented
```
jthegedus: sequentially

jthegedus:
Should we specify a minimum time for us to respond and address the issue before the reporter should make a public disclosure?

Stratus3D:
>    1. Should we specify a minimum time for us to respond and address the issue before the reporter should make a public disclosure?

I always try to respond to important issues as fast as possible, but my response times vary. I'm not willing to commit to addressing security issues within a set time frame. I will give them priority, but I'm not willing to commit to anything more.

>    2. I think we should explicitly call out that this only applies to the asdf-vm/asdf core tool and not any plugins whose authors are responsible for their own security policies and guarantees. Thoughts?

Yes, totally agree. I will update the PR.

Stratus3D:
I'm using my own email server and Thunderbird is my email client, it shouldn't be too hard for me to setup, I just have to take the time to learn how to do it.

Issue #607, opened 2019-12-01
`asdf help` does not work with Homebrew installation


deiga:
### Steps to reproduce

- `brew install asdf`
- `asdf` or `asdf help`

### Expected behavior

Should list help text for `asdf`

### Actual behavior

```
cat: /usr/local/VERSION: No such file or directory
version:

cat: /usr/local/help.txt: No such file or directory

"Late but latest"
-- Rajinikanth
```

### Environment

**OS**: macOS 10.14.6

**asdf version**: 0.75

vic:
I'm not a Homebrew user, it'd be awesome if you can find what's the problem, looks like the `asdf_dir` function thinks it's somehow at `/usr/local/lib/utils.sh` which would be weird.

deiga:
@vic The issue has probably to do with how Homebrew links files. So the brew installation links everything to `/usr/local`, like this:
```
$ brew link asdf --dry-run
Would link:
/usr/local/etc/bash_completion.d/asdf.bash
/usr/local/bin/asdf
/usr/local/share/fish/vendor_completions.d/asdf.fish
/usr/local/lib/asdf.fish
/usr/local/lib/asdf.sh
/usr/local/lib/commands
/usr/local/lib/utils.sh
```

When running `/usr/local/lib/commands/command-help` it tries to read `help.txt` from `asdf_dir()` which does not resolve to the brew install path but the symlink path.
One option would be to make an conditional to return `$(brew --prefix asdf)` from `asdf_dir()` if `asdf` was installed from Homebrew.

vic:
Hm, that's weird, so your `/usr/local/lib/utils.sh` is a symlink to the real `$ASDF_DIR/lib/utils.sh` ? I wouldn't expect brew to do that, what if another package had their own `lib/utils.sh` file, that would cause problems. And.. even when you have `lib/utils.sh` symlinked, you don't have a `/usr/local/VERSION` symlink, right?

On our side, I believe we have to make `bin/asdf` determine if it's own file is a link and somehow resolve to the real ASDF dir, probably using something like `realpath` (from coreutils but that would cause asdf to depend on coreutils being available) or `readlink`. If I recall correctly both of `realpath`and `readlink` were being checked somewhere inside test/ as forbidden commands, and there was also a function of our own to resolve links.

vic:
In the mean time you could use the `bin/asdf` from the real location where asdf was installed by homebrew.

vic:
[...]

PR #667, opened 2020-03-06
Ensure we are using installation location instead of symlink location

deiga:
# Summary

Change to resolve actual asdf script location before sourcing anything.
Ensures that we are not using symlink location as basis for sourcing.

Fixes:
- https://github.com/asdf-vm/asdf/issues/607

### DIFF
--- a/test/banned_commands.bats
+++ b/test/banned_commands.bats
@@ -4,8 +4,8 @@ load test_helpers
 
 banned_commands=(
     realpath
-    # readlink on OSX behaves differently from readlink on other Unix systems
-    readlink
+    # readlink -f on OSX behaves differently from readlink -f on other Unix systems
+    "readlink -f"
     # It's best to avoid eval as it makes it easier to accidentally execute
     # arbitrary strings
     eval
--- a/bin/asdf
+++ b/bin/asdf
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+## Courtesy of https://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac
 normalized_readlink() {
   TARGET_FILE=$1
 
--- a/bin/asdf
+++ b/bin/asdf
@@ -1,8 +1,26 @@
 #!/usr/bin/env bash
 
-# Get files actual location
-SCRIPT=$(readlink -f "$0")
+normalized_readlink() {
+  TARGET_FILE=$1
 
+  cd "$(dirname "$TARGET_FILE")" || exit
+  TARGET_FILE=$(basename "$TARGET_FILE")
+
+  # Iterate down a (possible) chain of symlinks
+  while [ -L "$TARGET_FILE" ]; do
+    TARGET_FILE=$(readlink "$TARGET_FILE")
+    cd "$(dirname "$TARGET_FILE")" || exit
+    TARGET_FILE=$(basename "$TARGET_FILE")
+  done
+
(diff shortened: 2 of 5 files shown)


deiga:
I've also created a PR to Homebrew to change the formula, now that symlink resolution works the right way https://github.com/Homebrew/homebrew-core/pull/51200

vic:
Nice @deiga, thanks for fixing this.  👍 

There's also a `resolve_symlink` on `utils.bash`. May we remove it, and replace any `resolve_symlink` usage for `normalized_readlink` and see if everything is ok ?

test/banned_commands.bats
```diff
@@ -4,8 +4,8 @@ load test_helpers
 
 banned_commands=(
     realpath
-    # readlink on OSX behaves differently from readlink on other Unix systems
-    readlink
+    # readlink -f on OSX behaves differently from readlink -f on other Unix systems
+    "readlink -f"
```
Stratus3D: This isn't going to catch all uses of `readlink -f`.

Stratus3D:
It would be great to have a test for this so we can prevent regressions.

test/banned_commands.bats
vic: Agree with @Stratus3D, it's better to leave `readlink` here.
@deiga, could you please revert this change.

deiga: I'm not sure how to do that. The `normalize_readlink` function uses `readlink` and thus wouldn't pass the banned commands test.
Unless we build some exceptions logic to the banned commands test

deiga:
@Stratus3D @vic I actually realized a simpler way to fix this.
Any changes to the Homebrew formula are unnecessary now.

It also turned out, that `resolve_symlink` was not able to actually resolve symlinks properly.

We've still got the problem with `readlink` in banned commands test though.

jthegedus:
@deiga How is the work on this going? Can you rebase to remove the false negative of Appveyor?

deiga:
@jthegedus I'm waiting on further comments.
I rebased in the meantime

Stratus3D:
@deiga @jthegedus re-reviewing this now.

test/banned_commands.bats
hyperupcall: Could something like `readlink.+-.*f.+"` be good enough? And assume that some quoted variable reference follows the readlink invocation?

hyperupcall: @deiga To fix the test, you'll want to add some comment at the end of the line: `# asdf_allow: readlink`

That's probably the best way to go forward to absolutely ensure there are 0 chances of missing a `readlink -f` for the future

Stratus3D: Yeah I think a regex like you have there would work @hyperupcall.

Stratus3D:
I'm going to merge https://github.com/asdf-vm/asdf/pull/1539 and then this PR can be rebased and merged.

Stratus3D:
@deiga can you rebase this PR? I've merged some changes so only `readlink -f` is banned now. I think we can merge your changes after another rebase.

Stratus3D:
Hi @deiga I've been in the process of rewriting asdf in Go. Thank you for your contribution, unfortunately it doesn't make sense to consider these changes now as I'm planning to merge the Go code in the next week or two. Closing this PR now.

The Go version is available here - https://github.com/asdf-vm/asdf-core-go would you want to give it a try and see if it addresses the issue you were fixing here? I may be wrong, but I think I did fix this issue in the Go code.

PR #668, opened 2020-03-07
Default to empty version to avoid error on strict mode bash

lalloni:
In strict mode bash this needs to be changed to avoid throwing an error.


vic:
Nice, thanks @lalloni

PR #676, opened 2020-03-12
Use different exit code if updates are disabled

joschi:
# Summary

If asdf-vm was installed with a package manager and the user doesn't have the necessary permissions to update it with `asdf update`, asdf-vm emits an informational message and exits with exit code 1.

This makes it hard to programmatically detect whether the update failed or wasn't even attempted because it's not possible.

With this change, asdf-vm would exit with the exit code 2 if updates are disabled instead of exit code 1, which signals an error during update.

Refs r-darwish/topgrade#367

### DIFF
--- a/lib/commands/command-update.bash
+++ b/lib/commands/command-update.bash
@@ -8,7 +8,7 @@ update_command() {
 
     if [ -f asdf_updates_disabled ] || ! git rev-parse --is-inside-work-tree &>/dev/null; then
       echo "Update command disabled. Please use the package manager that you used to install asdf to upgrade asdf."
-      exit 1
+      exit 42
     else
       do_update "$update_to_head"
     fi
--- a/test/update_command.bats
+++ b/test/update_command.bats
@@ -53,11 +53,24 @@ teardown() {
   [ "$?" -eq 0 ]
 }
 
+@test "asdf update is a noop for when updates are disabled" {
+  touch $ASDF_DIR/asdf_updates_disabled
+  run asdf update
+  [ "$status" -eq 42 ]
+  [ "$(echo -e "Update command disabled. Please use the package manager that you used to install asdf to upgrade asdf.")" == "$output" ]
+}
+
 @test "asdf update is a noop for non-git repos" {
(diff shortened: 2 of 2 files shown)


joschi:
The reason for the failing build on AppVeyor seems unrelated to this PR.

Stratus3D:
Thanks for the PR @joschi ! I like these changes. And yes, the AppVeyor builts are going to be for Windows support, which isn't yet working.

I see you updated the tests, but I think it might make more sense to have separate tests for each possible exit code (1 and 2). Would it be difficult to add a new unit test for the exit code of 1?

joschi:
@Stratus3D Thanks for your quick response!

I will think of a way to test for the different scenarios when failing to run `asdf update`.

In the linked issue at `topgrade`, we also discussed introducing a kind of dry-run flag for `asdf update`:
https://github.com/r-darwish/topgrade/issues/367#issuecomment-598192568

Does that sound useful for you or would you prefer signaling the inability to perform an update with the exit code of the `asdf update` command exclusively?

Stratus3D:
@joschi good question. I guess my first thought would be to keep things as simple as possible. We want users to be able to get the info they need, but we also want our API (the `asdf` commands) to be as simple as possible. I like exit codes because of their simplicity. They are easy to implement and are usually enough to give the user all the information they need.

I'm not sure there is a clear benefit to a dry-run flag. Updating is as simple as pulling down a newer tag with Git or using some operating system-specific package manager to update asdf.

Is there a clear use case for a dry-run flag that could not be handled by the `asdf update` as you have it here (with two exit codes)? . And if so can you explain what that scenario would be?

joschi:
@Stratus3D I think you're right and we should keep it as simple as possible.

I've changed the exit code for disabled updates to 42 in order not to conflict with other conventions (http://www.tldp.org/LDP/abs/html/exitcodes.html) and updated the tests.

Stratus3D:
Thanks @joschi ! The code looks good. Thanks for taking the time to write the unit tests.

Issue #674, opened 2020-03-11
Completion changes in v0.77 causing slow startup and invalid zcompdump file location


andrewthauer:
### Overview

Changes in `v0.77` to `asdf.sh` now call `autoload -Uz compinit` & `compinit` automatically when using `zsh`. This has the following issues:

- The `compinit` command should only be called once per shell startup in `zsh`
- `compinit` is often called with different arguments. Sometimes this is to use a cached version or to relocation the `.zcompdump` file. Making assumptions on how it's called undermines any customization.
- This can drastically slow down the shell startup time as it could already be called somewhere else by a framework or customized setup. On my machine this adds on average an additional 1 second delay.

See https://github.com/asdf-vm/asdf/blob/master/asdf.sh#L37

### Steps to reproduce

- Make sure your using `zsh`
- Run `/usr/bin/time zsh -i -l -c exit` to get a baseline
- Comment out lines 37 & 38 from `asdf.sh`
- Run `/usr/bin/time zsh -i -l -c exit` again and compare

NOTE: This could vary depending on a bunch of factors.

### Expected behavior

There should be no noticeable difference in startup time. It also shouldn't affect the behaviour of any custom zsh completion initialization.

Ideally only the `fpath` is set and `asdf` assumes zsh completions will be initialized later, either by a zsh framework or manually in the `.zshrc` (or equivalent). Therefore documenting this instead of automatically calling it will be less intrusive.

### Actual behavior

Shell startup time is significantly slower then in v0.76. It can also affect the behaviour of the zsh completion initialization in unwanted ways. For instance, creating a `.zcompdump` file when the location has been moved from the default location using `compinit -i -d "$_zcompdump"`.

### Environment

macOS

**asdf version**:

0.77

jthegedus:
Hi @andrewthauer , thanks for the thorough explanation.

Looking at the diff between https://github.com/asdf-vm/asdf/compare/v0.7.6...v0.7.7 the only changes to this file were

```diff
- if [ -n "$ZSH_VERSION" ]; then
-   autoload -U bashcompinit	
-   bashcompinit
+ if [ -n "${ZSH_VERSION:-}" ]; then
+   fpath=(${ASDF_DIR}/completions $fpath)
+   autoload -Uz compinit
+   compinit
+ fi
```

the bulk of which came from https://github.com/asdf-vm/asdf/pull/544

[...]

PR #678, opened 2020-03-13
fix: remove automatic compinit

jthegedus:
# Summary

#544 released in https://github.com/asdf-vm/asdf/releases/tag/v0.7.7 changed:

```diff
- if [ -n "$ZSH_VERSION" ]; then
-   autoload -U bashcompinit	
-   bashcompinit
+ if [ -n "${ZSH_VERSION:-}" ]; then
+   fpath=(${ASDF_DIR}/completions $fpath)
+   autoload -Uz compinit
+   compinit
+ fi
```

which results in `compinit` being executed multiple times (should people use a framework or explicitly use compinit in their own `zshrc``). `compinit` is designed to be called once, so we should not automatically run it here and double up on those already calling it.
 
This has performance implications and concerns were raised about asdf being **unexpectedly** intrusive.

Clearer documentation updates will rectify any issues with this removal. I would wager few ZSH users are relying on asdf to run `compinit` so hopefully we see minimal affect.

This was previously not a problem for ZSH users as we were:

- using `bashcompinit` and thus not interfering with `compinit`
- advising users to source the Bash completions in their `zshrc` files

Fixes: #674

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -31,7 +31,3 @@ PATH="${ASDF_USER_SHIMS}:$PATH"
 # shellcheck source=lib/asdf.sh
 # Load the asdf wrapper function
 . "${ASDF_DIR}/lib/asdf.sh"
-
-if [ -n "${ZSH_VERSION:-}" ]; then
-  fpath=(${ASDF_DIR}/completions $fpath)
-fi
--- a/asdf.sh
+++ b/asdf.sh
@@ -34,6 +34,4 @@ PATH="${ASDF_USER_SHIMS}:$PATH"
 
 if [ -n "${ZSH_VERSION:-}" ]; then
   fpath=(${ASDF_DIR}/completions $fpath)
-  autoload -Uz compinit
-  compinit
 fi
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -83,7 +83,7 @@ echo -e '\n. $HOME/.asdf/asdf.sh' >> ~/.zshrc
 ?>If you are using a framework, such as [oh-my-zsh](https://ohmyz.sh/), these lines remain _below_ the line
 where you source your framework.)
 
-!>If you are not using a framework, or if on starting your shell you get an error message like `command not found: compinit`, then add the below line before the ones above.
+!>If you are not using a framework or are using a custom `compinit` setup, ensure `compinit` is below your sourcing of `asdf.sh` as it adds completions to the `fpath`.
 
 ```shell
 autoload -Uz compinit && compinit
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -86,7 +86,7 @@ or use a framework plugin like [asdf for oh-my-zsh](https://github.com/ohmyzsh/o
 
 ZSH Completions:
 
-asdf ships with ZSH completions, however they need to be setup. Add the following to your `.zshrc`:
+asdf ships with ZSH completions which will need to be setup if you're not using a ZSH framework plugin that does this for you. Add the following to your `.zshrc`:
 
 ```shell
 # append completions to fpath
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -80,14 +80,13 @@ Installation via **Git**:
 echo -e '\n. $HOME/.asdf/asdf.sh' >> ~/.zshrc
 ```
 
-?>If you are using a framework, such as [oh-my-zsh](https://ohmyz.sh/), these lines remain _below_ the line
-where you source your framework.)
+?>Sourcing `asdf.sh` should occur after you source any ZSH framework such as [oh-my-zsh](https://ohmyz.sh/).
 
-!>If you are not using a framework or are using a custom `compinit` setup, ensure `compinit` is below your sourcing of `asdf.sh` as it adds completions to the `fpath`.
+Completions are automatically added to your `fpath` by `asdf.sh`, though `compinit` needs to be run.
 
-```shell
-autoload -Uz compinit && compinit
-```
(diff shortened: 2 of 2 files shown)


jthegedus:
Update to docs still pending hence the WIP badge

jthegedus:
@andrewthauer @Strayer Could you test and review to ensure these changes do as you expect and resolve any performance issues you may have seen? Thanks

Strayer:
Works great for me, I now only source `asdf.sh` in my `.zshrc` and let zinit handle the rest. Loads MUCH quicker and completions work as expected. Thank you @jthegedus! Any way to donate you a beer/coffee/tea/whateveryoulove?

andrewthauer:
I should point out that I noticed the following error could still occur if a user is sourcing the bash completions manually and using zsh (instead of bash):

```
asdf/completions/asdf.bash:68: command not found: complete
```

I have a script that wraps the `asdf` initialization that is intended to work with both `zsh` and `bash`. I was previously always sourcing the completions, but sine v0.77, I started getting this error as well. This was easily worked around by conditionally checking if the session is bash before sourcing. However, this leads me to think this could have started for some others as well.

Strayer:
> However, this leads me to think this could have started for some others as well.

That was exactly what I was doing and got me to even look up the issue here. I think it is reasonable to assume that sourcing a file that specifically states `bash` may not work in zsh, but maybe a warning in that file when it is run inside zsh might make sense.

andrewthauer:
> I think it is reasonable to assume that sourcing a file that specifically states `bash` may not work in zsh, but maybe a warning in that file when it is run inside zsh might make sense.

Exactly. It didn't affect anything noticeable which I'm probably didn't think twice about it. I did start seeing it in `v0.77` though. I would say my fix is not a workaround and is the proper course of action. Maybe it should be noted to not source the bash completions if your using zsh? Not sure on where this actually stems from.

jthegedus:
>I should point out that I noticed the following error could still occur if a user is sourcing the bash completions manually and using zsh (instead of bash):

Given we were advising people to manually source the `.bash` completions before v0.7.7 I think this will be a common issue. I shall update the docs to point people in the right direction. Thanks for pointing this out.

docs before v0.7.7:

- https://github.com/asdf-vm/asdf/pull/544/files#diff-bb5d293cc2573b1db2aad9d78bca7be6L85
- https://github.com/asdf-vm/asdf/pull/544/files#diff-bb5d293cc2573b1db2aad9d78bca7be6L97

Anyone have feelings on how we should handle deprecating this like https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/asdf/asdf.plugin.zsh as they are now doing needless work?

>Thank you @jthegedus! Any way to donate you a beer/coffee/tea/whateveryoulove?

@Strayer thanks, but I am yet to setup GitHub sponsors yet. Has been on my backlog since the beta 😅

jthegedus:
The more I get into the weeds of this the more I think we should not be adding things to the `fpath` at all 🤔 

We recommend sourcing `asdf.sh` after ZSH frameworks, but if the framework calls `compinit`, then the addition to `fpath` isn't picked up, so users would have to run `compinit` themselves anyway.

So I think the solution here is to
- remove `fpath` additions in `asdf.sh`
- update ZSH framework plugins to add completions to `fpath` to properly integrate with their lifecycle and usage of `compinit`. oh-my-zsh is the only one I am aware of that needs updating.
- update Brew installation to use the native ZSH completions
- update documentation for completions to direct people to either roll their own `fpath` addition and `compinit` OR use a framework plugin

Thoughts?

philpennock:
Documenting required steps and otherwise not doing anything WFM.  Packagers can drop the completion into whatever directory their system sets up for such things (Homebrew/whatever), power-users who bypass such things can read the docs.

andrewthauer:
As a power use myself, I already have the expectation of manually appending things to the `fpath`. I've also had to factor in the lifecycle when modifying the `fpath`. Therefore, this strategy makes perfect sense to me and seems consistent with how many other libraries deal with zsh completions.

For reference:

- homebrew installed packages typically symlink their completions to `[brew-prefix]/share/zsh/site-functions`. If a user or framework adds this to their fpath (as I do), they will get completions for all that comply to that convention. See https://docs.brew.sh/Shell-Completion
- [zsh-completions](https://github.com/zsh-users/zsh-completions) is a prominent project that adds zsh completions for many utilities. Their instructions tell users to add to the `fpath` manually as well.

jthegedus:
Thanks for weighing in @philpennock 

Thanks for the information @andrewthauer

Then [Homebrew ZSH change](https://github.com/Homebrew/homebrew-core/pull/51720) has been merged, so users installing with Homebrew should get the correct ZSH completions through that. I will update the docs to reflect this.

I will try the symlink method used by zsh-completions in the oh-my-zsh asdf plugin and see if we can remove the usage of the bash completions there.

And I will update the docs for the rest of people using other ZSH frameworks about manual fpath setup or looking at the aforementioned plugins to update their own ecosystems.

Thanks for all your input everyone.

@Stratus3D Perhaps once I get this fix in we can organise a patch release so others don't run into these same issues?

juanibiapina:
The call to `compinit` will still be removed, right ?

jthegedus:
Yes, the compinit will also be removed.

Stratus3D:
@jthegedus yes, assuming no breaking changes have been merged since `0.7.7` was tagged. I don't think there were any, but as soon as you merge this I'll check and then tag `0.7.8`.

jthegedus:
I think we are good to go now. Updates to plugins can hopefully be socialised by those who use them once they find this "feature" was reversed.

jthegedus:
@Stratus3D Please cut a new release when you have the time, cheers 🍻

Stratus3D:
`0.7.8` has been tagged. 🎉

PR #680, opened 2020-03-15
Update README.md

imbsky:


jthegedus:
LGTM. @Stratus3D can we also actually disable Appveyor in favour of GitHub Actions Windows workflows?

Stratus3D:
@jthegedus I am fine with anything so long as we are testing on Windows during the build. It seems like we've moved everything over to Github Actions since the time I added Appveyor, so switching it to GitHub Actions makes sense.

Issue #681, opened 2020-03-20
Configuration settings for plugins


joschi:
Is there a way for asdf plugins to access the configuration settings in `$HOME/.asdfrc` except for parsing the file themselves?

It would we great, if plugins were provided with the configuration settings in some way.

One possibility would be via environment variables similar to `ASDF_INSTALL_VERSION`, for example `legacy_version_file` in `.asdfrc` could be mapped to `ASDF_CONFIG_LEGACY_VERSION_FILE`.

joschi:
I found this in the documentation:
> When invoked, if extension commands do not have their executable-bit set, they will be
> sourced as bash scripts, having all of the functions from `$ASDF_DIR/lib/utils.bash` available.
> Also, the `$ASDF_CMD_FILE` resolves to the full path of the file being sourced.
> If the executable bit is set, they are just executed and replace the asdf execution.

This means that plugins can use [`get_asdf_config_value()`](https://github.com/asdf-vm/asdf/blob/3246c26d52f0d0368110ac96f82d60301b8f6bc8/lib/utils.bash#L344). 👍

jthegedus:
@joschi Yes, the documentation for plugins needs much improvement. I am working on an update to the docs atm and will add this to the list of things to make more clear. Thanks for raising this and pointing to the solution 👍

PR #682, opened 2020-03-21
Move sort_versions() into utils.bash

joschi:
Instead of almost every asdf plugin [reinventing the wheel](https://github.com/search?q=org%3Aasdf-community+sort_versions&type=Code), the `sort_versions()` function should be made available through `utils.bash`.

### DIFF
--- a/lib/commands/command-update.bash
+++ b/lib/commands/command-update.bash
@@ -42,10 +42,4 @@ do_update() {
   fi
 }
 
-# stolen from https://github.com/rbenv/ruby-build/pull/631/files#diff-fdcfb8a18714b33b07529b7d02b54f1dR942
-sort_versions() {
-  sed 'h; s/[+-]/./g; s/.p\([[:digit:]]\)/.z\1/; s/$/.z/; G; s/\n/ /' |
-    LC_ALL=C sort -t. -k 1,1 -k 2,2n -k 3,3n -k 4,4n -k 5,5n | awk '{print $2}'
-}
-
 update_command "$@"
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -760,3 +760,9 @@ with_shim_executable() {
 
   return 126
 }
+
+# stolen from https://github.com/rbenv/ruby-build/pull/631/files#diff-fdcfb8a18714b33b07529b7d02b54f1dR942
+sort_versions() {
+  sed 'h; s/[+-]/./g; s/.p\([[:digit:]]\)/.z\1/; s/$/.z/; G; s/\n/ /' |
+    LC_ALL=C sort -t. -k 1,1 -k 2,2n -k 3,3n -k 4,4n -k 5,5n | awk '{print $2}'
+}


jthegedus:
Great observation and idea!

vic:
Also, while doing https://github.com/asdf-vm/asdf-plugin-template/blob/master/template/lib/utils.bash, I was thinking that maybe we should provide a set of functions (API) for plugin authors to use.

Some tools like direnv [do that](https://direnv.net/man/direnv-stdlib.1.html).

This way we could provide things like `sort_versions`, `fetch_github_releases`, `download_file` and common stuff we have in almost every plugin. 

asdf's own `util.bash` has access to many *asdf internals*, and IMHO it should be kept private for asdf's own use only. So, regarding this original issue topic, I believe no plugin code *should* be sourcing our `util.bash`.

Perhaps it'd be better to have some `asdf plugin-api`in the same line as the direnv equivalent mentioned above.

joschi:
> So, regarding this original issue topic, I believe no plugin code should be sourcing our `util.bash`.

@vic I agree that a well-defined API to be used by plugins would be great.

But for the time being the `utils.bash` file is automatically sourced in almost every plugin (#681):

> When invoked, if extension commands do not have their executable-bit set, they will be
> sourced as bash scripts, having all of the functions from `$ASDF_DIR/lib/utils.bash` available.

https://github.com/asdf-vm/asdf/blob/c1253a8dfe7c341ca3fa211b75fd6c1a37dbb107/docs/plugins-create.md#L146-L147

jthegedus:
I do like the idea of a plugin-api in the core of asdf, though it would make version mismatches between asdf and the plugins more of a problem as the plugins are then no-longer standalone scripts executed by asdf

jthegedus:
@joschi @vic I was trying to implement a quick fix for https://github.com/asdf-vm/asdf/issues/698 and have found the `sort_versions` function doesn't handle sorting values above `9`. I am unsure if this is known or not, but running the following in your shell should demonstrate the issue:

```shell
asdf list-all nodejs "[0-9]" |
  grep  -vE "(^Available versions:|-src|-dev|-latest|-stm|[-\\.]rc|-alpha|-beta|[-\\.]pre|-next|(a|b|c)[0-9]+|snapshot|master)" |
  sed 's/^\s\+//' |
  sed 'h; s/[+-]/./g; s/.p\([[:digit:]]\)/.z\1/; s/$/.z/; G; s/\n/ /' |
  LC_ALL=C sort -t. -k 1,1 -k 2,2n -k 3,3n -k 4,4n -k 5,5n |
  awk '{print $2}'
```

^^ This is just the `asdf` `latest_command` code with the `sort --version-sort` swapped for the custom `sort_versions`

Stratus3D:
I think we should remove this from the docs:

> When invoked, if extension commands do not have their executable-bit set, they will be
sourced as bash scripts, having all of the functions from $ASDF_DIR/lib/utils.bash available.

I didn't realize that was in there and changes to `utils.bash` functions are not considered to be breaking changes, but if other plugins are relying on them then they are.

Stratus3D:
Closing this in favor of https://github.com/asdf-vm/asdf/pull/1247

PR #683, opened 2020-03-21
Support customizing plugin-add and plugin-remove

### LINKED ISSUE #670: asdf "plugin remove" hook for plugin authors
As with `asdf uninstall <name> <version>` it would be good if plugin authors could perform additional actions on the removal of their plugin.

In my particular instance, I would like to remind users to remove some global config shared across each version of the plugin ([see here for what I mean](https://github.com/jthegedus/asdf-gcloud/blob/0a309723eb2ffccecfb71720f7e2c59cfb576642/bin/uninstall#L7)).

If this exists in the documentation I could not find it. Nor could I with a cursory look through the codebase.

joschi:
* Add support for custom `plugin-add` and `plugin-remove` in plugins
* Add configurable command hooks for plugin installation and removal

```shell
pre_asdf_plugin_remove = echo will remove plugin ${1}
pre_asdf_plugin_remove_foo = echo will remove plugin foo
post_asdf_plugin_remove = echo removed plugin ${1}
post_asdf_plugin_remove_foo = echo removed plugin foo
```

Closes #670

### DIFF
--- a/docs/plugins-create.md
+++ b/docs/plugins-create.md
@@ -88,6 +88,32 @@ Note: This will only apply for users who have enabled the `legacy_version_file` 
 
 This can be used to further parse the legacy file found by asdf. If `parse-legacy-file` isn't implemented, asdf will simply cat the file to determine the version. The script will be passed the file path as its first argument.
 
+#### bin/plugin-add
+
+This can be used to run any post-installation actions after the plugin has been added to asdf.
+
+The script has access to the path the plugin was installed (`${ASDF_PLUGIN_PATH}`) and the source URL (`${ASDF_PLUGIN_SOURCE_URL}`), if any was used.
+
+See also the related hooks:
+
+* `pre_asdf_plugin_add`
+* `pre_asdf_plugin_add_${plugin_name}`
+* `post_asdf_plugin_add`
+* `post_asdf_plugin_add_${plugin_name}`
+
+#### bin/plugin-remove
+
+This can be used to run any pre-removal actions before the plugin will be removed from asdf.
+
+The script has access to the path the plugin was installed in (`${ASDF_PLUGIN_PATH}`).
+
+See also the related hooks:
+
+* `pre_asdf_plugin_remove`
+* `pre_asdf_plugin_remove_${plugin_name}`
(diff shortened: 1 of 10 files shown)


vic:
Awesome, thanks a lot @joschi

docs/plugins-create.md
```diff
@@ -88,6 +88,32 @@ Note: This will only apply for users who have enabled the `legacy_version_file`
...
+This can be used to run any post-installation actions after the plugin has been added to asdf.
+
+The script has access to the path the plugin was installed (`${ASDF_PLUGIN_PATH}`) and the source URL (`${ASDF_PLUGIN_SOURCE_URL}`), if any was used.
+
+See also the related hooks:
+
+* `pre_asdf_plugin_add`
+* `pre_asdf_plugin_add_${plugin_name}`
+* `post_asdf_plugin_add`
+* `post_asdf_plugin_add_${plugin_name}`
```
jthegedus: @joschi @vic where are these documented? Seems odd to reference these hooks in the guide here if they aren't documented anywhere else.

Also, I think we should have pre and post in the names of the files:

- post-plugin-add
- pre-plugin-remove

joschi: @jthegedus These aren't filenames but hook names which can be used in `.asdfrc`.

I didn't find any documentation for the other hooks (such as `pre_asdf_reshim_${plugin_name}` or `pre_asdf_uninstall_${plugin_name}` except for the changelog entries, so I decided not to spend too much time on documenting these either.

See also: https://github.com/asdf-vm/asdf/search?q=asdf_run_hook&type=Code

joschi: > Also, I think we should have pre and post in the names of the files:
> 
> * post-plugin-add
> * pre-plugin-remove
>
> As it will better communicate the intent.

Yes, I think this makes a lot of sense. Should I create a follow-up PR?

jthegedus: Another PR would be appreciated 🙏

jthegedus: I will add documenting those missing hooks to my list of docs updates

joschi: See #684.

Issue #659, opened 2020-02-24
Already installed plugins throwing stderr


odelucca:
### Steps to reproduce

- Install a new plugin, `asdf plugin-add nodejs`, for example
- Try to install the same plugin again

You should notice that the message that the plugin were already installed was an `stderr`

### Expected behavior

Since this is not actually an error, it should only return the message, without any error status. This is important because some automation scripts can't handle this.

For example, I'm developing an Ansible role that install a bunch of plugins on my workstation, and I need to manually ignore those errors. This can cause a lot of trouble.

### Actual behavior

It is being executed like if it was an error.

jthegedus:
In you reproduction, are you "importing the Node.js release team's OpenPGP keys to main keyring" as in the [plugin's instructions](https://github.com/asdf-vm/asdf-nodejs#install)?

odelucca:
Yes, the error that I ge tis specific regarding that `nodejs` was already installed.

As I've said, on the first install it works smoothly, but in the next one it throws an error saying that `nodejs` was already installed.

I need it not to print anything on `stderr` since this is not actually an error. I'm building my own auomated workstation setup with Ansible and it may try to install those multiple times.

jthegedus:
Fair enough. Thanks.

jthegedus:
@odelucca This is being reverted as the prior discussions from essentially the same question resulted in the exit code of 2 for adding a plugin that already exists - see original issue https://github.com/asdf-vm/asdf/issues/322 and the actual discussion in the PR https://github.com/asdf-vm/asdf/pull/323

Stratus3D:
If you have code that needs to verify that a plugin is installed, and don't care if it was already installed, something like this should do what you need:

```
status_code=$(asdf plugin-add <name> <url>)
if [ $status_code -eq 0 ] || [ $status_code -eq 2 ]; then
    echo "Plugin is installed"
fi
```

odelucca:
Strange.

In my opinion `exit 1` should be only used when an error occur. This is not actually an error and adding `exit 1` makes harder to automated workstation setup.

For example, I use `ansible` to automate the installation of my tools. For every tool that I use, everytime I ran the `install` command it syncs my entire configs. Most of those I try to reinstall.

`asdf-vm` is the only tool that throws `exit 1` when I try to install an already present plugin. For that instance I need to do a condition (like @Stratus3D mentioned) to avoid this messing up with my sync.

[...]

PR #688, opened 2020-03-30
Change exit code to 0 and output message to stdout

PauloRSF:
# Summary

Changed exit code to 0 and message output to stdout when trying to add an already added plugin

Fixes: #659

### DIFF
--- a/lib/commands/command-plugin-add.bash
+++ b/lib/commands/command-plugin-add.bash
@@ -27,8 +27,8 @@ plugin_add_command() {
   mkdir -p "$(asdf_data_dir)/plugins"
 
   if [ -d "$plugin_path" ]; then
-    display_error "Plugin named $plugin_name already added"
-    exit 2
+    echo "Plugin named $plugin_name already added"
+    exit 0
   else
     asdf_run_hook "pre_asdf_plugin_add" "$plugin_name"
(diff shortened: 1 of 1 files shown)


Stratus3D:
I think we intentionally chose an exit code of 2 to indicate when a plugin already exists. Exit code of 1 is for a plugin fails to be added due to any error from the `git clone` command. And an exit code of 0 is for plugins that successfully clone down.

Stratus3D:
Sorry for the late comment. I think we should revert this change. We added this exit code for a reason.

Stratus3D:
This also is not a proper fix for #659. Any automated script that may want to ignore this "already added plugin" exit code can do that with `asdf plugin add ... || true`.

Stratus3D:
I have created https://github.com/asdf-vm/asdf/pull/689 to revert these changes and add a unit test.

PauloRSF:
Ok, fair enough. But, is a non-zero exit code really needed for this? I mean, using "|| true" would work but would ignore real errors. In a script, trying to add an already added plugin won't have any unexpected effects, so why not exit with 0?

Stratus3D:
Looked through the issue tracker and found where we added this exit code:

* Issue: https://github.com/asdf-vm/asdf/issues/322
* Pull Request: https://github.com/asdf-vm/asdf/pull/323

Sadly we didn't write unit tests for it.

PR #697, opened 2020-04-10
Add each subcommand

sakuro:
# Summary

This PR adds `each` subcommand to asdf which will run given command (after the plugin name) for all installed versions of the tool/language.

## Example usage

To run `gem update` for all installed rubies:

```
$ asdf each ruby gem update
## ruby 2.6.5 ##
Updating installed gems
:
## ruby 2.7.1 ##
Updating installed gems
:
```

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -28,7 +28,7 @@ find_asdf_cmd() {
   case "$1" in
     'exec' | 'current' | 'env' | 'global' | 'install' | 'latest' | 'local' | \
       'reshim' | 'uninstall' | 'update' | 'where' | 'which' | \
-      'export-shell-version')
+      'export-shell-version' | 'each')
       echo "$asdf_cmd_dir/command-$1.bash" 2
       ;;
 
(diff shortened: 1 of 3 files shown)


Stratus3D:
I'm not sure about this feature. This can be done with a oneline `for` loop in Bash:

```
for version in $(asdf list ruby | xargs); do; asdf shell ruby "$version"; gem update; done
```

I'm reluctant to add an entirely new command to asdf to do this. `for` loops are relatively simple and easy to use.

Stratus3D:
Would a "cookbook" style resource be useful for this sort of thing? I understand that the above command will not be obvious to many people, but I also see a lot of unnecessary overhead to adding a new command to asdf.

jthegedus:
I think a cookbook would be useful. I feel these types of commands we should encourage people to setup as aliases for their shell.

sakuro:
@Stratus3D I understand your policy.  Feel free to close 😀

Stratus3D:
@jthegedus I'm going to start keeping a document that I'll add commands like this too when I see things like this. Hopefully I'll be able to post a good list of commands for tasks like this.

PR #699, opened 2020-04-18
Documentation install section update

jthegedus:
# Summary

Hi all, this is an update to the `Install asdf-vm` documentation. The goal with these changes was to make the install and shell config docs more clear for each user depending on their setup and reduce the visual noise.

To date we've been using [`docsify-tabs`](https://jhildenbiddle.github.io/docsify-tabs/#/) to allow readers to view the docs they wish to see for their system. This worked okay, but didn't allows us to easily display content for a combination of options (say OS, shell and method of installation) and so our guide showed instructions which may not have been relevant to some users and made it difficult to explain specific corner cases. Below we show Git and Homebrew installation for Bash on macOS which makes commenting on each specific setup harder to communicate (like Homebrew's completions) or the ZSH change.

Screenshot from 2020-04-18 11-56-23

My vision for the docs aligned more with the [Yarnpkg Installation page](https://classic.yarnpkg.com/en/docs/install/) where users choose the OS they have and the page only shows the content for their needs:

<details>
<summary>Yarnpkg install screen</summary>

Screenshot from 2020-04-18 12-00-37

</details>

To this end, I developed [docsify-select](https://github.com/jthegedus/docsify-select) which allows showing content based on a combination of a number of HTML Select menus. This was based on the `docsify-tabs` Docsify plugin, so kudos to it's author. In addition to the simple Markdown definition of the menus, syncing of the selections across the page was mirrored and a new feature of automatic Operating System detection added. Run this PR locally or see https://jthegedus.github.io/docsify-select/#/ for a live example of this plugin.

Screenshot from 2020-04-18 14-51-37

I believe this greatly simplifies the user experience of our docs.

Improvements made in this PR:

- [x] `docsify-select` over `docsify-tabs`
- [x] code snippets referring to changes in `.*rc` files are now the content that needs to be edited and not a shell command to append to the end of the file. This was made as most `.*rc` files are very custom and appending to the end isn't universal. Also it makes it easier to read for those not editing `.*rc` files often.
- [x] separation of the "Add to shell" instructions and the code to enable Completions

Issues with this implementation:

- duplicate examples. Seen across the `macOS+Fish+Git` and `Linux+Fish+Git` instructions and `macOS+ZSH+Git` and `Linux+ZSH+Git` instructions. I wanted to use [Docsify File Embeds](https://docsify.js.org/#/embed-files) and extract these common pieces out to their own files to remove the duplication and just embed them multiple times, but Docsify has issues with the processing of embeds, so until that is fixed we will just maintain duplicate sections.

Future Improvements:

- `docsify-select` save selections in local storage to keep selections across page reloads
- `docsify-select` improved support for partial selections. Allow duplicate content for these selection combinations (`macOS+Fish+Git` and `Linux+Fish+Git`) to be covered by a single `*+Fish+Git` selection which matches both of the above combinations.

To test yourself:
- clone this branch
- install `docsify-cli` with `npm i docsify-cli`
- run `docsify serve docs`


docs/index.html
```diff
@@ -85,7 +98,7 @@
     <style>
       :root {
         /* Reduce the font size */
-        --base-font-size: 14px;
```
jthegedus, line 88: I also removed the custom font size to go back to the default Docsify font size of 16px as it is more standard

jthegedus:
@Stratus3D glad you like it too.

Regarding the increase of installation options, we can certainly simplify this once I improve the `docsify-select` plugin and multiple Docsify File Embeds are corrected. Each combination could be taken out to their own `.md` file to make maintenance easier.

I will merge and raise the pacman addition with another PR to improve the wording/update other parts of the documentation

Issue #698, opened 2020-04-16
asdf crash with error "sort: unrecognized option '--version-sort'"


m0ast:
Hello,
On osx some commands fails with `sort: unrecognized option '--version-sort'`

**To Reproduce**
On Osx try
`asdf plugin test nodejs https://github.com/asdf-vm/asdf-nodejs.git node --version`
or since after investigating it seems related
`asdf latest nodejs`

**Expected behavior**

In first case should invoke nodejs with `--version` on the second, should list latest for nodejs

**Actual behavior**

```
➜ asdf plugin test nodejs https://github.com/asdf-vm/asdf-nodejs.git node --version
Updating nodejs...
Already on 'master'
Your branch is up-to-date with 'origin/master'.
sort: unrecognized option `--version-sort'
Try `sort --help' for more information.
nodejs  is already installed
unknown command: node. Perhaps you have to reshim?
FAILED: node --version failed with exit code 0
```
or with latest
```
➜ asdf latest nodejs
sort: unrecognized option `--version-sort'
Try `sort --help' for more information.
```

### Environment

**OS**:
```
sw_vers
ProductName:    Mac OS X
ProductVersion: 10.12.6
BuildVersion:   16G2136
```

**asdf version**: v0.7.8-4a3e3d6

**asdf plugins installed**: ```elixir
erlang
haxe
neko
nodejs```

Ps: Actually these commands aren't listed for Cli usage

jthegedus:
Good find. This is the line causing the issue https://github.com/asdf-vm/asdf/blob/7cdfc0fa2a5f0f924a4486a971f5110102f58e54/lib/commands/command-latest.bash#L15

We should be using https://github.com/asdf-vm/asdf/blob/8a5c70b2dd43f918c51548a52c150c3c9ff4a547/lib/commands/command-update.bash#L46 instead. I will raise a fix.

jthegedus:
Should be fixed on latest commit and in our next release

PR #700, opened 2020-04-18
fix: remove sort --version-sort from command-latest as list is already sorted

jthegedus:
# Summary

`sort`'s `--version-sort` flag isn't available on all OSs as reported in the linked issue and the reason for a custom `sort_versions` function in `command-update.bash` https://github.com/asdf-vm/asdf/blob/8a5c70b2dd43f918c51548a52c150c3c9ff4a547/lib/commands/command-update.bash#L46

Since the `asdf list-all "$plugin_name"` part of the command returns a sorted list, we do not need to perform any sort here. We should be able to safely remove it.

Fixes: #698 

## Other Information

I originally replaced this sort with the aforementioned custom sort_versions function and it seems that it is not idempotent. Something of note since we recommend this function for plugin authors.


jthegedus:
@asdf-vm/core Anyone see any issues with this?

jthegedus:
@Stratus3D does this scan to you?

Stratus3D:
Sorry for the extremely late review @jthegedus 🤦 . Looks good to me!

PR #709, opened 2020-04-24
Documentation: Fix Fish source syntax

JamesMcMahon:
# Summary

The previous instructions did not work on modern Fish (3.1.0) and lead to the following error:

> source: Error encountered while sourcing file '(brew --prefix asdf)/asdf.fish':
> source: No such file or directory


jthegedus:
@asdf-vm/core These required GitHub status checks never get fulfilled as the GitHub Worflow's exclude running on changes to docs. Our options are to run all GitHub Workflows on all changes or not have required checks. Unfortunately GitHub doesn't yet support variably-required checks

Issue #718, opened 2020-04-30
Syntax error when getting the preset version


mauriciovieira:
**Describe the bug**

There is a syntax error inside utils.bash `select_from_preset_error()` function. This might be related to #650 
**To Reproduce**

Tell us what actions you performed before the issue occurred

1. Run `sh -x $HOME/.asdf/bin/asdf`
2. You'll get

```
+++ dirname /Users/mauriciovieira/.asdf/bin/asdf
++ dirname /Users/mauriciovieira/.asdf/bin
+ source /Users/mauriciovieira/.asdf/lib/utils.bash
++ GREP_OPTIONS=--color=never
++ GREP_COLORS=
++ ASDF_DIR=/Users/mauriciovieira/.asdf
++ ASDF_DATA_DIR=
/Users/mauriciovieira/.asdf/lib/utils.bash: line 638: syntax error near unexpected token `('
/Users/mauriciovieira/.asdf/lib/utils.bash: line 638: `  grep -f <(get_shim_versions "$shim_name") <(preset_versions "$shim_name") | head -n 1 | xargs echo'
```
3. Notice the error on the last 2 pasted lines

**Expected behavior**

No syntax error.

**Actual behavior**

Syntax error

### Environment

**OS**: `Darwin XXXX.local 18.7.0 Darwin Kernel Version 18.7.0: Mon Feb 10 21:08:45 PST 2020; root:xnu-4903.278.28~1/RELEASE_X86_64 x86_64`

**asdf version**: `v0.7.8-4a3e3d6`

**asdf plugins affected (if relevant)**: all of them

**asdf plugins installed**: ruby

**Additional context**

I suspected about this when some rails application started behaving weirdly here.

mauriciovieira:
This bug report was a mistake. `sh -x` should not be used. 

I couldn't reproduce the error anymore, so I'll consider this not an error.

Issue #650, opened 2020-02-04
Can't find shims for system version of plugin


dynamotn:
### Steps to reproduce
```sh
~ asdf global ruby system
~ ruby --version
ruby 2.5.7p206 (2019-10-01 revision 67816) [x86_64-linux]
~ gem environment
RubyGems Environment:
  - RUBYGEMS VERSION: 2.7.10
  - RUBY VERSION: 2.5.7 (2019-10-01 patchlevel 206) [x86_64-linux]
  - INSTALLATION DIRECTORY: /home/dynamo/.gem/ruby/2.5.0
  - USER INSTALLATION DIRECTORY: /home/dynamo/.gem/ruby/2.5.0
  - RUBY EXECUTABLE: /usr/bin/ruby25
  - EXECUTABLE DIRECTORY: /home/dynamo/.gem/ruby/2.5.0/bin
  - SPEC CACHE DIRECTORY: /home/dynamo/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /etc
  - RUBYGEMS PLATFORMS:
    - ruby
    - x86_64-linux
  - GEM PATHS:
     - /home/dynamo/.gem/ruby/2.5.0
     - /usr/local/lib64/ruby/gems/2.5.0
     - /usr/lib64/ruby/gems/2.5.0
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - /home/dynamo/.asdf/shims
     - /home/dynamo/.asdf/bin
     - /home/dynamo/.fzf/bin
     - /home/dynamo/.local/bin
     - /home/dynamo/.fzf/bin
     - /home/dynamo/.local/bin
     - /home/dynamo/.fzf/bin
     - /home/dynamo/.local/bin
     - /home/dynamo/.fzf/bin
     - /usr/local/sbin
     - /usr/local/bin
     - /usr/sbin
     - /usr/bin
     - /sbin
     - /bin
     - /opt/bin
     - /usr/lib/llvm/9/bin
~ gem install solargraph
Successfully installed solargraph-0.38.5
Parsing documentation for solargraph-0.38.5
Done installing documentation for solargraph after 0 seconds
1 gem installed
~ solargraph
No solargraph executable found for ruby system
```
### Expected behavior
[...]

PR #719, opened 2020-04-30
Fix select_from_preset_version

mauriciovieira:
# Summary

I reported a syntax error in #718 that affects asdf when getting the preset version of the packages.

Fixes: #718 

## Other Information

I reported #718 but this bugfix might also correct the behavior described in #650

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -635,7 +635,7 @@ preset_versions() {
 
 select_from_preset_version() {
   local shim_name=$1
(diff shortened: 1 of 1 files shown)


Stratus3D:
@mauriciovieira thanks for the PR. I am not sure line 638 is correct. Also, the shellcheck is complaining about line 638 as well.

Stratus3D:
Reading this indicates to me the subshell syntax is not what we want - https://unix.stackexchange.com/questions/393349/difference-between-subshells-and-process-substitution

PR #721, opened 2020-05-08
fix: remove `column` command in favor of awk

homburg:
- asdf plugin list
- asdf plugin list all

Avoid hard dependency on `column` command. This command
is not included in a fresh ubuntu container and not
defined in posix:

https://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html

Also makes the output of `asdf plugin list all` line-streamed
on the second column value

### DIFF
--- a/lib/commands/command-plugin-list-all.bash
+++ b/lib/commands/command-plugin-list-all.bash
@@ -20,7 +20,7 @@ plugin_list_all_command() {
 
         printf "%s\\t%s\\n" "$index_plugin_name" "$installed_flag$source_url"
       done
-    ) | (column -t -s $'\t' 2>/dev/null || awk '{ printf("%-28s", $1); sub(/^[^*]/, " &", $2); $1=""; print $0 }')
+    ) | awk '{ printf("%-28s", $1); sub(/^[^*]/, " &", $2); $1=""; print $0 }'
   else
     printf "%s%s\\n" "error: index of plugins not found at " "$plugins_index_path"
   fi
--- a/lib/commands/command-plugin-list-all.bash
+++ b/lib/commands/command-plugin-list-all.bash
@@ -10,17 +10,23 @@ plugin_list_all_command() {
   plugins_local_path="$(get_plugin_path)"
 
   if ls "$plugins_index_path" &>/dev/null; then
-    (
+      list=""
+      width=0
       for index_plugin in "$plugins_index_path"/*; do
         index_plugin_name=$(basename "$index_plugin")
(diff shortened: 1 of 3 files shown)


Stratus3D:
It looks like one of the tests failed because of the change in command output. Can you correct the failing test?

Stratus3D:
Why not just get rid of `column` entirely and just always use `awk`?

homburg:
Column has an advantage in auto-sizing the first column. Awk just has a hardcoded size for the first column.

I can remove the column call if you prefer.

jthegedus:
After looking into many options for this solution, I think we **should remove `column`** and extract the `awk` scripts into our own `format` file for us to manage our screen output as functions in their own place.

Adding deps to `asfd` is difficult (though we are pre v1.0 so now would be the time). We're already using `awk` so as long as we keep `awk` script to POSIX compliant than we should be good.

If the content changes to require further modifications to column padding it shouldn't be too difficult to update. Annoying, yes, but less so than getting all users and systems to rely on another dep.

homburg:
@jthegedus Thanks for the feedback, I've removed the `column` calls

jthegedus:
@Stratus3D are we okay with merging this, and then having a follow-up PR that addresses converting all other tabular output to `awk` and potentially word-wrapping?

lib/commands/command-plugin-list-all.bash
```diff
@@ -20,7 +20,7 @@ plugin_list_all_command() {
 
         printf "%s\\t%s\\n" "$index_plugin_name" "$installed_flag$source_url"
       done
-    ) | column -t -s $'\t'
+    ) | awk '{ printf("%-28s", $1); sub(/^[^*]/, " &", $2); $1=""; print $0 }'
```
Stratus3D, line 23: Why do these two awk commands differ? The `column` commands they replace are the same.

Stratus3D:
Note to whoever merges: probably best to squash and merge this PR.

lib/commands/command-plugin-list-all.bash
homburg: This one:

1. prints the first column at 28 char width, left-aligned
2. Adds a space to column 2's without leading "*" for alignment

```
28-chars-wide                 next-column
28-chars-wide                *column-with-asterisk
```

The other one just prints the _first_ column at 28 char width, left-aligned

homburg:
@jthegedus Are your change requests addressed or do you have further requests?

jthegedus:
@homburg You've addressed them, was just waiting to see your response to @Stratus3D 's questions. LGTM too, will merge

PR #733, opened 2020-05-28
Fix dead link by changing the extension to sh

raimon49:
# Summary

Hi, I tried creating a plugin for asdf-vm. I found a dead link in the documentation and fixed it.

The reason is that [the commit changed the file extension](https://github.com/asdf-vm/asdf/commit/52cbf1165c79c7efbb1c9b2f26167a047a7335d0).

I also think there is a way to change the link to the master branch.However, that selection line number is unstable. e.g) https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash#L98

This P-R keeps a commit hash to stabilize the line number. What do you think?

My plugin was easy to create. Thanks to the maintainers of the documentation.

### DIFF
--- a/docs/plugins-create.md
+++ b/docs/plugins-create.md
@@ -53,7 +53,7 @@ This script should install the version, in the path mentioned in `ASDF_INSTALL_P
 
 The install script should exit with a status of `0` when the installation is successful. If the installation fails the script should exit with any non-zero exit status.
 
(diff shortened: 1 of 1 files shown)


Stratus3D:
Ideally we'd link to the latest version of the function, but linking to master will be unstable as you said, so I think what you did here is best.

jthegedus:
Versioning strategies for the documentation is on my mind, will get to it eventually :sweat_smile:

Issue #715, opened 2020-04-26
asdf shell: set plugin version for shell session only


noraj:
![image](https://user-images.githubusercontent.com/16578570/80313246-07223200-87ea-11ea-9484-4358397b044a.png)

We would like a command to set a version of a plugin for the current shell session only. 

**Describe the solution you'd like**

Add `asdf shell` in addition to `asdf {global, local}` to set a version only for the current shell session.

**Describe similar asdf features and why they are not sufficient**

rbenv is the core project that gave birth to nodenv, pyenv, phpenv, <whateverlangage>env. It seems that asdf is totally inspired from rbenv [using the same principles](https://github.com/rbenv/rbenv/wiki/Why-rbenv%3F) so I'm very surprised [rbenv shell](https://github.com/rbenv/rbenv#rbenv-shell) is not implemented.

**Describe workarounds you've considered**

It's already possible to set a version just for a shell session via environment variables but it's not very handy.

Stratus3D:
This command actually already exists, and does exactly what you are describing.

@jthegedus what do you think about adding documention for `asdf shell` to this section of the website? https://asdf-vm.com/#/core-manage-versions?id=set-current-version We've got a info box that mentions the environment variable, but we don't mention the `asdf shell` command anywhere.

jthegedus:
Yeah no problem :+1: It seems the `asdf` command doesn't list the `asdf shell` command either, so we should add it to that output as well.

PR #737, opened 2020-06-08
fix: asdf shell command docs & help output

noraj:
# Summary

Add the documentation and output for the shell command

Fixes: #715

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -27,7 +27,7 @@ find_asdf_cmd() {
   asdf_cmd_dir="$(asdf_dir)/lib/commands"
   case "$1" in
     'exec' | 'current' | 'env' | 'global' | 'install' | 'latest' | 'local' | \
-      'reshim' | 'uninstall' | 'update' | 'where' | 'which' | \
+      'reshim' | shell | 'uninstall' | 'update' | 'where' | 'which' | \
       'export-shell-version')
       echo "$asdf_cmd_dir/command-$1.bash" 2
       ;;
--- a/bin/asdf
+++ b/bin/asdf
@@ -27,7 +27,7 @@ find_asdf_cmd() {
   asdf_cmd_dir="$(asdf_dir)/lib/commands"
   case "$1" in
     'exec' | 'current' | 'env' | 'global' | 'install' | 'latest' | 'local' | \
-      'reshim' | shell | 'uninstall' | 'update' | 'where' | 'which' | \
+      'reshim' | 'uninstall' | 'update' | 'where' | 'which' | \
       'export-shell-version')
       echo "$asdf_cmd_dir/command-$1.bash" 2
       ;;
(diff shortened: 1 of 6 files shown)


bin/asdf
```diff
@@ -27,7 +27,7 @@ find_asdf_cmd() {
   asdf_cmd_dir="$(asdf_dir)/lib/commands"
   case "$1" in
     'exec' | 'current' | 'env' | 'global' | 'install' | 'latest' | 'local' | \
-      'reshim' | 'uninstall' | 'update' | 'where' | 'which' | \
+      'reshim' | shell | 'uninstall' | 'update' | 'where' | 'which' | \
```
Stratus3D, line 30: Was there a problem when `shell` was not listed here?

docs/core-manage-versions.md
```diff
@@ -60,12 +60,15 @@ asdf latest <name> <version>
...
 ```shell
 asdf global <name> <version> [<version>...]
+asdf shell <name> <version> [<version>...]
 asdf local <name> <version> [<version>...]
 # asdf global elixir 1.2.4
 ```
 
 `global` writes the version to `$HOME/.tool-versions`.
 
+`shell` set the version to an environment variable `ASDF_$LANG_VERSION`, for the current shell session only.
```
Stratus3D, line 70: What about "set the version as an environment variable named `ASDF_${LANG}_VERSION`, for the current shell session only."?

Stratus3D:
Overall looks good. Do you have more changes planned for this PR @noraj ?

bin/asdf
noraj: You want me to remove it?

noraj:
> Overall looks good. Do you have more changes planned for this PR @noraj ?

No, I put [WIP] as I'm not sure my PR is complete, so if someone with better knowledge of the project want to complete it before merging, else I'm good.

bin/asdf
jthegedus: `shell` command has an escape hatch in an earlier call in the stack as linked as it may require exporting variables which is supported by calling https://github.com/asdf-vm/asdf/blob/master/lib/commands/command-export-shell-version.bash

I think this should be removed to avoid confusion.

jthegedus:
This test needs updating. https://github.com/asdf-vm/asdf/blob/45fffa9ce3c09d317f3ccef2ed0415e56bc550eb/test/current_command.bats#L95

Stratus3D:
Looks like we've still got some failing tests in the build.

noraj:
> Looks like we've still got some failing tests in the build.

https://github.com/asdf-vm/asdf/pull/737/checks?check_run_id=765306992

```
not ok 236 where should error when no current version selected and version not specified
# (in test file test/where_command.bats, line 63)
#   `[ "$output" = "$expected" ]' failed
```

Issue #738, opened 2020-06-10
Installation script directing users to doc site


jthegedus:
**Is your feature request related to a problem? Please describe.**

Homebrew users expect the `brew install asdf` process to complete the installation of `asdf`. As we require shell config changes this obviously doesn't happen and we subsequently get Issues about `asdf` not working after a Homebrew install.

**Describe the solution you'd like**

I would like the documentation site logged during the 'install' process with a link and message to the docs to complete the setup process.

My suggestion is a minimal bash script that users can run using `curl`/`wget` that performs 2 things:
- git clone (as our current install method)
- log a link to our docs site to complete the Shell setup process

This would be like the `nvm` install script: https://github.com/nvm-sh/nvm#install--update-script IE:

```sh
curl -o- https://raw.githubusercontent.com/asdf-vm/asdf/v0.7.8/install.sh | bash
# or
wget -qO- https://raw.githubusercontent.com/asdf-vm/asdf/v0.7.8/install.sh | bash
```

Installing would be:
- Regular users run this script, logging out the doc site and warning.
- Advanced users can still just `git clone` as before.
- Homebrew would run this script, logging out the doc site and warning.

IMPORTANTLY, I am not suggesting we do what `nvm` does with this script and editing peoples `.<shell>rc` files

**Describe similar asdf features and why they are not sufficient**

Git Clone doesn't allow us to output anything during the Homebrew installation.

**Describe workarounds you've considered**

An attempt was made to use Homebrew's built-in "caveats" logging to do this, but this apparently doesn't align with intended Homebrew usage - see https://github.com/Homebrew/homebrew-core/pull/54315

**Additional context**

Reducing erroneous bug reports would allow us to focus elsewhere and improve the Developer Experience.

jthegedus:
Feedback from @asdf-vm/core would be great.

Stratus3D:
I have a couple concerns about this:

1. If I understand correctly, this would be another way of installing asdf. Currently we have installation via package managers (e.g. brew) or manually via Git clone. I don't really want us to have to support a third installation method.
2. Does this installation method offer any benefits over the manual `git clone` approach? It doesn't appear to automate anything. I'd rather see users standardize on installation via package manager than on running a script.

I guess I'd rather see us focus on improving installation via OS package managers. If I'm not mistaken we could just put the installation instruction code in the Homebrew formula, and then we wouldn't have to have conditional logic in the script as we would know already that it is a Homebrew installation.

[...]

PR #739, opened 2020-06-10
feat: install script for logging on installation

jthegedus:
# Summary

Use an install script to perform the `git clone` installation process. This allows us to log on installation which is required for the Homebrew installation as outlined in #738 (Homebrew users expect `brew install asdf` to perform the complete setup of asdf and subsequently report a GitHub issue before going to the docs site to see further instructions)

Fixes: #738 

## Other Information

- [x] update tag regex (probably can be improved as I'm no `sed` expert)
- [x] we no longer have instructions in README so deprecate that tag update
- [x] change instructions to `curl`/`wget` the install script from the GitHub remote
- [x] use `--git-dir` and `--work-tree` to avoid filesystem traversal as `git` provides this functionality

## Notes

- the `git checkout` uses `--quiet` flag to supress "detached head" warning on tag checkout. Perhaps we let it log this?
- potentially could reuse `ASDF_DATA_DIR` instead of introducing another env var

---

Testing output:

with `--quiet`:

```shell
asdf on  feat/install-script [!] 
➜ ./install.bash        
Cloning into '/home/jthegedus/projects/test'...
remote: Enumerating objects: 31, done.
remote: Counting objects: 100% (31/31), done.
remote: Compressing objects: 100% (26/26), done.
remote: Total 5350 (delta 12), reused 13 (delta 5), pack-reused 5319
Receiving objects: 100% (5350/5350), 1007.11 KiB | 808.00 KiB/s, done.
Resolving deltas: 100% (3010/3010), done.

asdf setup
1: install asdf                 completed!
2: add asdf to your shell       https://asdf-vm.com/#/core-manage-asdf-vm
3: add a plugin                 https://asdf-vm.com/#/core-manage-plugins
4: install a tool version       https://asdf-vm.com/#/core-manage-versions
```

without `--quiet`:

```sell
asdf on  feat/install-script [!] 
➜ ./install.bash        
Cloning into '/home/jthegedus/projects/test'...
remote: Enumerating objects: 31, done.
remote: Counting objects: 100% (31/31), done.
remote: Compressing objects: 100% (26/26), done.
remote: Total 5350 (delta 12), reused 13 (delta 5), pack-reused 5319
Receiving objects: 100% (5350/5350), 1007.11 KiB | 808.00 KiB/s, done.
Resolving deltas: 100% (3010/3010), done.
Note: switching to 'v0.7.8'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 4a3e3d6 Update version to 0.7.8

asdf setup
1: install asdf                 completed!
2: add asdf to your shell       https://asdf-vm.com/#/core-manage-asdf-vm
3: add a plugin                 https://asdf-vm.com/#/core-manage-plugins
4: install a tool version       https://asdf-vm.com/#/core-manage-versions
```

### DIFF
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -5,7 +5,7 @@
 
 ### --Git--
 
-Run the install script to `git clone` the latest branch:
+Run the install script to `git clone` the latest release:
 
 ```shell
 curl -o- https://raw.githubusercontent.com/asdf-vm/asdf/v0.7.8/install.bash | bash
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -5,18 +5,12 @@
 
 ### --Git--
 
-Clone only the latest branch:
+Run the install script to `git clone` the latest branch:
 
 ```shell
-git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.7.8
-```
-
-Alternately, you can clone the whole repo and checkout the latest branch:
-
-```shell
-git clone https://github.com/asdf-vm/asdf.git ~/.asdf
-cd ~/.asdf
-git checkout "$(git describe --abbrev=0 --tags)"
+curl -o- https://raw.githubusercontent.com/asdf-vm/asdf/v0.7.8/install.bash | bash
+# or
+wget -qO- https://raw.githubusercontent.com/asdf-vm/asdf/v0.7.8/install.bash | bash
 ```
 
 ### --Homebrew--
--- a/install.bash
+++ b/install.bash
@@ -6,14 +6,15 @@ set -euo pipefail
 #ORIGINAL_IFS=$IFS
 IFS=$'\t\n' # Stricter IFS settings
 
-asdf_install_dir=${ASDF_INSTALL_DIR:-"~/.asdf"}
+asdf_install_dir=${ASDF_INSTALL_DIR:-"$HOME/.asdf"}
 
 git clone https://github.com/asdf-vm/asdf.git "${asdf_install_dir}"
-# checkout latest tag
-git --git-dir "${asdf_install_dir}" checkout "$(git --git-dir "${asdf_install_dir}" describe --abbrev=0 --tags)"
+# checkout latest tag in repo in another dir
+# credit: https://stackoverflow.com/a/6073628/7911479 and https://stackoverflow.com/a/31811385/7911479
+git --git-dir "${asdf_install_dir}/.git" --work-tree "${asdf_install_dir}" checkout "$(git describe --abbrev=0 --tags)" --quiet
 
-printf "asdf setup"
-printf "1: install asdf - completed!"
-printf "2: add asdf to your shell - https://asdf-vm.com/#/core-manage-asdf-vm"
-printf "3: add a plugin - https://asdf-vm.com/#/core-manage-plugins"
-printf "4: install a tool version - https://asdf-vm.com/#/core-manage-versions"
+printf "\n%s\n" "asdf setup"
+printf "%s\t\t\t%s\n" "1: install asdf" "completed!"
+printf "%s\t%s\n" "2: add asdf to your shell" "https://asdf-vm.com/#/core-manage-asdf-vm"
+printf "%s\t\t\t%s\n" "3: add a plugin" "https://asdf-vm.com/#/core-manage-plugins"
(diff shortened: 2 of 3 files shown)


docs/core-manage-asdf-vm.md
```diff
@@ -59,18 +59,12 @@ No match for _Operating System_ and _Installation Method_ selections. Please try
...
-
-Alternately, you can clone the whole repo and checkout the latest branch:
-
-```shell
-git clone https://github.com/asdf-vm/asdf.git ~/.asdf
-cd ~/.asdf
-git checkout "$(git describe --abbrev=0 --tags)"
+curl -o- https://raw.githubusercontent.com/asdf-vm/asdf/v0.7.8/install.bash | bash
+# or
+wget -qO- https://raw.githubusercontent.com/asdf-vm/asdf/v0.7.8/install.bash | bash
```
Stratus3D, line 67: I don't think we should adopt this pattern. I know it's popular but it has some risks and I don't think we should push our users in this direction.

See:

* https://sysdig.com/blog/friends-dont-let-friends-curl-bash/
* https://www.seancassidy.me/dont-pipe-to-your-shell.html
* https://news.ycombinator.com/item?id=12766049

install.bash
```diff
@@ -0,0 +1,20 @@
...
+git clone https://github.com/asdf-vm/asdf.git "${asdf_install_dir}"
+# checkout latest tag in repo in another dir
+# credit: https://stackoverflow.com/a/6073628/7911479 and https://stackoverflow.com/a/31811385/7911479
+git --git-dir "${asdf_install_dir}/.git" --work-tree "${asdf_install_dir}" checkout "$(git describe --abbrev=0 --tags)" --quiet
+
+printf "\n%s\n" "asdf setup"
+printf "%s\t\t\t%s\n" "1: install asdf" "completed!"
+printf "%s\t%s\n" "2: add asdf to your shell" "https://asdf-vm.com/#/core-manage-asdf-vm"
+printf "%s\t\t\t%s\n" "3: add a plugin" "https://asdf-vm.com/#/core-manage-plugins"
+printf "%s\t%s\n" "4: install a tool version" "https://asdf-vm.com/#/core-manage-versions"
```
Stratus3D, line 20: I'm not sure that providing links provides much benefit to the user. Why can't we just tell the user what do to add asdf to their shell? If this script is invoked by the package manager we should be able to tell the user exactly what do to for their specific OS/package manager combination.

jthegedus: I think that solution is unmaintainable for us. The number of permutations is making the docs difficult to update as is. `nvm` does this and their script is 100s of lines of conditional shell scripting which it almost certainly doesn't work in all cases.

docs/core-manage-asdf-vm.md
jthegedus: I agree with all of this. Given it is the install method of choice for Homebrew & OMZSH users I didn't think it was too much an ask, especially since others can still just `git clone`. Additionally, my understanding of prior discussion of package manager code in this repo was that we didn't want to maintain those either.

install.bash
jthegedus: >I'm not sure that providing links provides much benefit to the user

The idea was to capture users who hear about `asdf` and just run their package manager of choices installation, eg `brew install asdf`, and haven't ever come to the documentation site. These users have find that it doesn't work, they were given no further instructions, and instead of looking for a documentation site to see if there is something they missed, they check GitHub for issues where people used the same package manager and report they themselves are having issues.

I would like to show these people 2 things:
1. there's more steps before you can use `asdf`
2. these steps are on our documentation site

jthegedus: Given we don't need to inform users who are installing from the documentation site, we could leave the docs as is, and only use the install script in Homebrew and co after they clone the repo as they currently do.

jthegedus:
I will create an Issue in the repo to pin for Homebrew users to direct them to the documentation. As most Google the issue and end up finding old Issues and resolving the problem via an old method.

jthegedus:
I will create an Issue in the repo to pin for Homebrew users to direct them to the documentation. As most people Google the issue and end up finding old Issues and resolving the problem via an old method.

PR #742, opened 2020-06-14
docs: synchronize help text

jthegedus:
# Summary

The help text and "All Commands" doc site page have been getting updates differently. They've fallen out of sync. While crude, I would like to solve the discrepancy by directly importing the `help.txt` file into the docs.

I don't imagine that page to be heavily visited, it was only included for completeness of the documentation site. This is a better in-between.  

## Other Information

I removed the 2 space indentation of the `help.txt` for all commands. This allows for more content to be horizontally filled.

I removed the inline word wrapping. I will follow this PR up with a change to render the help text via `awk` on `asdf`/`asdf help` and perform wrapping.

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -20,6 +20,7 @@ MANAGE PACKAGES
   asdf shell <name> <version>              Set the package version in the current shell
   asdf local <name> <version>              Set the package local version
   asdf global <name> <version>             Set the package global version
+  asdf latest <name> [<version>]           Show latest stable version of a package
   asdf list <name>                         List installed versions of a package
   asdf list all <name>                     List all versions of a package
 
--- a/help.txt
+++ b/help.txt
@@ -32,8 +32,7 @@ asdf which <command>                    Display the path to an executable
 asdf local <name> <version>             Set the package local version
 asdf global <name> <version>            Set the package global version
 asdf shell <name> <version>             Set the package version to
-                                        `ASDF_${LANG}_VERSION` in the current
-                                        shell
+                                        `ASDF_${LANG}_VERSION` in the current shell
 asdf latest <name> [<version>]          Show latest stable version of a package
 asdf list <name>                        List installed versions of a package
 asdf list all <name> [<version>]        List all versions of a package and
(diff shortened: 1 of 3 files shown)


help.txt
```diff
@@ -1,33 +1,32 @@
 MANAGE PLUGINS
-  asdf plugin add <name> [<git-url>]       Add a plugin from the plugin repo OR, add a Git repo
-                                           as a plugin by specifying the name and repo url
-  asdf plugin list [--urls] [--refs]       List installed plugins. Optionally show git urls and git-ref.
-  asdf plugin list all                     List plugins registered on asdf-plugins repository with URLs
-  asdf plugin remove <name>                Remove plugin and package versions
-  asdf plugin update <name> [<git-ref>]    Update a plugin to latest commit or a particular git-ref.
-  asdf plugin update --all                 Update all plugins
+asdf plugin add <name> [<git-url>]      Add a plugin from the plugin repo OR, add a Git repo as a plugin by specifying the name and repo url
```
Stratus3D, line 2: I think we should still wrap this text at 80 columns. I work on pretty small screens and my terminal is usually around 80 columns.

help.txt
```diff
@@ -1,33 +1,32 @@
...
-  asdf which <command>                     Display the path to an executable
-  asdf shell <name> <version>              Set the package version in the current shell
-  asdf local <name> <version>              Set the package local version
-  asdf global <name> <version>             Set the package global version
-  asdf list <name>                         List installed versions of a package
-  asdf list all <name>                     List all versions of a package
+asdf install [<name> <version>]         Install a specific version of a package or, with no arguments, install all the package versions listed in the .tool-versions file
+asdf install <name> latest[:<version>]  Install the latest stable version of a package, or with optional version, install the latest stable version that begins with the given string
+asdf uninstall <name> <version>         Remove a specific version of a package
+asdf current                            Display current version set or being used for all packages
```
Stratus3D, line 14: This is what I see with wrapping text. I find it hard to follow when it wraps all the way around. Screen Shot 2020-06-15 at 12 25 30 PM

jthegedus:
My goal here was to remove the inline formatting and rely on the print mechanism to perform formatting. Currently help is output with `cat`, I want to change this to `awk` for consistency. I will revert the inline wrapping (though it wasn't at 80, more like 90-95) and include the final formatting change with `awk` formatting

jthegedus:
Here's a diff from the last two commits showing strict vs relaxed 80 char wrapping. Relaxed doesn't wrap individual words.

https://github.com/asdf-vm/asdf/compare/6638ba0..0f275d1

jthegedus:
I'd like to get this in and follow up later with formatting being moved into the printing function and not the content.

Issue #692, opened 2020-04-06
$HOME/.asdf/completions/asdf.bash:68: command not found: complete


Quintasan:
### Steps to reproduce

0. Use zsh
1. Install asdf
2. Install completions according to [documentation](https://asdf-vm.com/#/core-manage-asdf-vm)

### Expected behavior

No errors happen

### Actual behavior

I'm getting:

```
/home/quintasan/.asdf/completions/asdf.bash:68: command not found: complete
```

on every terminal launch

### Environment

**OS**: Ubuntu 18.04

**asdf version**: v0.7.8-4a3e3d6

**zsh version**: zsh 5.4.2 (x86_64-ubuntu-linux-gnu)

jthegedus:
Interesting. `completions/asdf.bash` should not be executed by any ZSH setup. Which specific setup have you followed for `asdf` completions as the documentation point to a number of ways to setup ZSH completions but none direct you to using the `asdf.bash` script itself.

goetzc:
@jthegedus a couple of Zsh plugins also do this same sourcing of the completion:

https://github.com/kiurchv/asdf.plugin.zsh/blob/master/asdf.plugin.zsh#L5
https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/asdf/asdf.plugin.zsh#L17

jthegedus:
Ah, yes, a number of the asdf plugins will source this as the ZSH completions are new and those plugins have not been updated. I have opened a discussion around how to correctly implement the new completions with the `oh-my-zsh` asdf plugin https://github.com/ohmyzsh/ohmyzsh/issues/8779

As with the `command not found: complete` error, `complete` has been a requirement for the bash completions and these plugins for a few years, so you will need to install [`bash-completion`](https://github.com/cykerway/complete-alias)

jthegedus:
:trumpet: 
The asdf core team are not the authors of most (if any) of these shell plugins, so we are relying on people from the community to help update these plugins for a better ZSH user experience. If you're familiar with any plugins and have the time to help, it would be much appreciated. 

I would advise anyone looking to help to read https://github.com/ohmyzsh/ohmyzsh/issues/8779 as the `fpath`/`compinit` lifecycle of the tool the plugin is for is important for the implementation.

shipstar:
Chiming in with my experience in case it helps someone track this down:

### Works for me

**asdf version**: v0.7.6-6207e42

[...]

PR #744, opened 2020-06-15
Added shell completions for zsh

LexMalta:
# Summary
Simple script to get shell/bash completions for `zsh` working.

I honestly don't know much about shell completions or how they work, but a quick bit of googling found: https://github.com/eddiezane/lunchy/pull/59#issuecomment-122780541 .I tried that locally and it worked for me - figured it would be worth sharing upstream.

Are there any docs that need to be updated?

### DIFF
--- a/completions/asdf.zsh
+++ b/completions/asdf.zsh
@@ -1,0 +1,3 @@
+autoload bashcompinit
+bashcompinit
+source $(dirname $0)/asdf.bash


jthegedus:
Hi mate, thanks for contributing.

Unfortunately, this is code we used to have in our repo but have since removed. This implementation of using the Bash completions was how we originally supported ZSH users for a few years, but we now have native ZSH completions.

Previous issues were #674 #692 

The documentation site https://asdf-vm.com/#/core-manage-asdf-vm covers how to setup the Shell completions for ZSH vs Bash etc.

If you find yourself itching to contribute more we have plenty of improvements to make, thanks!

Issue #528, opened 2019-05-28
`asdf current` output is not correct for tools that are not installed


Stratus3D:
### Steps to reproduce
Run `asdf current` in a directory with a `.tool-versions` file containing versions that are not installed. Output will look like this:

```
version 1.8.2-otp-21 is not installed for elixir
elixir         
erlang         21.3.8.2 (set by ~/devel/dir/.tool-versions)
version 11.9.0 is not installed for nodejs
nodejs         
```

### Expected behavior

For all tools listed, both the name and the version should be displayed on one line just like the installed versions. e.g.

```
elixir    version 1.8.2-otp-21 is not installed for elixir
```

### Actual behavior

For tools with versions that are not installed, the message for the tool is shown on a separate line.

```
version 1.8.2-otp-21 is not installed for elixir
elixir
```

### Environment

**OS**: OSX 10.14.3

**asdf version**: v0.7.2

jthegedus:
This issue is fixed for me in `0.7.8`:

```
➜ asdf current
elixir         No version set for elixir; please run `asdf <global | local> elixir <version>`
firebase       8.0.2    (set by /home/jthegedus/.tool-versions)
gcloud         288.0.0  (set by /home/jthegedus/.tool-versions)
gradle         6.3      (set by /home/jthegedus/.tool-versions)
hadolint       v1.17.5  (set by /home/jthegedus/.tool-versions)
java           adopt-openjdk-11.0.6+10 (set by /home/jthegedus/.tool-versions)
maven          3.6.3    (set by /home/jthegedus/.tool-versions)
nodejs         12.16.1  (set by /home/jthegedus/projects/docsify-select/.tool-versions)
ocaml          4.10.0   (set by /home/jthegedus/.tool-versions)
python         3.8.2    (set by /home/jthegedus/.tool-versions)
shellcheck     0.7.1    (set by /home/jthegedus/.tool-versions)
terraform      0.12.24  (set by /home/jthegedus/.tool-versions)
```

Or am I misunderstanding the problem?

Stratus3D:
This is still a problem. This issue exists for tool versions that have been set in a `.tool-versions` file but have not be installed. I tested on my OSX machine just now and it is still a problem. After setting a version of Elixir that I do not yet have installed I see:

```
clojure        No version set for clojure; please run `asdf <global | local> clojure <version>`
version 1.10.0 is not installed for elixir
elixir         
[...]

Issue #646, opened 2020-01-30
[enhancement] Wrap terminal line output to 80 characters


jsejcksn:
Scannability would be improved greatly if lines were wrapped at 80 characters.

Here is example terminal output from the command `asdf help`. The longest unwrapped line is 125 characters:

```
asdf help
version: v0.7.6-6207e42

MANAGE PLUGINS
  asdf plugin add <name> [<git-url>]       Add a plugin from the plugin repo OR,
 add a Git repo
                                           as a plugin by specifying the name an
d repo url
  asdf plugin list [--urls] [--refs]       List installed plugins. Optionally sh
ow git urls and git-ref.
  asdf plugin list all                     List plugins registered on asdf-plugi
ns repository with URLs
  asdf plugin remove <name>                Remove plugin and package versions
  asdf plugin update <name> [<git-ref>]    Update a plugin to latest commit or a
 particular git-ref.
  asdf plugin update --all                 Update all plugins


MANAGE PACKAGES
  asdf install [<name> <version>]          Install a specific version of a packa
ge or,
                                           with no arguments, install all the pa
ckage
                                           versions listed in the .tool-versions
 file
  asdf uninstall <name> <version>          Remove a specific version of a packag
e
  asdf current                             Display current version set or being 
used for all packages
  asdf current <name>                      Display current version set or being 
used for package
  asdf where <name> [<version>]            Display install path for an installed
 or current version
  asdf which <command>                     Display the path to an executable
  asdf shell <name> <version>              Set the package version in the curren
t shell
  asdf local <name> <version>              Set the package local version
  asdf global <name> <version>             Set the package global version
  asdf list <name>                         List installed versions of a package
  asdf list all <name>                     List all versions of a package


UTILS
  asdf exec <command> [args..]             Executes the command shim for current
 version
  asdf env <command> [util]                Runs util (default: `env`) inside the
 environment used for command shim execution.
  asdf reshim <name> <version>             Recreate shims for version of a packa
ge
  asdf shim-versions <command>             List on which plugins and versions is
 command available
  asdf update                              Update asdf to the latest stable rele
ase
  asdf update --head                       Update asdf to the latest on the mast
er branch

"Late but latest"
-- Rajinikanth
```

Stratus3D:
Thanks for opening this issue. I think we should limit line lengths to 80 characters so they can fit on terminals of the traditional size.

[...]

PR #746, opened 2020-06-17
fix: incorrect version output in asdf current

jthegedus:
## Issue:

`asdf current` does not correctly output when a version in `.tool-versions` is not found for an installed plugin.

```shell
➜ asdf current
deno           1.0.0    (set by /home/jthegedus/.tool-versions)
version 8.0.2 is not installed for firebase
firebase
gcloud         288.0.0  (set by /home/jthegedus/.tool-versions)
```

## Cause:

The `utils.bash#display_error()` function outputting to `stderr` and the `current_command` handling the rest of the output.

## Goals:

- [x] Properly tabulate the output (fix the original problem)
- [x] Print the row of data to `stdout` for the happy path
- [x] Print the row to `stderr` for the non-happy path (important for scripting)
- [x] If a version is installed and found, `exit 0`
- [x] If a version is not installed, `exit 1`
- [x] If a version is not set `exit 126`.
- [x] Use `printf`

## Evidence:

```shell
.asdf on  master [⇣!?] took 1s 
➜ asdf current
PLUGIN          VERSION         SET BY CONFIG
deno            1.0.0           /home/jthegedus/.tool-versions
firebase        8.0.2           Not installed. Run "asdf install firebase 8.0.2"
gcloud          290.0.0         Not installed. Run "asdf install gcloud 290.0.0"
golang          ______          No version set. Run "asdf <global|shell|local> golang <version>"
gradle          6.3             /home/jthegedus/.tool-versions
hadolint        v1.17.5         /home/jthegedus/.tool-versions
java            adopt-openjdk-11.0.6+10 /home/jthegedus/.tool-versions
maven           3.6.3           /home/jthegedus/.tool-versions
nodejs          12.16.1         /home/jthegedus/.tool-versions
ocaml           4.10.0          /home/jthegedus/.tool-versions
python          3.8.2           /home/jthegedus/.tool-versions
shellcheck      0.7.1           /home/jthegedus/.tool-versions
terraform       0.12.24         /home/jthegedus/.tool-versions

.asdf on  master [⇣!?] took 1s 
➜ echo "$?"
0

.asdf on  master [⇣!?] 
➜ asdf current firebase
PLUGIN          VERSION         SET BY CONFIG
firebase        8.0.2           Not installed. Run "asdf install firebase 8.0.2"

.asdf on  master [⇣!?] 
➜ echo "$?"            
1

.asdf on  master [⇣!?] 
➜ asdf current golang
PLUGIN          VERSION         SET BY CONFIG
golang          ______          No version set. Run "asdf <global|shell|local> golang <version>"

.asdf on  master [⇣!?] 
➜ echo "$?"          
126

.asdf on  master [⇣!?] 
➜ asdf current 1> /dev/null
firebase        8.0.2           Not installed. Run "asdf install firebase 8.0.2"
gcloud          290.0.0         Not installed. Run "asdf install gcloud 290.0.0"
golang          ______          No version set. Run "asdf <global|shell|local> golang <version>"
```

## Further Information:

Output of tabular data must be handled at the top level, in this case `command-current.bash`, as each piece of information needs to be collated into an output row before the entirety can be sent to `stdout`/`stderr`, otherwise these formatting issues will arise.

IMO the shared functions in `utils.bash`, `check_if_version_exists` and `check_if_plugin_exists` are going to cause a lot of pain with #646 and will need to be untangled. `check_if_version_exists` calling `check_if_plugin_exists` and each subsequently outputting branching error states to `stderr` without caller control of the output is the root issue.

I updated `check_if_version_exists` to only report truthiness and pulled out the error text to another function which returned the text instead of printing to `stdout`/`stderr` leaving all control to the caller. This affected `lib/commands/command-export-shell-version.bash` and `lib/commands/version_commands.bash` and was caught by their tests. What I didn't account for was the `check_if_plugin_exists` error outputs. Our test cases are yet to cover scenarios where `check_if_plugin_exists` `stderr` logging is within tabular data.

`check_if_version_exists` is a much less called function than `check_if_plugins_exists` so I tackled that here to solve the initial issue. Addressing these same concerns with `check_if_plugins_exists` will be a lot more impactful to the codebase and has been left for later (probably #646). In the meantime I updated the test cases for `lib/commands/command-export-shell-version.bash` and `lib/commands/version_commands.bash` to account for the extra logging my change introduced as it's not easily tackled without addressing everything else mentioned.

## Future

IMO we need to address in future:

- `asdf current` test cases covering more tabular data, specifically cases where we get errors from `check_if_plugin_exists`
- stop `check_if_version_exists` calling `check_if_plugin_exists`
- remove `stderr` logging from `check_if_plugin_exists` and bubble to callers

---

<details>
<summary>initial PR summary</summary>

# Summary

`asdf current` does not correctly format the output when a version in `.tool-versions` is not found for an installed plugin.

```shell
➜ asdf current
deno           1.0.0    (set by /home/jthegedus/.tool-versions)
version 8.0.2 is not installed for firebase
firebase
gcloud         288.0.0  (set by /home/jthegedus/.tool-versions)
```

The output should be:

```shell
➜ asdf current
deno           1.0.0    (set by /home/jthegedus/.tool-versions)
firebase       version 8.0.2 is not installed for firebase
gcloud         288.0.0  (set by /home/jthegedus/.tool-versions)
```

or perhaps even

```shell
➜ asdf current
deno           1.0.0    (set by /home/jthegedus/.tool-versions)
firebase       8.0.2	version is not installed
gcloud         288.0.0  (set by /home/jthegedus/.tool-versions)
```

:thinking: thoughts?

Fixes: #528 

## Other Information

I dislike this **initial** solution.

I do not think the logic for `utils.bash` `check_if_version_exists` should display any errors. I feel it should just perform the check, and have the caller use the returned `0` or `1` return value to then itself handle output and any calls to `display_error`.

The other call sites of `check_if_version_exists` are:
- `command-export-shell-version.bash#L32`
- `version_commands.bash#L37`

If no one is opposed, I will make my desired changes tomorrow.

</details>


jthegedus:
After diving much deeper into this problem the core issue is we're sharing rendering logic for commands with a single line output and tabular output.

`utils.bash#display_error()` outputs to `stderr` which is the cause for the misaligned output.

To render via `stdout` we need to redirect from `stderr` to `stdout` for the call to `display_error`. Since that defeats the purpose of using `stderr` I don't think it is a good solution.

This leaves us with two options for printing data:
1. print to `stderr` for the individual row of data that would use `stderr` for a single plugin/version.
2. don't use any calls to stderr for tabular rendering

Personally, I think we should render the tabular data via `stderr` if the command would do so for a singular plugin/version. I will pursue this type of solution unless someone has concerns.

Stratus3D:
Is this ready for review? I see it still has the WIP label.

jthegedus:
Very much WIP. Happy for comments, though I'm very far past this solution atm. Just documenting my thoughts/progress for myself and reviewers when the time comes. CODEOWNERS auto tagged sorry

jthegedus:
@Stratus3D All good, as mentioned in the title PR message, this solution has a few downstream implications and identified other issues which we should address later. I'll raise separate Issues for those.

juliogc:
Sorry if it's not the right place to comment, but this PR implemented a "bug" in others softwares that uses `asdf current` like the spaceship prompt.

image

Because new headers on the output, it is loosing his string reference and pointing up my ruby version as `PLUGIN`

**Before**
image

**After**
image

I will also report this issue on spaceship github.

jthegedus:
@juliogc I'm sorry my change caused this issue. Long-term we want all commands to output header rows with a flag for scripts to use to toggle the headers off, in the short-term I think we can comment out the header. What do you think @Stratus3D

jthegedus:
The `SET BY CONFIG` heading holds context for the information in the right column (I think the rest is easily inferred) which is why I have yet to just remove the heading row, though that might just have to do for now

jthegedus:
@juliogc Notably, we will be releasing these header row additions at some point, perhaps after `0.8.0` stable release, so there will be a period where it will not be ideal for you to use the repo `head` as tools like spaceship will need to migrate to support our newer outputs

Stratus3D:
> @juliogc I'm sorry my change caused this issue. Long-term we want all commands to output header rows with a flag for scripts to use to toggle the headers off, in the short-term I think we can comment out the header. What do you think @Stratus3D

As I previously mentioned on #646:

> Tools like git provide a --porcelain flag that causes commands to format output in an easy to parse format. Typically this is used by scripts that want to parse the output of a git command with another command in Bash. We could implement a similar feature. With the --porcelain flag things like indentation and column headers would be omitted.

I think adding a flag like this would cover @juliogc use case.

> Notably, we will be releasing these header row additions at some point, perhaps after 0.8.0 stable release

When do you want these changes to be released? I've only tagged the 0.8.0 RC1 release. The 0.8.0 release hasn't been tagged yet, will be doing it this week hopefully.

juliogc:
> @juliogc Notably, we will be releasing these header row additions at some point, perhaps after `0.8.0` stable release, so there will be a period where it will not be ideal for you to use the repo `head` as tools like spaceship will need to migrate to support our newer outputs

@jthegedus Don't worry! For now I fixed the issue pointing to `v0.8.0-rc1` and will keep attention on the `head` in further releases.

I know the scpaceship it's only a PS1 theme and I can really live whithout it, but I only made the comment as a advice having in mind that could exist some critical automation that could break.

Thank you so much for your attention and the quicky fix!

Issue #749, opened 2020-06-23
Show path to active .tool-versions in error messages


eproxus:
**Is your feature request related to a problem? Please describe.**
When an older `.tool-versions` file exists _somewhere_ in the folder hierarchy above the current working directory, it is hard to know why e.g. a `asdf global <name> <version>` command doesn't work. For example this output:
```
$ asdf global erlang 23.0.2
$ erl
asdf: No preset version installed for command erl
Please install the missing version by running one of the following:

asdf install erlang 21.3.8.10

or add one of the following in your .tool-versions file:

erlang 23.0.2
```

When you are deep in a directory hierarchy, it can sometimes be hard to know if you have a `.tool-version` somewhere. After getting this error, I ran:

```
$ asdf which erlang
unknown command: erlang. Perhaps you have to reshim?
```

(_**Note:** `erlang` is obviously wrong, because `which` needs a command name, but I didn't know that_). After reshimming, nothing changed of course. 

**Describe the solution you'd like**

The error message should look something like this:
```
$ pwd
/Users/user/some/path/with/many/folders/and/projects
$ erl
asdf: Version 21.3.8.10 not installed for command erl
(set by /Users/user/some/path/.tool-versions)

Please install the missing version by running one of the following:

    asdf install erlang 21.3.8.10

or add one of the following in your .tool-versions file at /Users/user/some/path/.tool-versions:

    erlang 23.0.2
```

**Describe similar asdf features and why they are not sufficient**

`asdf current` lists which version is active and from which `.tool-versions` file it came from. However, it never occurred to me to run this command because I knew which version I set globally already. I also didn't know that this version listed the `.tool-version` file, I only discovered this in desperation when running all kinds of other commands to try to fix the issue.

**Describe workarounds you've considered**

None.

PR #750, opened 2020-06-27
fix: print tool-versions file dir on shim error

jthegedus:
# Summary

Add the path to the active `.tool-versions` file on shim errors

```shell
➜ asdf env gcloud            
No version set for command gcloud
Please add one of the following versions in your config file at /home/jthegedus/.tool-versions

gcloud 287.0.0
gcloud 288.0.0
gcloud 293.0.0

asdf-gcloud on  master 
➜ asdf env gcloud
No preset version installed for command gcloud
Please install a version by running one of the following:

asdf install gcloud 290.0.0

or add one of the following versions in your config file at /home/jthegedus/.tool-versions

gcloud 287.0.0
gcloud 288.0.0
gcloud 293.0.0
```

Fixes: #749 

## Other Information

- moved from `echo` to `printf`
- slimmed error output

### DIFF
--- a/test/shim_exec.bats
+++ b/test/shim_exec.bats
@@ -90,7 +90,7 @@ teardown() {
   [ "$status" -eq 126 ]
 
   echo "$output" | grep -q "No version set for command dummy" 2>/dev/null
-  echo "$output" | grep -q "Please add one of the following versions in your config file at $PROJECT_DIR/.tool-versions" 2>/dev/null
+  echo "$output" | grep -q "Consider adding one of the following versions in your config file at $PROJECT_DIR/.tool-versions" 2>/dev/null
   echo "$output" | grep -q "dummy 1.0" 2>/dev/null
   echo "$output" | grep -q "dummy 2.0" 2>/dev/null
 }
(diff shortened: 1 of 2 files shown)


test/shim_exec.bats
```diff
@@ -90,7 +90,7 @@ teardown() {
   [ "$status" -eq 126 ]
 
   echo "$output" | grep -q "No version set for command dummy" 2>/dev/null
-  echo "$output" | grep -q "you might want to add one of the following in your .tool-versions file" 2>/dev/null
+  echo "$output" | grep -q "Please add one of the following versions in your config file at $PROJECT_DIR/.tool-versions" 2>/dev/null
```
Stratus3D, line 93: There may be some benefit to the more vague `You might want to` wording as setting the version in the `.tool-versions` file is the wrong action in a couple scenarios:

* When you are using legacy version file support and should set or update a legacy version file like `.ruby-version`.
* When you should add the version to your global `.tool-versions` file or another `.tool-versions` file in a parent directory.

Also, should we be directing users to add versions to their `.tool-versions` files or direct them to run `asdf local|global <tool> <version>`? I don't feel strongly either way, but both are valid.

jthegedus: >When you are using legacy version file support 

Fair point.

>Also, should we be directing users to add versions to their .tool-versions files or direct them to run asdf local|global <tool> <version>? I don't feel strongly either way, but both are valid.

Since we have commands to manage the editing of the version file, we should probably suggest that. I will update now.

Handling `legacy_version` on appropriate commands should be addressed across the board, perhaps in another issue.

jthegedus: `asdf local` creates a `.tool-versions` if there is only the `asdf global` version file available, so perhaps editing `.tool-versions` is okay in this instance 🤔

jthegedus: I've taken not of these points and will craft issues to address them at a proj-wide level

PR #754, opened 2020-07-06
Add `column` to list of banned commands for the asdf codebase

Stratus3D:
# Summary

Add `column` to list of banned commands for the asdf codebase. Tests now pass since https://github.com/asdf-vm/asdf/pull/721 has been merged.

Fixes: https://github.com/asdf-vm/asdf/issues/661 
Related: https://github.com/asdf-vm/asdf/pull/721

PR #755, opened 2020-07-07
fix: ban the sort -V command

Stratus3D:
Addresses @jthegedus comment on #754


jthegedus:
Do these bans also capture the long flag name, ie `--sort-versions`?

Stratus3D:
> Do these bans also capture the long flag name, ie `--sort-versions`? @Stratus3D

good catch! No they don't. I need to update this.

Issue #512, opened 2019-04-11
Make dependencies installation more straightforward 


danhper:
As discussed a little in #511, it is not always easy to get all the dependencies right for a plugin.

Although I have been using asdf and asdf-python for a while, I often forget to install tk or sqlite dev libraries before compiling Python and find myself having to recompile it later on.

I think this process could be improved, although I am not really sure how yet.
I agree with @Stratus3D that dependencies should be part of the plugin rather than asdf core.

Here are the main difficulties I can think about

* We have tons of plugins, so dependencies must be maintained at the plugin level
* Too many different platforms, so it is not reasonable to hope that every plugin maintainer will maintain a list as comprehensive as the one found in [pyenv deps page](https://github.com/pyenv/pyenv/wiki#suggested-build-environment). I do think it is fine if we focus on the most popular platforms though (macOS and Ubuntu?)
* Many plugins will have optional dependencies. For example [asdf-erlang](https://github.com/asdf-vm/asdf-erlang)
* Some plugins dependencies may, or may not, be handled by system package managers. For example, in the case of [adsf-elixir](https://github.com/asdf-vm/asdf-elixir), some people might want to install Erlang through their package manager while others might want to do so with asdf.

Most of these are not major blockers but things that I think we should at least consider if we are going to try to do something about this issue.

jthegedus:
On this topic and in case other find this useful, my plugins (eg [`asdf-gcloud`](https://github.com/jthegedus/asdf-gcloud)) have a `check_dependencies` function that loops over a text file of my dependencies and runs `command -v` for each. On `plugin add` it warns of missing dependencies. On `install` it exits if there are missing dependencies.

- https://github.com/jthegedus/asdf-gcloud/blob/5a74763ba33731c076b615827f67a9e4ed5cf658/lib/utils.bash#L28
- https://github.com/jthegedus/asdf-gcloud/blob/master/lib/dependencies.txt

The main motivation was to stop failures on installation leaving a bad state, though I think this can be a better way for people to find the dependencies than what is usually a partial list in the README.

Originally posted by @jthegedus in #511 [(comment)](https://github.com/asdf-vm/asdf/issues/511#issuecomment-605539567)

Stratus3D:
Interested in everyone's input on the specification I wrote for the plugin `bin/documentation` callback in my open PR - https://github.com/asdf-vm/asdf/pull/757.

PR #757, opened 2020-07-09
feat: plugin documentation callback support 🎉

Stratus3D:
# Summary

* Add support for an optional `help` callback scripts for plugins
* Document the required behavior for the `help` callback scripts on the create plugins page
* Write basic unit tests for the new code

Fixes: #512


jthegedus:
I'll give this a test over the next 48hrs with a plugin.

Externally sourced data is going to be tricky to format well. The only solution to this would be where we get control of parsing the content before writing to STDOUT, and in that we could either capture (redirect) the output from the plugin or provide variables for the plugin to populate which we would then print.

I raise this now (though we can certainly iterate on this) because I would like to address output to both STD OUT/ERR across the board for 0.9.0 milestone (I'll create all the issues and milestone for this tomorrow)

Stratus3D:
@jthegedus yes that is one of my concerns with this. The output is so open ended that even if all plugins comply with the requirements I've specified here there will still be variations between plugins. I don't think this is ideal.

One alternative solution would be to have multiple callbacks for each kind of help for each plugin. So we could have a callback named `bin/help.overview`, `bin/help.deps`, `bin/help.config`, `bin/help.links`, etc... Then asdf core would be free to use (or ignore if needed) the output from these scripts in whatever way we deem best. And the scripts themselves would output to STDOUT with minimal formatting. The downside is that we could have to call and support more callbacks (and allow for missing callbacks as not every plugin would need them all). Perhaps this is kind of how man pages work? I've not looked into how they are organized, but this seems like it might be similar (with the exception that these are executable scripts that can vary their output based on the environment).

Stratus3D:
I'm open to whatever everything thinks is best. I only wrote this code today and haven't used it extensively on my own, so I'll try to get more experience with this feature before proceeding with anything.

jthegedus:
I like that our plugin interface for commands requires a specific format, and that we capture the output from the plugins and parse them ourselves in the core to STDOUT. Perhaps we can apply the same thing here requiring specific formats for each type of documentation?

- `bin/help.overview` is freeform text. We will take this and print to 80 cols
- `bin/help.deps` is a newline or space separated list (like `asdf list all`) which we render as a column of data
- `bin/help.config` what type of format would we expect people to use here?
- `bin/help.links` same as `bin/help.deps`

What are your thoughts on this?

Whether this would be implemented as separate callbacks is another question. It is perhaps better to separate them so the output can be format in a more distinct way. IF they were separate and optional, would we notify the users at all or just skip as we do now? (I'm just trying to consider the feedback we're getting regarding exit codes and our use of stderr and if this is something users would like to know at all. IE is the absence of `bin/help.deps` indicative of no deps, or just the absence of the plugin author to document the deps?)

Re man pages: I am also unsure how they work and perhaps it is time I learned 😅 

I don't want to hold up the work being done here, but realise changing our plugin API later may cause issues and it **may** be easier to address these now (though we're pre 1.0 so that's part of the semver contract)

Stratus3D:
> Perhaps we can apply the same thing here requiring specific formats for each type of documentation?
>
>    bin/help.overview is freeform text. We will take this and print to 80 cols

Makes sense. Freeform text should be good.

>    bin/help.deps is a newline or space separated list (like asdf list all) which we render as a column of data

A single column of data representing package names? like:

```
coreutils
libxslt
...
```

This may be too restrictive. Sometimes dependencies may need to be available in OS package managers, and some package managers may lack certain packages, so some instructions on manually installation of dependencies may be needed occasionally. I guess we could make things work if each line of output represented only one dependency.

>    bin/help.config what type of format would we expect people to use here?

I think this would have to be freeform as well. Configuration instructions will vary greatly between tools and operating systems. Instructions may include setting environment variables, creating/editing config files, etc... To much to enforce a strict structure on the output.

>    bin/help.links same as bin/help.deps

This can be one link per line. Formats could be:

```
Link to API docs: http://somedomain.com/with/docs/
Another Link: http://someotherdomain.com/
```

Or:

```
http://somedomain.com/with/docs/
http://someotherdomain.com/
```

Even with both formats it should give us enough control over the final output.

Stratus3D:
> Whether this would be implemented as separate callbacks is another question. It is perhaps better to separate them so the output can be format in a more distinct way. IF they were separate and optional, would we notify the users at all or just skip as we do now?

I think the callbacks should be separate and optional. I think if say the `bin/help.overview` callback were missing we would just report that documentation for the plugin is not available and not look for any additional callbacks.

>(I'm just trying to consider the feedback we're getting regarding exit codes and our use of stderr and if this is something users would like to know at all. IE is the absence of bin/help.deps indicative of no deps, or just the absence of the plugin author to document the deps?)

Not certain, but I guess perhaps that should indicate no deps? I think plugin authors should either implement all needed callbacks for help, or none of them at all.

Stratus3D:
@jthegedus I found this answer about man pages helpful - https://unix.stackexchange.com/a/3587/74959

Working on this again today. Hopefully will get everything updated in a couple hours.

Stratus3D:
@jthegedus the code has been updated based on our discussion. I may write some additional unit tests later today, but either way the code is ready for you to review again.

lib/commands/command-help.bash
```diff
@@ -31,9 +31,84 @@ asdf_extension_cmds() {
...
+    asdf_help
+    asdf_extension_cmds
+    asdf_moto
+  fi
+}
+
+print_plugin_help() {
+  local plugin_path=$1
+
+  # Eventually @jthegedus or someone else will format the output from these
```
jthegedus, line 97: Yes!

Issue #759, opened 2020-07-09
Install single tool from .tool-versions


fajpunk:
**Is your feature request related to a problem? Please describe.**
We have a large repo that contains code that uses many different tools.  We'd like to test different things in this repo in different CI jobs, and we don't want to install _all_ of the tools specified in `.tool-versions` for every job, but we do want to install the correct versions of some subset of the tools we need, at the versions specified in  `.tool-versions`.

For example, we have repo containing a ruby app and Terraform config to provision infrastructure for this app.  We want to validate the Terraform config in a CI job, so we want to install the correct version of Terraform, but we don't want to take the time/resources to install Ruby.

**Describe the solution you'd like**
One possible implementation would be to leave the version off of the install command:
```asdf install terraform```
And have that install the version of terraform specified in `.tool-versions`

**Describe workarounds you've considered**
I'm aware that if a plugin is not installed that matches a tool in `.tool-versions` then that tool won't be installed with a bare `asdf install`, but the environment that I'm running these jobs in may have all matching plugins installed.

I've also considered putting different `.tool-versions` files in the different directories containing the different code, but it is nice for development to be able to do a bare `asdf install` and have all the tools that I'll need.  I want to have my cake and eat it too :)

jthegedus:
> I want to have my cake and eat it too :)
Don't we all! 😄 

`asdf install` -> install all versions defined in `.tool-versions`.
🆕 `asdf install <name>` -> install version for tool <name> in `.tool-versions`.
`asdf install <name> <version>` -> install specified version for tool.

This type of behaviour makes sense to me, especially if it speeds up my CI on a large project.

jthegedus:
Thanks @fajpunk !

ulwlu:
@jthegedus 
Hi. I strongly want to `plugin add` automatically as well.

What do you think?

If you are agreed, I will send PR.

jthegedus:
@ulwlu I strongly disagree with automatic install of `plugin add`. I will write up why in a GitHub Discussion and share here as this Issue covers a different feature request which has been implemented.

ulwlu:
I see, ok I will wait for your link.

In the case I understand that reason, I will give up.

trothwell:
Perhaps a better way than: `cut -d\  -f1 .tool-versions|grep -E '^[^#]'|xargs -L1 asdf plugin add`

givensuman:
[...]

PR #760, opened 2020-07-09
Install one tool specified in .tool-versions

fajpunk:
# Summary
Install one tool specified in .tool-versions

Add a one-argument version of install command `asdf install <name>` that installs that single tool at the version specified in `.tool-versions`.  If there is no `.tool-versions` or if that tool is not in it, an error message is shown and `asdf` exits `1`.

Fixes: #759

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -15,9 +15,9 @@ asdf plugin update --all                Update all plugins
 MANAGE PACKAGES
 asdf install                            Install all the package versions listed
                                         in the .tool-versions file 
-asdf install [<name>]                   Install one tool at the version
+asdf install <name>                     Install one tool at the version
                                         specified in the .tool-versions file
-asdf install [<name> <version>]         Install a specific version of a package
+asdf install <name> <version>           Install a specific version of a package
 asdf install <name> latest[:<version>]  Install the latest stable version of a
                                         package, or with optional version,
                                         install the latest stable version that
--- a/help.txt
+++ b/help.txt
@@ -13,13 +13,11 @@ asdf plugin update --all                Update all plugins
 
 
 MANAGE PACKAGES
+asdf install                            Install all the package versions listed
+                                        in the .tool-versions file 
+asdf install [<name>]                   Install one tool at the version
+                                        specified in the .tool-versions file
 asdf install [<name> <version>]         Install a specific version of a package
-                                        With no arguments, install all the
-                                        package versions listed in the
-                                        .tool-versions file.  With only one
-                                        argument, install that tool at the
-                                        version specified in the .tool-versions
-                                        file.
 asdf install <name> latest[:<version>]  Install the latest stable version of a
                                         package, or with optional version,
                                         install the latest stable version that
--- a/help.txt
+++ b/help.txt
@@ -14,9 +14,12 @@ asdf plugin update --all                Update all plugins
 
 MANAGE PACKAGES
 asdf install [<name> <version>]         Install a specific version of a package
-                                        or, with no arguments, install all the
+                                        With no arguments, install all the
                                         package versions listed in the
-                                        .tool-versions file
+                                        .tool-versions file.  With only one
+                                        argument, install that tool at the
+                                        version specified in the .tool-versions
+                                        file.
 asdf install <name> latest[:<version>]  Install the latest stable version of a
                                         package, or with optional version,
                                         install the latest stable version that
--- a/docs/core-configuration.md
+++ b/docs/core-configuration.md
@@ -28,6 +28,8 @@ python 3.7.2 2.7.15 system
 
 To install all the tools defined in a `.tool-versions` file run `asdf install` with no other arguments in the directory containing the `.tool-versions` file.
 
+To install a single tool defined in a `.tool-versions` file run `asdf install <name>` in the directory containing the `.tool-versions` file.  The tool will be installed at the version specified in the `.tool-versions` file.
+
 Edit the file directly or use `asdf local` (or `asdf global`) which updates it.
 
 ## \$HOME/.asdfrc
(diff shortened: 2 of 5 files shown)


jthegedus:
Hey @fajpunk thanks for the PR 💯 

I don't think the documentation for the command now correctly describes the behaviour, specifically:

```
asdf install [<name> <version>]
```

This syntax to me does not describe the possible actions a user can take. I am leaning towards splitting the instructions out to the 3 options a user can take as it would be clearer, eg:

```
asdf install
asdf install [<name>]
asdf install [<name> <version>]
```

But I would like other peoples opinions here.

Additionally, is this use case now supported?

```
asdf install gcloud firebase 8.5.0 nodejs
```

or are the uses of the 2nd and 3rd option mutually exclusive?

fajpunk:
I like splitting the instructions into those three options in the help text.  I made that change, glad to change it back, or to whatever else folks decide.  It doesn't quite match the way that the 'latest' install invocation is listed though:
```asdf install <name> latest[:<version>]```
If we went this way, it seems like maybe it should be changed to:
```asdf install [<name> latest[:<version>]]```
To be consistent with these other three?

As for:
```asdf install gcloud firebase 8.5.0 nodejs```
I wasn't planning on this change supporting something like that.  As the PR stands, if I'm understanding the code correctly, the install command still only accepts two arguments at most (plus the --keep-download flag for the two-argument case).

I was imagining issuing multiple install commands if you want to mix-and-match like that, something like:
```
asdf install gcloud
asdf install firebase 8.5.0
asdf install nodejs
```

This does bring up a potential issue though.  If you try to pass multiple tools to the install command (which I agree is a reasonable thing to expect), it will usually fail because it will try to install the first tool using the second tool as the version.  But it's _possible_ that the name of the second tool could actually be a valid version of the first tool!

This was in fact the case in the test case that I added:
```asdf install dummy other-dummy```
Until I modifed the mock dummy plugin to fail when passed an `ASDF_INSTALL_VERSION` of `other-dummy`.

1. Are you OK with the mix-and-match case being handled with multiple invocations rather than multiple args to one invocation?
1. Is this potential surprise-success case in the two-arg invocation OK? The only way I can think of to not have this issue is to make this new one-arg invocation a separate command entirely, rather than a variation on `install`.
1. Does the test case I added cover the functionality you were thinking of that needed additional test cases?  I think I may not have understood what you were getting at with that; glad to add/change whatever if I missed it.

**Comment by fajpunk** [comment]
For the help text, I ditched the brackets around the args.  That makes it consistent across all of the install invocations, and I could argue that, since each invocation is mentioned separately, the args to each one are not optional to achieve the functionality described :)

```
asdf install
asdf install <name>
asdf install <name> <version>
asdf install <name> latest[:<version>]
```

File: help.txt [hunk 1]
```diff
@@ -13,13 +13,11 @@ asdf plugin update --all                Update all plugins
 
 
 MANAGE PACKAGES
+asdf install                            Install all the package versions listed
+                                        in the .tool-versions file 
+asdf install [<name>]                   Install one tool at the version
```
jthegedus: The test below indicates that we can only use this to install a single tool at at time, not multiple which `[<name>]` implies. Perhaps just `<name>`?

help.txt
```diff
@@ -13,13 +13,11 @@ asdf plugin update --all                Update all plugins
...
+                                        specified in the .tool-versions file
 asdf install [<name> <version>]         Install a specific version of a package
```
jthegedus: If this is a legal usage

```shell
asdf install gcloud 299.0.0 firebase 8.5.0
```

Then should 

```shell
asdf install gcloud firebase
```

not also be legal when given a `.tool-versions` of

```
gcloud 299.0.0
firebase 8.5.0
```

?

jthegedus:
Sorry @fajpunk I left those comments in the code before reading your comments in this tab.

help.txt
jthegedus: Ignore this thread. See other comment

help.txt
jthegedus: See my thoughts in comment in "conversation" tab

fajpunk:
> But asdf install [<name> <version>] with multiple name and version combos is still valid, no?
I don't think asdf supports this, with or without this PR.  From looking at the code, it looks like if more than two args are ever passed to the install command, they'll get passed as flags to one of the bash install functions, and of those flags, the only only one that's used is `--keep-download`.  I gave it a shot also, with asdf `v0.7.8`, and it only ever installed the first tool and version that I specified.

Is there some option to set or something else I'm missing that enables this behavior?

jthegedus:
>Is there some option to set or something else I'm missing that enables this behavior?

Not that I am aware. I am not fully across all features and aspects of the codebase yet so just assumed from reading the docs that `[]` meant array of items. I would guess I am not the only person to make this assumption, however wrong it may be. Unless it denotes optional params, though surely there's a clearer way to communicate this?

If indeed each of these commands are to be used with a singular tool (<name>), then I guess we don't have any issues.

I would like to get some other feedback @Stratus3D thoughts?

Stratus3D:
@jthegedus I haven't taken `[]` to mean an array of items when referring to the asdf help output. I always interpreted arguments within `[]` as optional arguments.

Regardless of what was originally intended by the help text for the `asdf install` command, I think only supporting one tool (and optionally the version for it) is best. I think the command should stay as simple as possible, and the current install one or all functionality is sufficient for all use cases.

Great work @fajpunk ! I think this is a nice addition.

Stratus3D:
Going to merge this now as I don't see anything else that needs to be discussed.

jthegedus:
Perhaps `[]` is a unix idiom I am not familiar with

Stratus3D:
@jthegedus perhaps you've already seen this, but if not, here is a good StackOverflow answer on the subject - https://stackoverflow.com/a/23242584/1245380

jthegedus:
I had not seen that, thanks @Stratus3D

Issue #771, opened 2020-08-09
Slow shell prompt (bash, zsh, fish) w/ Homebrew (macOS)


wulymammoth:
I'm on macOS Catalina, but I don't believe the version matters here as the issue would likely apply only to those that are using Homebrew. I've had the following issue for a long time and was finally fed up to the point that I began to search for the culprit. This may help someone, so I'd like to update the documentation located at https://asdf-vm.com/#/core-manage-asdf-vm?id=install, but the guide doesn't describe how this can be done

I'm using bash, but I think this, too, applies to Fish and Zsh users -- anyone using Homebrew

**Issue**

It takes a good couple of seconds before my shell prompt is loaded and ready for action. By process of deduction, I discovered that asking Hombrew to find the prefix for a given utility, `asdf` in this case, took anywhere from 2-3 seconds. This would be extremely bothersome for anyone using multiple tabs or even a terminal multiplexer. The documentation asks us to add the following to our `.bash_profile`:

```sh
. $(brew --prefix asdf)/asdf.sh
. $(brew --prefix asdf)/etc/bash_completion.d/asdf.bash
```

**Solution**

Easy - hard-code the Homebrew prefix in a local bash variable or environment variable:
```sh
BREW_PREFIX=/usr/local/opt
. $BREW_PREFIX/asdf/asdf.sh
. $BREW_PREFIX/asdf/etc/bash_completion.d/asdf.bash
```

Less easy - try to speed up Homebrew's prefix option and resolution upstream

jthegedus:
I'm unfamiliar with how `brew --prefix` works, but given there's a command to find it is it possible it can change? If the `asdf` Homebrew package gets updated with a different install location and Brew updates all it's packages, then would this break people's systems?

I have Homebrew on one system, though I barely use `brew` and have not experienced this. Is `brew --prefix` affected by the number of packages installed with Homebrew?

A quick Google landed me here: https://stackoverflow.com/questions/24377639/can-i-change-homebrews-prefix Which references `HOMEBREW_PREFIX` during Homebrew installation. Is this value available in your shell after initializing Homebrew? Can it be used instead?

wulymammoth:
@jthegedus valid questions. I guess I wasn't as clear as I thought

`$HOMEBREW_PREFIX` is set by Homebrew and it appears that Homebrew provides a utility function to apply that prefix to any installed utility, like `asdf` via `$(brew --prefix asdf)`, but as noted, incredibly slow

[...]

PR #774, opened 2020-08-09
fix: revert to cmds to precompute brew prefix on addition to shell config

jthegedus:
# Summary

Revert installation instructions for Homebrew to be commands that append to shell configs instead of the contents to be added. `$(brew --prefix)` will now be computed on adding `asdf` to the shell config instead of being computed on each execution.

Fixes: #771 

## Other Information

This has potential to break when the `asdf` Homebrew package is updated and changes the installation location. The consensus is that we will deal with those issues if/when they arise.


danhper:
Thanks!

PR #777, opened 2020-08-11
make sure the shims folder is appended to path only once

### LINKED ISSUE #767: system version doesn't get resolved correctly
**Describe the bug**

After installing any version of a package, using the system version will cause asdf to hang.

**To Reproduce**

1. Install the nodejs plugin (other plugins show the same behaviour)
2. Install any version of node
3. `asdf global nodejs system`
4. `node`

**Expected behavior**

The system's version of the nodejs repl should start

**Actual behavior**

It hangs. In https://github.com/asdf-vm/asdf/blob/master/lib/commands/command-exec.bash#L16 the variable `executable_path` is always assigned the value `/home/lucas/.asdf/shims/node`, starting an infinite loop.

### Environment

**OS**: Arch Linux 5.7.7-arch-1-1

**asdf version**: v0.7.8

**asdf plugins installed**: asdf-nodejs, asdf-python

vscncls:
# Summary

This makes sure the shims folder is appended to $PATH only once. Before, if the user's $HOME contained a trailing `/` two shims folders would get appended to $PATH (e.g; `/home/lucas//.asdf/shims` && `/home/lucas/.asdf/shim`), preventing the user from using the `system` version as described in #767.

Fixes #767

## Other Information

This pretty much just replicates the logic in this line [https://github.com/asdf-vm/asdf/blob/master/asdf.sh#L24](https://github.com/asdf-vm/asdf/blob/master/asdf.sh#L24
) to the fish script

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -1,10 +1,13 @@
 set -x ASDF_DIR (dirname (status -f))
-set -l asdf_data_dir (
-  if test -n "$ASDF_DATA_DIR"; echo $ASDF_DATA_DIR;
-  else; echo $HOME/.asdf; end)
+
+if test -n $ASDF_DATA_DIR
+  set -l asdf_user_shims $ASDF_DATA_DIR/shims
+else
+  set -l asdf_user_shims $HOME/.asdf
+end
 
 # Add asdf to PATH
-set -l asdf_bin_dirs $ASDF_DIR/bin $ASDF_DIR/shims $asdf_data_dir/shims
+set -l asdf_bin_dirs $ASDF_DIR/bin $asdf_user_shims 
 
 for x in $asdf_bin_dirs
   if test -d $x


asdf.fish
```diff
@@ -1,10 +1,13 @@
 set -x ASDF_DIR (dirname (status -f))
-set -l asdf_data_dir (
-  if test -n "$ASDF_DATA_DIR"; echo $ASDF_DATA_DIR;
-  else; echo $HOME/.asdf; end)
+
+if test -n $ASDF_DATA_DIR
+  set -l asdf_user_shims $ASDF_DATA_DIR/shims
+else
+  set -l asdf_user_shims $HOME/.asdf
```
Stratus3D, line 6: Shouldn't this be `$HOME/.asdf/shims`? The `$HOME/.asdf/` directory won't contain anything worth adding to the `PATH`.

Stratus3D:
Thanks for the fix @vscncls ! There one thing in your changes that differs from what I see in https://github.com/asdf-vm/asdf/blob/master/asdf.sh#L24. Can you fix that as well?

asdf.fish
vscncls: oh, I completely missed that, I'll fix it

PR #778, opened 2020-08-12
fix shims folder path

vscncls:
# Summary

Append `/shims` to `$HOME/.asdf` in $asdf_user_shims.

I also had to move the if statement inside `set`, because if you use the `-l` flag inside the scope of a if statement, the variable doesn't become available in the global scope.

Fixes: https://github.com/asdf-vm/asdf/pull/777#pullrequestreview-465867789

```
> echo $PATH | grep asdf
/home/lucas/.asdf/shims /home/lucas/.asdf/bin ...
```

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -1,10 +1,12 @@
 set -x ASDF_DIR (dirname (status -f))
 
-if test -n $ASDF_DATA_DIR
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR! Changes look good to me. FYI @danhper.

PR #783, opened 2020-08-19
chore: add gitattributes file

jthegedus:
# Summary

This seems relevant now we have users on Windows operating systems (I am one on occasion)

---

~I happy to trim some of these items as I copied from an old project. It seems https://gitattributes.io is down now so I can't grab a new default version.~

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,191 @@
+## GITATTRIBUTES FOR WEB PROJECTS
+#
+# These settings are for any web project.
+#
+# Details per file setting:
+#   text    These files should be normalized (i.e. convert CRLF to LF).
+#   binary  These files are binary and should be left untouched.
+#
+# Note that binary is a macro for -text -diff.
+######################################################################
+
+## AUTO-DETECT
+##   Handle line endings automatically for files detected as
+##   text and leave all files detected as binary untouched.
+##   This will handle all files NOT defined below.
+* text=auto
+
+## SOURCE CODE
+*.bat      text eol=crlf
+*.coffee   text
+*.css      text
+*.htm      text
+*.html     text
+*.inc      text
+*.ini      text
+*.js       text
+*.json     text
+*.jsx      text
+*.less     text
+*.od       text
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,191 @@
...
+*.od       text
+*.onlydata text
+*.php      text
+*.pl       text
+*.py       text
+*.rb       text
+*.sass     text
+*.scm      text
+*.scss     text
+*.sh       text eol=lf
```
Stratus3D, line 39: Many of our files have the extension `.bash`, and one has the extension `.fish`. Do those extensions need to be added here as well?

jthegedus: Yes, I'll update the file extensions we have in our repo

jthegedus: L14 covers anything not explicitly defined here.

jthegedus: `*.sh       text` will default to LF in the git object tree, but not in working directory. So check-in will always be LF and checkout will be system dependent.

.gitattributes
```diff
@@ -0,0 +1,65 @@
...
+*.html     text
+*.js       text
+*.sh       text
+*.bash     text
+*.fish     text
+#### asdf/bin/* explicitly define as text
+*asdf      text
+*asdf-exec text
+*_asdf     text
+#### asdf/test/fixtures/* should be covered by * text=auto on L14
```
jthegedus, line 29: Documentation for an event where `* text=auto` does not pick up any of these files as `text` files. It's possible, but extremely unlikely.

.gitattributes
```diff
@@ -0,0 +1,65 @@
...
+## SOURCE CODE
+*.bats     text
+*.css      text
+*.htm      text
+*.html     text
+*.js       text
+*.sh       text
+*.bash     text
+*.fish     text
+#### asdf/bin/* explicitly define as text
```
jthegedus, line 25: This section is out of an abundance of caution re `* text=auto` as noted in comment on L29

PR #792, opened 2020-08-31
fix: fish init command

dnicolson:
This pull request fixes the fish init script.

Currently, the subcommand works when run from bash-like shells but doesn't work from a fish shell.

Now, the subcommand uses `()` instead of `$()`. Also, the interpolated subcommand is now parsed properly as quotes need to be exited from. Using `source /usr/local/opt/asdf/asdf.fish` instead of `source (brew --prefix asdf)/asdf.fish` often saves well over a second when starting a new terminal session.

### DIFF
--- a/docs/core-manage-asdf-vm.md
+++ b/docs/core-manage-asdf-vm.md
@@ -223,7 +223,7 @@ echo -e "\n. $(brew --prefix asdf)/etc/bash_completion.d/asdf.bash" >> ~/.bash_p
 Add `asdf.fish` to your `~/.config/fish/config.fish` with:
 
 ```shell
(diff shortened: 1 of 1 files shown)


jthegedus:
Thanks for the fix.

>Using source /usr/local/opt/asdf/asdf.fish instead of source (brew --prefix asdf)/asdf.fish often saves well over a second when starting a new terminal session.

Because of the stdout to file redirect `>>` the `brew --prefix asdf` should only be computed once when this command is run. Raised and addressed in https://github.com/asdf-vm/asdf/issues/771

jthegedus (approved):
LGTM

PR #793, opened 2020-09-01
fix: rm asdf current header row for now

jthegedus:
# Summary

https://github.com/asdf-vm/asdf/pull/746 introduced a header row for `asdf current` command. This has had downstream affects with tools that rely on the output, eg spaceship prompt as noted here https://github.com/asdf-vm/asdf/pull/746#issuecomment-684042798

This PR comments out that header row for now. We should probably bundle all header rows changes into a single release of `asdf`, eg `0.9.0`, and not fragment the development over two point releases.

### DIFF
--- a/lib/commands/command-current.bash
+++ b/lib/commands/command-current.bash
@@ -45,7 +45,7 @@ current_command() {
   local terminal_format="%-15s %-15s %-10s\\n"
   local exit_status=0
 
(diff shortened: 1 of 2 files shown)


jthegedus:
Tests will need to be updated before merge

jthegedus:
:pray: sorry for creating the noise

Issue #794, opened 2020-09-02
Unknown command when running Homebrew test


dnicolson:
**Describe the bug**

Running Homebrew tests that require binaries managed by `asdf` can fail, in this instance with this message:
`unknown command: node. Perhaps you have to reshim?`

This appears to be because in the test environment the `HOME` environment variable is [changed](https://github.com/Homebrew/brew/blob/5e5dabcadd38dc62b55e5b541047d926035f4ed8/Library/Homebrew/formula.rb#L1844) by Homebrew. It appears as though `asdf` [depends](https://github.com/asdf-vm/asdf/blob/e9c149ea212cf5bf822bbce8d01aae88caaa30b2/lib/utils.bash#L45) on this.

Homebrew Node.js installations and `nvm` do not have this problem, but as mentioned in https://github.com/asdf-vm/asdf/issues/785 maybe this is unavoidable and should just be added to the Common Homebrew Issues section?

**To Reproduce**

1. brew update
2. brew install emscripten
3. brew test emscripten

**Expected behavior**

The test to run.

**Actual behavior**

The `node` command was not found.

jthegedus:
Thanks for reporting @dnicolson 

If the [`ASDF_DATA_DIR`](https://github.com/asdf-vm/asdf/blob/e9c149ea212cf5bf822bbce8d01aae88caaa30b2/lib/utils.bash#L42) env var was used would this issue be resolved?

Perhaps we should recommend as much to Homebrew users if this is the case? 🤔

dnicolson:
Setting the `ASDF_DATA_DIR` environment variable doesn't seem to have any effect.

There is a predefined list (Java, Go, Rust, CURL) of environment variables that are included in the [Homebrew test environment](https://github.com/Homebrew/brew/blob/5e5dabcadd38dc62b55e5b541047d926035f4ed8/Library/Homebrew/formula.rb#L2153).

Adding `ASDF_DATA_DIR` to that list resolves the issue, but then there is the familiar warning:
 `asdf: No version set for command node`

This is only an issue with locally developing Homebrew `test` methods though (CI works fine), so maybe it should just be documented as a caveat.

jthegedus:
>This is only an issue with locally developing Homebrew test methods
This seems very odd.

[...]

PR #795, opened 2020-09-04
docs: link to Homebrew common issues from documentation site

jthegedus:
# Summary

Link to #785 in the documentation site in the Homebrew installation section as #749 raised an actual caveat/incompatibility between Homebrew and `asdf` (`homebrew test <brew pkg>` has issues if it depends on an `asdf` installed tool)

## Other Information

Given we now have a legitimate caveat with Homebrew in #794 perhaps we can get usage of the caveat logging on Homebrew installation :thinking:


danhper:
LGTM, thanks!

danhper (approved):

Issue #796, opened 2020-09-08
What's asdf-vm meaning?


lihsai0:
I apologize if bother you.

The name `asdf-vm` make me confused.

Is it a abbreviation?

Stratus3D:
The name is actually just "asdf". I believe Akash chose asdf-vm for the org name was just for additional clarification of what asdf is - a Version Manager (vm). The name "asdf" doesn't have any meaning behind it. I believe it was simply chosen because of the arrangement of the home row keys on the keyboard.

jthegedus:
asdf - (left-hand home row)
vm - Version Manager

It is also likely that the GitHub name for `asdf` was already taken.

We could communicate vm = Version Manager better on the website and README

lihsai0:
@Stratus3D @jthegedus Thank you for answer!

PR #798, opened 2020-09-09
chore: remove -vm from common usage as the tool is just asdf

jthegedus:
# Summary

#796 raised questions around the name of `asdf-vm` and how the `-vm` can be confusing. While easy to clarify, the repo and commands are just `asdf` with the org and website being `asdf-vm` because of availability issues.

The **PROPOSAL** here is to remove `-vm` from all documentation and repo just referring to the tool by it's name `asdf` leaving `-vm` to the GitHub org and website.

I don't foresee this causing too much confusion, but it could perhaps generate some, whether the confusion is enough to warrant not doing this is open for debate. A few examples:
- I've added an OR statement to the StackOverflow tag search to cover both `asdf` and `asdf-vm` for legacy reasons 
- My linking people to the documentation site pages like https://asdf-vm.com/#/core-manage-asdf-vm would need updating throughout the Issue history of the project (easily done in ten mins IMO)
- am I missing more???

Fixes: #796

## Other Information

The alternative and my original plan was to clarify the meaning of `vm` in the readme EG:

```diff
-# asdf-vm [![Main workflow](https://github.com/asdf-vm/asdf/workflows/Main%20workflow/badge.svg)](https://github.com/asdf-vm/asdf/actions)
+# asdf-vm: asdf version manager [![Main workflow](https://github.com/asdf-vm/asdf/workflows/Main%20workflow/badge.svg)](https://github.com/asdf-vm/asdf/actions)
```

but that lead me to question where else we should add this clarification (should it also be in the homepage for the docs, each page of the docs, won't all that look a bit messy?). After doing this I decided to see what removing the `-vm` would look like and it seems more simple to me.


Stratus3D (approved):

Stratus3D:
I like this! I agree we should use just `asdf` when possible.

Just to note, on StackOverflow we **must** use the tag `asdf-vm`. The `asdf` tag has already been reserved for the common lisp project with the same name - https://stackoverflow.com/questions/tagged/asdf

jthegedus:
Corrected the link to Stackoverflow to just use `asdf-vm`. Will merge once checks pass :)

jthegedus:
I've surely missed some links when renaming some of these pages, so sorry for that when it bites you.

Issue #801, opened 2020-09-11
Support using "latest" version of a program (`asdf {global,local,shell} <plugin> latest`)


joschi:
`asdf` kind of knows what version `latest` is describing and allows installing the latest version of a program with `asdf install <plugin> latest`, but setting the latest version of a program in the shell, in a directory, or globally requires specifying the exact version.

Example:
```
# asdf install jbang latest
# jbang version
No version set for command jbang
Consider adding one of the following versions in your config file at /Users/joschi/.tool-versions
jbang 0.45.0
# asdf shell jbang latest
version latest is not installed for jbang
```

**Describe the solution you'd like**

It would be great if `asdf {global,local,shell} <plugin> <version>` supported `latest` as version.

```
# asdf install jbang latest
# asdf shell jbang latest
# jbang version
0.45.0
```

Ideally the version ending up in `.tool-versions` would be the resolved latest version of the program, e. g. `jbang 0.45.0` in the example above.

`asdf` already supports this with `asdf latest <plugin>`.

jthegedus:
Currently `latest` is resolved by `asdf`, though the implementation should defer to the plugin to determine what "latest" means for that plugin's ecosystem, see https://github.com/asdf-vm/asdf/issues/763

`latest` in the `.tool-versions` file is a bad idea IMO as it removes guarantees `asdf` provides that each person is on a specific version.

Examples of potential problems with this feature:

- If I have no network access, `asdf` is unable to perform `asdf list all <tool>` to determine `latest`, what happens in this case?
- If I have an older version of the plugin itself to my colleagues, then `latest` may resolve to a different provider leading to erroneous behaviour (EG: Python or Java plugins added a new provider which the project uses but I cannot see this).

I'm sure there are more scenarios like this.

Requiring specific versions, while strict, removes a class of issues and is the intended goal of `asdf` to manage a projects tools' versions.

joschi:
> `latest` in the `.tool-versions` file is a bad idea IMO as it removes guarantees `asdf` provides that each person is on a specific version.

That's what I meant with "Ideally the version ending up in `.tool-versions` would be the resolved latest version of the program". 😃

jthegedus:
Sorry, I didn't understand that to be what you meant. If it worked that way, then it doesn't have any issues I can think of off the top of my head :+1:

aaronmallen:
[...]

PR #802, opened 2020-09-13
feat: support for "latest" in shell, local, and global

joschi:
# Summary

Support the version "latest" in the `shell`, `local`, and `global` commands.

This allows running `asdf shell <plugin> latest` to use the latest version of the program installed with the mentioned plugin.

Example:

```
# asdf plugin add jbang
# asdf install jbang latest
# asdf shell latest
# jbang version
0.45.0
```

Fixes: #801

### DIFF
--- a/test/version_commands.bats
+++ b/test/version_commands.bats
@@ -5,8 +5,9 @@ load test_helpers
 setup() {
   setup_asdf_dir
   install_dummy_plugin
-  install_dummy_version "1.0.0"
-  install_dummy_version "1.1.0"
+  install_dummy_version "1.0"
+  install_dummy_version "1.1"
+  install_dummy_version "2.0"
 
   PROJECT_DIR=$HOME/project
   mkdir -p $PROJECT_DIR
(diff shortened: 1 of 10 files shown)


test/version_commands.bats
```diff
@@ -5,8 +5,9 @@ load test_helpers
 setup() {
   setup_asdf_dir
   install_dummy_plugin
-  install_dummy_version "1.0.0"
-  install_dummy_version "1.1.0"
+  install_dummy_version "1.0"
+  install_dummy_version "1.1"
+  install_dummy_version "2.0"
```
joschi, line 10: Sync versions with the versions emitted by `dummy-plugin`:

https://github.com/asdf-vm/asdf/blob/c6145d08d01691a6e16c990e7d1c8778733e8c76/test/fixtures/dummy_plugin/bin/list-all

jthegedus: I actually think it would be better to sync the other way as semver test versions support any future tests around patch versions (I have an idea in the future for semver support of `asdf plugin update` which would require tests with patch versions)

joschi: Makes sense. I've updated the branch accordingly.

joschi:
The GitHub workflow failed because of an unrelated timeout.

Successful run: https://github.com/joschi/asdf/runs/1111871786

jthegedus (approved):
This looks great to me :) Any objections to this solution from @asdf-vm/core ?

jthegedus:
Merging this to get it in and keep PRs short-lived. If issues arise we can do more PRs :smile:

PR #803, opened 2020-09-17
Fix plugin-update --all when there are no plugins

joelparkerhenderson:
Issue: when asdf is first installed, there are no plugins, and running `asdf plugin-update --all` crashes because the plugin directory doesn't exist; similarly, the command crashes if the plugins directory exists yet doesn't contain any subdirectories.

Solution: This patch changes the directory loop from using a wildcard `plugins/*` to a more robust approach: first verify the plugins directory exists, then use the`find` command to list subdirectories and correctly handle the corner case of no subdirectories.

Future: Consider printing an error message and/or advice message. Consider automatically creating the plugins directory as needed. Consider prompting the user to install some popular plugins.

# Summary

## Other Information

information here.

### DIFF
--- a/lib/commands/command-plugin-update.bash
+++ b/lib/commands/command-plugin-update.bash
@@ -9,10 +9,12 @@ plugin_update_command() {
   local plugin_name="$1"
   local gitref="${2:-master}"
   if [ "$plugin_name" = "--all" ]; then
-    for dir in "$(asdf_data_dir)"/plugins/*; do
-      echo "Updating $(basename "$dir")..."
-      (cd "$dir" && git fetch -p -u origin "$gitref:$gitref" && git checkout -f "$gitref")
-    done
+    if [ -d "$(asdf_data_dir)"/plugins ]; then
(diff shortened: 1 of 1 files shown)


lib/commands/command-plugin-update.bash
```diff
@@ -9,10 +9,12 @@ plugin_update_command() {
   local plugin_name="$1"
   local gitref="${2:-master}"
   if [ "$plugin_name" = "--all" ]; then
-    for dir in "$(asdf_data_dir)"/plugins/*; do
-      echo "Updating $(basename "$dir")..."
-      (cd "$dir" && git fetch -p -u origin "$gitref:$gitref" && git checkout -f "$gitref")
-    done
+    if [ -d "$(asdf_data_dir)"/plugins ]; then
+      for dir in $(find "$(asdf_data_dir)"/plugins -type d -mindepth 1 -maxdepth 1);
```
Stratus3D, line 13: You are missing the `do` keyword on this line.

Stratus3D:
Thanks for the PR! I have addressed the syntax errors and merged your changes into master.

Issue #781, opened 2020-08-18
Consider Banning echo command


Stratus3D:
**Is your feature request related to a problem? Please describe.**

As described in https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/65819#65819, `echo` has a lot of disadvantages and should never be passed uncontrolled data, which is something we often do. 

**Describe the solution you'd like**

Banning the `echo` command entirely would be an effective way of avoiding these issues and would force us to be more explicit with `printf` formatting.

**Describe workarounds you've considered**

We can continue use `echo` when it printing controlled data, but the inconsistency may not be caught by reviewers or our code quality tools. Banning the command would ensure these issues are not possible.

Stratus3D:
Related - http://www.etalabs.net/sh_tricks.html

jthegedus:
I was going to suggest banning `echo` on my output refactor/standardisation. `printf` is definitely my preference here.

Stratus3D:
@jthegedus https://github.com/asdf-vm/asdf/pull/806 is ready for review now.

PR #806, opened 2020-09-21
chore: ban echo command

Stratus3D:
# Summary

* Add `echo` to list of banned commands.
* Replace `echo` calls with calls to `printf`.
* Correct existing `printf` calls that were generating ShellCheck warnings.

Fixes: #781

### DIFF
--- a/lib/commands/command-info.bash
+++ b/lib/commands/command-info.bash
@@ -1,11 +1,11 @@
 # -*- sh -*-
 
 info_command() {
-  printf "%s:\n%s\n\n" "OS" "$(uname -a)"
-  printf "%s:\n%s\n\n" "SHELL" "$($SHELL --version)"
-  printf "%s:\n%s\n\n" "ASDF VERSION" "$(asdf_version)"
-  printf "%s:\n%s\n\n" "ASDF ENVIRONMENT VARIABLES" "$(env | grep -E "ASDF_DIR|ASDF_DATA_DIR|ASDF_CONFIG_FILE|ASDF_DEFAULT_TOOL_VERSIONS_FILENAME")"
-  printf "%s:\n%s\n\n" "ASDF INSTALLED PLUGINS" "$(asdf plugin list --urls)"
+  printf "%s:\\n%s\\n\\n" "OS" "$(uname -a)"
+  printf "%s:\\n%s\\n\\n" "SHELL" "$($SHELL --version)"
+  printf "%s:\\n%s\\n\\n" "ASDF VERSION" "$(asdf_version)"
+  printf "%s:\\n%s\\n\\n" "ASDF ENVIRONMENT VARIABLES" "$(env | grep -E "ASDF_DIR|ASDF_DATA_DIR|ASDF_CONFIG_FILE|ASDF_DEFAULT_TOOL_VERSIONS_FILENAME")"
+  printf "%s:\\n%s\\n\\n" "ASDF INSTALLED PLUGINS" "$(asdf plugin list --urls)"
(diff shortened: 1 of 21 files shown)


lib/commands/command-info.bash
```diff
@@ -1,11 +1,11 @@
...
-  printf "%s:\n%s\n\n" "OS" "$(uname -a)"
-  printf "%s:\n%s\n\n" "SHELL" "$($SHELL --version)"
-  printf "%s:\n%s\n\n" "ASDF VERSION" "$(asdf_version)"
-  printf "%s:\n%s\n\n" "ASDF ENVIRONMENT VARIABLES" "$(env | grep -E "ASDF_DIR|ASDF_DATA_DIR|ASDF_CONFIG_FILE|ASDF_DEFAULT_TOOL_VERSIONS_FILENAME")"
-  printf "%s:\n%s\n\n" "ASDF INSTALLED PLUGINS" "$(asdf plugin list --urls)"
+  printf "%s:\\n%s\\n\\n" "OS" "$(uname -a)"
+  printf "%s:\\n%s\\n\\n" "SHELL" "$($SHELL --version)"
+  printf "%s:\\n%s\\n\\n" "ASDF VERSION" "$(asdf_version)"
+  printf "%s:\\n%s\\n\\n" "ASDF ENVIRONMENT VARIABLES" "$(env | grep -E "ASDF_DIR|ASDF_DATA_DIR|ASDF_CONFIG_FILE|ASDF_DEFAULT_TOOL_VERSIONS_FILENAME")"
+  printf "%s:\\n%s\\n\\n" "ASDF INSTALLED PLUGINS" "$(asdf plugin list --urls)"
```
Stratus3D, line 8: ShellCheck recommends `\\n` for double quoted strings.

jthegedus (approved):
LGTM

jthegedus:
As a general point of discussion, what are your thoughts on:

1. first param to `printf` always being formatting only (no content)? EG:
	```diff
	-printf "asdf is self upgrading shims to new asdf exec ...\\n"
	+printf "%s\\n" "asdf is self upgrading shims to new asdf exec ..."
	```
2. re-using a formatting variable for many `printf` statements?
	```diff
	+format="%s %s\\n"
	-printf "%s %s\\n" "$asdf_cmd_dir/command-$1.bash" 2
	+printf "$format" "$asdf_cmd_dir/command-$1.bash" 2
	...
	-printf "%s %s\\n" "$asdf_cmd_dir/command-version.bash" 2
	+printf "$format" "$asdf_cmd_dir/command-version.bash" 2
	```

	Shellcheck does warn against this, but it's docs do mention explicitly ignoring for format reuse.

Stratus3D:
> 1. first param to printf always being formatting only (no content)? EG:

Either way is fine with me. Assuming the format string doesn't contain anything that would be interpreted as special characters by `printf` I think including text in it is ok. In the programming languages I use, I typically hardcode strings of text in the format string, and only use placeholders for values that are dynamic (ie variables).

> 2. re-using a formatting variable for many printf statements?

If a pattern is repeated in several places, it may make sense to encapsulate the print in a function. Something like:

```
function command_output() {
  printf "%s %s\\n" "$1" "$2"
}

command_output "$asdf_cmd_dir/command-version.bash" 2
```

Although I think in cases like this is it overkill (format string is just item, space, item, newline).

jthegedus:
I disagree on 2 as too often I see the function gets co-opted to do more work and often moved further out of context into another file which exacerbates the problem. Thanks for the food for thought :+1:

PR #817, opened 2020-10-16
Bash Strict Mode: Part 1 of 3

Stratus3D:
# Summary

Instead of implementing Bash strict mode in one pull request (as I tried to do with #520 and failed) I am going to do it in three parts:

* Don't allow unset variables (this PR)
* Exit on non-zero exit codes (including in pipes)
* Set a stricter IFS

This PR addresses part of #160

### DIFF
--- a/lib/commands/command-env.bash
+++ b/lib/commands/command-env.bash
@@ -16,7 +16,10 @@ shim_env_command() {
   fi
 
   shim_env() {
-    "$env_cmd" ${env_args[@]:-}
+    # Yes, the syntax for the expansion is ugly and confusing, but it is what we
+    # need to do to make it work with the nounset option. See
+    # https://stackoverflow.com/q/7577052/ for the details
+    "$env_cmd" ${env_args[@]+"${env_args[@]}"}
   }
 
   with_shim_executable "$shim_name" shim_env || exit $?
--- a/lib/commands/command-env.bash
+++ b/lib/commands/command-env.bash
@@ -1,8 +1,9 @@
 # -*- sh -*-
+set -o nounset
 
 shim_env_command() {
-  local shim_name="$1"
-  local env_cmd="${2}"
+  local shim_name="${1:-}"
+  local env_cmd="${2:-}"
   local env_args=("${@:3}")
 
   if [ -z "$shim_name" ]; then
@@ -15,7 +16,7 @@ shim_env_command() {
   fi
 
   shim_env() {
-    "$env_cmd" "${env_args[@]}"
+    "$env_cmd" "${env_args[@]:-}"
   }
 
   with_shim_executable "$shim_name" shim_env || exit $?
--- a/lib/commands/command-env.bash
+++ b/lib/commands/command-env.bash
@@ -16,7 +16,7 @@ shim_env_command() {
   fi
 
   shim_env() {
-    "$env_cmd" "${env_args[@]:-}"
+    "$env_cmd" ${env_args[@]:-}
   }
 
   with_shim_executable "$shim_name" shim_env || exit $?
--- a/lib/commands/command-reshim.bash
+++ b/lib/commands/command-reshim.bash
@@ -1,5 +1,7 @@
 # -*- sh -*-
 
+set -o nounset
+
 # shellcheck source=lib/commands/reshim.bash
 . "$(dirname "$ASDF_CMD_FILE")/reshim.bash"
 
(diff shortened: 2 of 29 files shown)


lib/commands/command-reshim.bash
```diff
@@ -25,9 +26,11 @@ reshim_command() {
 
   if [ "$full_version" != "" ]; then
     # generate for the whole package version
+    local full_version_name
+    full_version_name=$(basename "$full_version" | sed 's/ref\-/ref\:/')
     asdf_run_hook "pre_asdf_reshim_$plugin_name" "$full_version_name"
-    generate_shims_for_version "$plugin_name" "$full_version"
-    asdf_run_hook "post_asdf_reshim_$plugin_name" "$full_version"
+    generate_shims_for_version "$plugin_name" "$full_version_name"
+    asdf_run_hook "post_asdf_reshim_$plugin_name" "$full_version_name"
```
Stratus3D, line 33: This was actually a bug! These functions were invoked in a way that was inconsistent with the code in the `else` block below.

test/shim_env_command.bats
```diff
@@ -80,5 +80,5 @@ teardown() {
 
   # Should not contain duplicate colon
   run grep '::' <(echo "$path_line")
-  [ "$duplicate_colon" == "" ]
```
Stratus3D, line 83: Another bug. This variable was never set so this line wasn't actually testing anything.

test/utils.bats
```diff
@@ -119,7 +119,7 @@ teardown() {
 @test "parse_asdf_version_file should output path on project with spaces" {
   PROJECT_DIR="$PROJECT_DIR/outer space"
   mkdir -p "$PROJECT_DIR"
-  cd $outer
```
Stratus3D, line 122: Another test that wasn't testing what it was suppose to because of an unset variable.

jthegedus:
looking over this now :)

Stratus3D:
@jthegedus build is failing right now. I created this PR on my OSX machine but it looks like things aren't working on Ubuntu. I'm going pull these changes down on my Ubuntu machine and resume working through the remaining issues. I probably won't get to this until late next week though.

jthegedus:
No problem. This is a good start to enabling strict mode. I like this process has revealed bugs, proves the effort is valid :smile:

lib/commands/command-env.bash
```diff
@@ -15,7 +16,10 @@ shim_env_command() {
   fi
 
   shim_env() {
-    "$env_cmd" "${env_args[@]}"
+    # Yes, the syntax for the expansion is ugly and confusing, but it is what we
+    # need to do to make it work with the nounset option. See
+    # https://stackoverflow.com/q/7577052/ for the details
+    "$env_cmd" ${env_args[@]+"${env_args[@]}"}
```
jthegedus, line 22: In an effort to capture minimum versions we support, I would like capture the min Bash version that supports this in the docs if we could

jthegedus (approved):

jthegedus (changes requested):
Hey mate, can this be re-based? If I recall it was pretty close to done.

Stratus3D:
> Hey mate, can this be re-based? If I recall it was pretty close to done.

Yes, this is next on my agenda. I will rebase and try to fix the failing tests on Ubuntu tomorrow.

Stratus3D:
I just rebased this PR onto latest master. I'm hoping to finish this PR update and get it merged soon.

jthegedus:
Tag me when you wish to have a review.

jthegedus:
I am not sure if we can automate some of these fixes with Shellcheck/Shfmt, but perhaps these proposed changes can be enforced (and potentially auto-fixed) by expanding `scripts/stylecheck.py`?

jthegedus:
@Stratus3D should we close this?

hyperupcall:
We can definitely add some stuff to `checkstyle.py` to prepare things for `set -u`, and I feel reluctant to say this to a 3-year old PR adding this stuff, but in my opinion, `errunset` is more trouble than it's worth, especially with ShellCheck.

Unfortunately, I can't think of any way to add the latter two to `checkstyle.py`, but maybe there is a way.

jthegedus:
@hyperupcall I'm very happy with the level of code quality improvements you have brought to the project over the past year. I don't think this is worth tackling (and assumed Stratus would agree), so thought I would just ask before closing it.

I will just close it though as I think we're all better off spending our time on other issues than this in the short term. We can always dig it back up should we want.

Stratus3D:
Thanks @jthegedus. I don't have plans to tackle this anymore. I'm focusing my effort on the Rust rewrite.

PR #818, opened 2020-10-16
Update no plugin installed error message

aaronjensen:
# Summary

Fixes: Updates error message when no plugins are installed

## Other Information

The previous error message was overly informal

### DIFF
--- a/lib/commands/command-list.bash
+++ b/lib/commands/command-list.bash
@@ -14,7 +14,7 @@ list_command() {
         display_installed_versions "$plugin_name"
       done
     else
-      printf "%s\\n" 'Oohes nooes ~! No plugins installed'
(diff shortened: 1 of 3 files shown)


jthegedus (approved):
While informal, some character shouldn't hurt, no?

aaronjensen:
@jthegedus it made me waste time trying to figure out if I hit some sort of debug message, for one. 

There are other reasons it can do harm but I don’t feel strongly enough about it to wade into it on someone else’s project. If the maintainer wants to close this out because they value what you described as “character” then I won’t mind.

jthegedus:
If it caused you to waste time, then that is not ideal.

aaronjensen:
Thank you for the merge, I appreciate it.

PR #833, opened 2020-12-02
Fix .github/workflows/workflow.yml

smorimoto:


smorimoto:
The CI failed because we used Actions-specific commands that were deprecated.

jthegedus (approved):
LGTM, cheers

PR #839, opened 2020-12-18
docs: github dicsussions link

jthegedus:
# Summary

Adds links to GitHub Discussions in the *Community & Questions* section of README and Sidepane of docs site


Stratus3D (approved):

Stratus3D:
One thing to note: StackOverflow questions often rank very well in Google search results. I've not yet seen Github discussions ranked highly by Google, but this may be because they are new.

jthegedus:
I have had the experience too. I think GH Discussions is much nicer to highlight the discussion whereas SO that only occurrs after a proposed solution, thus is less collaborative in the solution design

PR #843, opened 2020-12-28
docs: improve zsh completion directions for macOS,ZSH,Homebrew

N1H1L0:
# Summary

Changed install directions as the ohmyzsh asdf plugin has been updated to support Homebrew and completion from ohmyzsh/ohmyzsh#6749.

### DIFF
--- a/docs/core-manage-asdf.md
+++ b/docs/core-manage-asdf.md
@@ -239,7 +239,7 @@ echo -e "\n. $(brew --prefix asdf)/asdf.sh" >> ~/.zshrc
 ```
 **OR** use a ZSH Framework plugin like [asdf for oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/asdf) which will source this script and setup completions.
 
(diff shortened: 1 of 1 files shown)


jthegedus:
I don't consider the OMZSH plugin up to date with `asdf`. If we're going to update the language here, I suggest we also warn users that the plugin may be out of date as per the other setup directions:

>if you are using a ZSH Framework the associated asdf plugin may need to be updated to use the new ZSH completions properly via fpath. The Oh-My-ZSH asdf plugin is yet to be updated, see https://github.com/ohmyzsh/ohmyzsh/pull/8837.

jthegedus (changes requested):
I don't consider the OMZSH plugin up to date with `asdf`. If we're going to update the language here, I suggest we also warn users that the plugin may be out of date as per the other setup directions:

>if you are using a ZSH Framework the associated asdf plugin may need to be updated to use the new ZSH completions properly via fpath. The Oh-My-ZSH asdf plugin is yet to be updated, see https://github.com/ohmyzsh/ohmyzsh/pull/8837.

jthegedus (approved):
Thanks :)

Issue #581, opened 2019-10-17
error using git-flow


yofriadi:
when releasing using git flow

```
➜ git-flow release finish 0.3.0
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
/Users/yofri/.asdf/lib/utils.sh: line 606: syntax error near unexpected token `('
/Users/yofri/.asdf/lib/utils.sh: line 606: `    grep -f <(get_shim_versions) <(preset_versions) | head -n 1 | xargs echo'
/Users/yofri/.asdf/lib/commands/shim-exec.sh: line 29: with_shim_executable: command not found
Could not check out master.
```

### Environment

**OS**: macOS catalina

**asdf version**: v0.7.4

Stratus3D:
@yofriadi you'll need to give us more information about what is going on. How did you install asdf? How did you install git-flow? What asdf plugins are you using?

yofriadi:
@Stratus3D yea, I used git-flow for my day to day development software, I did install asdf using either homebrew or curl (sorry I forget), I did install git-flow using homebrew and the plugin I using is go, nodejs and lua.

Stratus3D:
@yofriadi I'm not really sure I understand what is going on. Something on your system seems to be invoking something managed by asdf, but I don't know what. Do you know what that git-flow command actually does? Does it run tests or build your project?

yofriadi:
@Stratus3D git-flow is only for managing git development process, like committing, branching, release, hotfix, https://github.com/nvie/gitflow. in my case it broke in release and hotfix process, could you provide me on where to debug this thing?

Stratus3D:
I don't understand what the issue is. If you installed git-flow via brew then asdf should not interfere with it. Maybe open an issue with git-flow?

yofriadi:
ok, will check on it

joelsouza:
@yofriadi @Stratus3D 

I have the same issue here with the combination of OSX, zsh, git flow and asdf

When running `git flow feature start foo` I get:

```
~/.asdf/lib/utils.sh: line 606: `    grep -f <(get_shim_versions) <(preset_versions) | head -n 1 | xargs echo'
``` 

For some reason that I don't realize, just removing those parenthesis from this line it will work.

There is no test for `with_shim_executable` method so I don't know if removing it will introduce new bugs.

joelsouza:
@yofriadi @Stratus3D 

I have the same issue here with the combination OSX, zsh, git flow and asdf

[...]

PR #851, opened 2021-01-14
fix: shims break when POSIXLY_CORRECT=1

goodtouch:
# Summary

Process substitution isn't specified by POSIX and makes shims break when called by a script that `export POSIXLY_CORRECT=1` (like gitflow [here](https://github.com/nvie/gitflow/blob/develop/git-flow#L78)).

Here we replace `grep -f <(cmd1) <(cmd2)` with `cmd2 | grep -F "$(cmd1)"` so that we can provide a string instead of a file descriptor and get the same result as before.

(Note: We also check if the result of `cmd1` is empty before running `grep` to get the same behavior as with `-f`. This also prevent crashes if null results are piped to `xargs` and fits nicely with the `with_shim_executable` function that already check the result of `$selected_version` [here](https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash#L729) before going on).

Fixes: #581

## Other Information

The test case is from @tejanium in #581.

From the man page of `grep`:

```
-f FILE, --file=FILE
      Obtain  patterns  from  FILE, one per line.  The empty file contains zero patterns,
      and therefore matches nothing.  (-f is specified by POSIX.)

-F, --fixed-strings
      Interpret  PATTERN  as a list of fixed strings, separated by newlines, any of which
      is to be matched.  (-F is specified by POSIX.)
```

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -663,7 +663,10 @@ preset_versions() {
 
 select_from_preset_version() {
   shim_name=$1
-  grep -f <(get_shim_versions "$shim_name") <(preset_versions "$shim_name") | head -n 1 | xargs -IVERSION printf "%s\\n" VERSION
+  shim_versions=$(get_shim_versions "$shim_name")
+  if [ -n "$shim_versions" ]; then
+    preset_versions "$shim_name" | grep -F "$shim_versions" | head -n 1 | xargs -IVERSION printf "%s\\n" VERSION
+  fi
 }
 
 select_version() {
--- a/test/shim_exec.bats
+++ b/test/shim_exec.bats
@@ -436,3 +436,15 @@ EOM
   [ "$status" -eq 1 ]
 }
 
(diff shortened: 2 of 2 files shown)


Stratus3D:
Thanks for the detailed PR @goodtouch ! This looks good! We need to find other occurrences of process substitution in our code and replace them as well. If you have any suggestions on how best to go about that, please let me know.

Stratus3D (approved):

jthegedus:
@Stratus3D are we able to capture this with the banned command list to ensure it doesn't resurface?

Stratus3D:
@jthegedus my first thought was grepping for `<(` might be enough to find all process substitutions, but I haven't thought about it enough to know if that would result in false negatives or positives.

jthegedus:
Fair enough. Something we should consider trying out. Thanks @goodtouch

Stratus3D:
I've got a branch locally with the changes. Will create a PR as soon as I can.

PR #860, opened 2021-02-03
docs: fish symlink completions rather than copying

tekacs:
# Summary

At the moment, the fish installation instructions copy the completions to the ~/.config/fish/completions directory.

Having tried this locally, a symlink works just as well.

Could it be worth symlinking to bring along any future updates to the completions file?


Stratus3D:
@danhper @jthegedus any thoughts this? It seems like a good improvement to me.

jthegedus (approved):

jthegedus:
If we get a school of fish users having issues we can always revert.

PR #863, opened 2021-02-05
Ban process substitution from asdf codebase

Stratus3D:
# Summary

Process substitution is not POSIX compliant and has caused our users issues in the past. It's not to hard to work around not having it, and doing so makes asdf compatible in more environments.

Fixes: #851

### DIFF
--- a/lib/commands/command-plugin-update.bash
+++ b/lib/commands/command-plugin-update.bash
@@ -10,11 +10,12 @@ plugin_update_command() {
   local gitref="${2:-master}"
   if [ "$plugin_name" = "--all" ]; then
     if [ -d "$(asdf_data_dir)"/plugins ]; then
+      plugins=(find "$(asdf_data_dir)"/plugins -type d -mindepth 1 -maxdepth 1)
       while IFS= read -r -d '' dir; do
         local plugin_name
         plugin_name=$(basename "$dir")
         update_plugin "$plugin_name" "$dir" "$gitref" &
-      done < <(find "$(asdf_data_dir)"/plugins -type d -mindepth 1 -maxdepth 1)
+      done <<< "$plugins"
       wait
     fi
   else
--- a/lib/commands/command-reshim.bash
+++ b/lib/commands/command-reshim.bash
@@ -119,7 +119,19 @@ remove_obsolete_shims() {
   exec_names=$(plugin_executables "$plugin_name" "$full_version" | xargs -IX basename X | sort)
 
   local obsolete_shims
-  obsolete_shims=$(comm -23 <(printf "%s\\n" "$shims") <(printf "%s\\n" "$exec_names"))
+  local formatted_shims
+  local formatted_exec_names
+
+  # comm only takes to files, so we write this data to temp files so we can
+  # pass it to comm.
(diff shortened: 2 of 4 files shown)


lib/commands/command-plugin-update.bash
```diff
@@ -10,11 +10,12 @@ plugin_update_command() {
   local gitref="${2:-master}"
   if [ "$plugin_name" = "--all" ]; then
     if [ -d "$(asdf_data_dir)"/plugins ]; then
+      plugins=(find "$(asdf_data_dir)"/plugins -type d -mindepth 1 -maxdepth 1)
       while IFS= read -r -d '' dir; do
         local plugin_name
         plugin_name=$(basename "$dir")
         update_plugin "$plugin_name" "$dir" "$gitref" &
-      done < <(find "$(asdf_data_dir)"/plugins -type d -mindepth 1 -maxdepth 1)
+      done <<< "$plugins"
```
jthegedus, line 18: This `find` was introduced because the `for dir in plugins/*` didn't handle the case where the `plugins/` dir was not yet created. Is `find` more portable than `for..in` and would it be worth us either replacing all similar usages to the more portable version? (and in this specific case just creating the empty dir to address the root issue)

Stratus3D: I think `find` is more portable in general. There my be newer `find` flags that aren't widely available, but the existing flags are sufficient for our uses cases so I don't think it would matter.

jthegedus:
ShellCheck doesn't like the `<<<` and I am not sure the proposed solution is POSIX

```
In lib/commands/command-plugin-update.bash line 17:
      done <<<"$plugins"
               ^------^ SC2128: Expanding an array without an index only gives the first element.

For more information:
  shellcheck.net/wiki/SC2128 -- Expanding an array without an ind...
Error: Process completed with exit code 1.
```

Shellcheck error code page: https://github.com/koalaman/shellcheck/wiki/SC2128

>Problematic code:
```
myarray=(foo bar)
for f in $myarray
do
  cat "$f"
done
```

>Correct code:
```
myarray=(foo bar)
for f in "${myarray[@]}"
do
  cat "$f"
done
```

I have seen suggestions that piping in to while with find's `-print0` flag is best, eg: `find . -name '*.txt' -print0 | while ...`


Thoughts?

jthegedus (approved):
this LGTM

jthegedus (changes requested):
Our tests didn't capture that the `-print0` didn't resolve any dirs in 26d2c64477a1faabedd9a5f97aa7da706988cd72 so we should write more or check the robustness of our tests.

Stratus3D:
Sorry @jthegedus, this got away from me. Looking at your code changes now.

Stratus3D:
More tests are always welcome @jthegedus. It's hard to get folks to contribute sufficient tests for new features they contribute, and in the past I've written tests for things that were lacking them.

I think I did contribute those tests for the plugin update command that didn't catch the issue you pointed out above. I'll see if I can improve them further. Is it ok if I merge this now?

jthegedus:
I wrote the latest changes to those tests, so happy to write the test update (s).

Might track them in another issue so we can close this out and merge before further changes to `master`

Stratus3D:
I created #947.

PR #871, opened 2021-02-19
Support alternate location for .zshrc

brandondrew:
Z shell supports placing your zshrc file in an alternate location, such as inside `.config` (which makes it far more conveninent to store your dot files in a Git repo, or even just sync them across machines with other folder-oriented synchronization tools).  

The current command assumes the default location of putting the rc file directly in the home directory.  But it is easy enough to support all alternate locations as well.  This version defaults to `~` if the variable `ZDOTDIR` is not defined.

I'm only including the case I've tested in this PR, but if you want me to do the same thing for other shells and other cases with Zsh, I'm open to doing that too.

# Summary

## Other Information

information here.

### DIFF
--- a/docs/core-manage-asdf.md
+++ b/docs/core-manage-asdf.md
@@ -235,7 +235,7 @@ echo -e "\nsource "(brew --prefix asdf)"/asdf.fish" >> ~/.config/fish/config.fis
 Add `asdf.sh` to your `~/.zshrc` with:
 
 ```shell
-echo -e "\n. $(brew --prefix asdf)/asdf.sh" >> ~/.zshrc
+echo -e "\n. $(brew --prefix asdf)/asdf.sh" >> ${ZDOTDIR:-~}/.zshrc
 ```
 **OR** use a ZSH Framework plugin like [asdf for oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/asdf) which will source this script and setup completions.
 


Stratus3D:
Thanks for the PR! This seems like a good approach to me. @asdf-vm/core any thoughts on this? It does add a little extra complexity to the command, but makes the command as it is work on more systems.

Stratus3D (approved):

Stratus3D:
> I'm only including the case I've tested in this PR, but if you want me to do the same thing for other shells and other cases with Zsh, I'm open to doing that too.

@asdf-vm/core what do you think about this?

@brandondrew what other cases would there be for Zsh?

jthegedus:
I like this change. Only concerns are:
- is there an existing or standard naming convention for this `ZDOTDIR`?
- this change may need to be included in more places in the docs
- should we document in our [config/env-vars](https://asdf-vm.com/#/core-configuration?id=environment-variables) section non-asdf specific env var usage like this?

Stratus3D:
> I like this change. Only concerns are:
> 
>   * is there an existing or standard naming convention for this `ZDOTDIR`?
>
Yes, see http://zsh.sourceforge.net/Intro/intro_3.html#IDX11
>
>   * this change may need to be included in more places in the docs
>
You're right. I haven't searched the docs for other places we may want to use this variable.
>
>   * should we document in our [config/env-vars](https://asdf-vm.com/#/core-configuration?id=environment-variables) section non-asdf specific env var usage like this?

I don't think so. Being that this is an environment variable for Zsh I don't think we should document it. We only use it in the docs for convenience.

brandondrew:
> @brandondrew what other cases would there be for Zsh?

Sorry, I was probably a little vague.  I was referring to other installation options in that document that involved Zsh.  

Would you like me to look for other places that the same change could be made?

Stratus3D:
> Would you like me to look for other places that the same change could be made?

Yes please.

Issue #862, opened 2021-02-05
Website Improvements


Stratus3D:
**I want to prefix this by saying that having a website for asdf is wonderful and I'm thankful to everyone who has dedicated time to making it what it is today. The website is a great resource and a good place for our asdf documentation to grow and evolve.**

Since the website was started I've kept notes on the things I came across that caused by trouble, or that others complained about. Below are the things I believe we should change to make the website more helpful and accessible.

## Content Improvements

* possible version values are covered under the `.tool-versions` section, but the section on managing versions doesn't ever show the possible version values.

## Template Improvements

* The website should work without JavaScript. Right now with JavaScript disabled all you get a is a blank white page.
* Show top of content without having to scroll down - annoying to always have to scroll to see anything.
* Make installation page link more promenent.
* Define clear top level links for install, use, contribute, contact, etc...

## Websites I like:

Adding these here so we can have a few concrete examples to compare asdf-vm.com with:

* https://pa11y.org/
  * simple layout. Simple navigation. No animations or hidden elements. Narrow page.
* https://swaywm.org/
  * text content is front and center. No large header or graphical elements.

Stratus3D:
https://news.ycombinator.com/item?id=26019010

jthegedus:
I have been thinking on this topic for quite some time too and agree we should have a completely JS-free static site for the reasons you mention and link, amongst others.

Docsify has served us well to go from nothing to something, but perhaps not something to all-inclusive.

The next question I guess is, are we looking to write our own raw HTML or use a tool? Both could be hard to maintain for differing reasons.

If we're going down the tool route, I would recommend we try https://www.11ty.dev/ (what do others suggest?) https://docusaurus.io/ and the like seem like a bit much.

Features we currently use but would lose without JS:

- content always updated without CI becuase client fetches (simply remedied)
- search
- select dropdowns on [install page](https://asdf-vm.com/#/core-manage-asdf)
	- the current implementation also detects the user's platform and autoselects accordingly

I think losing the dropdowns is a bit of a bummer as it solves a real content issue. The next best alternative is generating every permutation of the page at build time. we could of course load everything and then if/when JS loads it filters and provides the dropdownds. What are others thoughts on this?

As for default styling which Docsify gives us, I would suggest we use something simple like [water.css](https://watercss.kognise.dev/) (single rel stylesheet in `head`)

---

[...]

PR #876, opened 2021-02-26
docs: button for adding new translations

sanjarcode:
The "language translation" section is somewhat diffcult to find, this addition makes it easily accessible even from the homepage.

Added a single line.


Stratus3D:
@jthegedus what do you think about this?

jthegedus:
This is a good idea. I think we should tackle #862 before we start accepting major translations efforts though. Will merge this now as a reminder to add it back once we update the site

jthegedus (approved):

Issue #734, opened 2020-06-04
Automatically warn/install versions on enter directory when mismatch with current


tomsquest:
**Is your feature request related to a problem?**

I am frustrated when I enter a directory containing a tool-version that is not installed and thus not used by asdf, then I don't use the expected tools. I am not warned about the mismatch between required tool (defined in `.tool-versions` and the one in `asdf current`).

Given a directory with a `.tool-versions`
When the user enters that directory or one of its children
Then asdf **warns** and/ask for **installation** of the tools versions defined in `.tool-versions`

**Describe the solution you'd like**

I would like that asdf detects that the current tools are not the one defined in the `.tool-versions` and install them automatically (or ask for install).

**Describe similar asdf features and why they are not sufficient**

asdf correctly **use** the tools but it did not warn/install them automatically

**Describe workarounds you've considered**

A shell plugin could detect that the current tools and the one in .tool-versions (in the hierarchy) do not match.
This plugin can warn the user that something is wrong and ask if it can install the tools.

Nvm (Node Version Manager) has [some instructions to do it](https://github.com/nvm-sh/nvm#calling-nvm-use-automatically-in-a-directory-with-a-nvmrc-file).
This snippet will use or install (depending of the versions of node installed and declared).

**Additional context**

```
$ asdf version
v0.7.8-4a3e3d6
```

Stratus3D:
We've avoided doing this because as you can see it requires aliasing `cd` to some other function in Bash which is 1). something that is frowned upon (for good reason) and 2). can have a significant affect on performance as it will run asdf code on *all* directory changes.

I'm not opposed to having this as a possible feature for some users to setup on their own, but I don't think this should be something we force on users.

Having an asdf command that performs the necessary check would allow us users to run it manually if they wished (without the `cd` hook) or configure their shell to run it on directory change (if they want the feature you are describing).

What do you think?

tomsquest:
No problem.
If there is a performance impact and moreover it does not bother users, then definitively the feature should be optional.

For the "implementation", this could be documented as NVM does in [this section of the Readme](https://github.com/nvm-sh/nvm#bash).

As I am on holidays and don't like having issues openend, I am closing this issue until I propose a PR.

Stratus3D:
I re-opened this issue so others can more easily find it. I think this would be a good feature if it is optional. A PR is welcome but no hurry.

ghostsquad:
[...]

PR #877, opened 2021-02-27
Add check command

jameslai:
# Summary

The check command returns exit code 0 or 1, depending if the version requirements for the current directory are met. Best used in conjunction with other tooling for automated installation, etc.

I have tested this in conjunction with a direnv containing:

```
if has asdf; then
  asdf check || asdf install
fi
```

...and it works like magic!

Fixes: #734 

## Other Information

This change goes out of its way to take advantage of another asdf function, `current` by parsing its output (rather than re-implementing roughly 80% of that function). Tests provided to ensure nothing breaks if ever the string parsing fails.

### DIFF
--- a/lib/commands/command-check.bash
+++ b/lib/commands/command-check.bash
@@ -1,0 +1,18 @@
+# -*- sh -*-
+
+# shellcheck disable=SC2059
+check_command() {
+  local exit_status=0
+  local not_installed_description
+  not_installed_description="Not installed"
+
+  while read -r line; do
+    if [[ $line == *"$not_installed_description"* ]]; then
+      exit_status=$((exit_status + 1))
+    fi
+  done <<<"$(asdf current 2>&1)"
+
+  exit "$exit_status"
+}
+
+check_command
--- a/test/check_command.bats
+++ b/test/check_command.bats
@@ -1,0 +1,39 @@
+#!/usr/bin/env bats
+
+load test_helpers
+
+setup() {
+  setup_asdf_dir
+  install_dummy_plugin
+  install_dummy_version "1.1.0"
+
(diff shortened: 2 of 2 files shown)


Stratus3D:
Hi @jameslai, thanks for the PR! The code looks good, but I don't think this functionality merits an entirely new command. Our goal has been to provide a simple tool with an interface makes the common tasks easy, and the hard task possible with a little extra code.

There are a couple options here:

* Add an option to the existing `asdf current` command to exit if any tool specified is not installed. e.g. a `-c/--check` flag. This would eliminate the need for this new command as users could just run `asdf current --check || asdf install` instead.
* Do nothing at all. This code is simple, and the same thing can be achieved outside of asdf in only one line of code:
  ```
  (asdf current 2>&1 | grep "Not installed") && asdf install
  ```

  Basically this code takes all the output of `asdf current` (both STDOUT and STDIN `2>&1`) and looks for the string "Not installed" if found the `grep` command succeeds and triggers `asdf install` to be run. If the string is not found the command fails and `asdf install` is not invoked.

I personally don't see the benefit of making any changes to asdf for this, since it can be done in a single line of shell code.

jthegedus:
This is yet another example we could document in a "recipies" section of the docs. Something I want to tackle once we migrate our docs site.

Stratus3D:
@jthegedus I actually have been keeping track of all the recipes that I have shared so far and my plan was to eventually put them into a free ebook. But, that may not happen, and this may be useful to you as well so I'm going to go ahead and share the gist I have - https://gist.github.com/Stratus3D/cf63bcb79695433075d3ce872a584e83. If you want me to convert it to a Git repo so you can send PRs let me know.

Issue #880, opened 2021-03-03
list-all does not allow for plugin to error


laidbackware:
**Describe the bug**

When calling list all, there is no way for the plugin to error and pass the message back to the user. The versions `variable` captures all output including the return code, which is ignored. Use cases could be github rate limiting, proxy certificate trust, network access.

Also due to the failure not being detected CI tests pass, even with a broken test.

**To Reproduce**

1. Intentionally break a break an installed github based plugin such as kubectx by changing the following in the list-all file within $HOME/plugins/kubectx/bin
```
# Existing line
releases_path=https://api.github.com/repos/ahmetb/kubectx/releases
# Change to
releases_path=https://api.github.com/repos/ahmetb/I_AM_BROKEN/releases
```
2. asdf list all kubectx

**Actual behavior**

You see an empty output and running `echo $?` should return 0. Looking like there are no versions.

**Expected behavior**

list-all should be able to output an error if it is generated by the plugin. Here is an example of a work in progress plugin which detects an error, outputs the error and exits with code 1.
https://github.com/laidbackware/asdf-github-tools/blob/edfa89e5f7f6c5f64895fcdb19118eb5591468e2/bin/list-all#L64

### Environment

```shell
OS:
Linux derek 5.8.0-41-generic #46-Ubuntu SMP Mon Jan 18 16:48:44 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

ASDF VERSION:
v0.8.0-c6145d0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/home/matt/.asdf

ASDF INSTALLED PLUGINS:
antctl                       https://github.com/laidbackware/asdf-github-tools
bosh                         https://github.com/laidbackware/asdf-github-tools
credhub                      https://github.com/laidbackware/asdf-github-tools
fly                          https://github.com/laidbackware/asdf-github-tools
govc                         https://github.com/laidbackware/asdf-github-tools
kubectl                      https://github.com/Banno/asdf-kubectl.git
kubectx                      https://gitlab.com/wt0f/asdf-kubectx.git
[...]

PR #881, opened 2021-03-03
fix: insert error handling in list-all & download plugin scripts

laidbackware:
# Summary

This fix will check if a plugin does not exit 0 when calling list-all or download. In the case of an error it will print all stdout and stderr returned by the plugin and then the list all script will exit 1.

Without this fix, it is not possible to test a plugin using CI and get reliable results, as errors are not handled.

Fixes: This fixes https://github.com/asdf-vm/asdf/issues/880 and https://github.com/asdf-vm/asdf/issues/886

## Other Information

This has been tested on Linux and MacOS to be working as expected.


Stratus3D:
Thanks for the PR! I think there are a couple things we need to do before merging this:

* Write tests for list-all failures, and assert that the command works as expected
* When plugin list-all fails, print a line before the plugin output with something along the lines of `Plugin <plug name>'s list-all callback script failed with output:`

Thanks for the PR! This is a good improvement.

laidbackware:
Refactored to include tests - which included fixed 2 existing tests.
I broke some affected string tests out to separate lines to aid readability and prevent long lines.
Added check to install script to allow for plugin download to cleanly exit.
The returncode is collected outside the subshell with no default set prior, as if the download/install block doesn't run the return code of the if command will be captured. The script will then tidy up if either download or install do not exist clearnly.

lib/commands/command-list-all.bash
```diff
@@ -8,7 +8,15 @@ list_all_command() {
...
 
   local versions
-  versions=$(bash "${plugin_path}/bin/list-all")
+  # Capture return code to allow error handling
+  return_code=0 && versions=$(bash "${plugin_path}/bin/list-all" 2>&1) || return_code=$?
+
+  if [[ $return_code -ne 0 ]]; then
+    # Printing all output to allow plugin to handle error formatting
+    printf "Plugin %s's list-all callback script failed with output:\\n" "${plugin_name}"
+    printf "%s\\n" "${versions}"
```
Stratus3D, line 17: I think these prints should go to STDERR instead of STDOUT. Generally only good data should be printed to STDOUT as it is often consumed by other programs.

test/list_command.bats
```diff
@@ -26,7 +28,10 @@ teardown() {
   run asdf install dummy 1.0.0
   run asdf install tummy 2.0.0
   run asdf list
-  [ "$(echo -e "dummy\n  1.0.0\nmummy\n  No versions installed\ntummy\n  2.0.0")" == "$output" ]
+  [[ "$output" == "$(echo -e "dummy\n  1.0.0")"* ]]
+  [[ "$output" == *"$(echo -e "dummy-broken\n  No versions installed")"* ]]
+  [[ "$output" == *"$(echo -e "mummy\n  No versions installed")"* ]]
+  [[ "$output" == *"$(echo -e "tummy\n  2.0.0")" ]]
```
Stratus3D, line 34: Why does this test code need to change?

Stratus3D:
@laidbackware changes look good! I did leave two comments, one about printing failure output to STDERR, and the other is a question about the tests. Let me know if you will address these comments or if you would like me to make the STDERR change before merging.

test/list_command.bats
laidbackware: Adding the broken plugin caused this test to fail, as the test was expecting an explicit output, which had changed due as the `asdf list` command now includes the broken plugin. I also chose to split over multiple lines to aid readabiliy.

laidbackware:
@Stratus3D I have also been thinking about another edge case which I'd be interested on your thoughts on before implementing?

This change is currently written to redirect all STDOUT and STDERR data to a single variable when calling list-all. This is fine if the plugin generates an error, but may cause issues if a plugin uses a tool which happens to write to STDERR under normal operation as some linux tools do.
https://github.com/asdf-vm/asdf/blob/e2c4f7c631e4ea242aad46623c936419001da99c/lib/commands/command-list-all.bash#L12
In this case it would skew the output of the asdf command.

The only clean solution I can think of to split the output is to acutally call the list-all script twice. Once just capturing STDOUT which is printed if no errrors occur, but if there is an error it is run a second time to capture STDOUT and STDERR.

What do you think?

lib/commands/command-list-all.bash
laidbackware: I've pushed the requested change and all looks good.

jthegedus (approved):
LGTM. @Stratus3D ?

laidbackware:
@Stratus3D @jthegedus I think I've solved the issue I mentioned previously without calling the plugin twice. I was concerned that if there were network timeouts, calling the plugin twice could be worse for the user. 
Now command-list-all.bash contains a function which splits stdout and stderr into separate variables. When the plugin list-all script exits with code 0, then stderr is ignored and when it exits with a non-zero code stderr then stdout is printed. I've also added tests to ensure stderr is ignored when sucessful and printed when unsuccessful.
All CI tests look good on my end.

jthegedus:
Thanks very much for contributing this @laidbackware Really appreciate it :smile:

Stratus3D:
Sorry @laidbackware for never responding, looks great :)

PR #887, opened 2021-03-10
Use source instead of dot alias

mijailr:
# Summary

Some users not familiarized with bash or the shell at all are having issues following the installation process, specifically with the dot builtin `source` .

Using source it's more clear they need to copy the whole line.

```diff
- . $HOME/.asdf/asdf.sh
+ source $HOME/.asdf/asdf.sh
```

### DIFF
--- a/docs/core-manage-asdf.md
+++ b/docs/core-manage-asdf.md
@@ -101,13 +101,13 @@ brew install asdf --HEAD
 Add the following to `~/.bashrc`:
 
 ```shell
-. $HOME/.asdf/asdf.sh
+source $HOME/.asdf/asdf.sh
 ```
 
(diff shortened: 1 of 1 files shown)


jthegedus:
Thanks for raising, perhaps we should explain this in the docs.

The reason for using `.` over `source` is that `.` is the POSIX standard for this syntax and therefore more exportable to other shells. Where possible we're preferring PSOSIX compatible code.

Stratus3D:
@jthegedus would it make sense to add `source` to our banned commands list?

jthegedus:
I think we should z then it's at least documented somewhere

mijailr:
Oh I see thanks for the clarification.

PR #890, opened 2021-03-11
Add `source` to list of banned commands

Stratus3D:
Fixes: #887

### DIFF
--- a/lib/commands/command-list-all.bash
+++ b/lib/commands/command-list-all.bash
@@ -13,7 +13,7 @@ list_all_command() {
   # Capture return code to allow error handling
   std_out_file="$(mktemp "/tmp/asdf-command-list-all-${plugin_name}.stdout.XXXXXX")"
   std_err_file="$(mktemp "/tmp/asdf-command-list-all-${plugin_name}.stderr.XXXXXX")"
-  return_code=0 && $(bash ${plugin_path}/bin/list-all > "$std_out_file" 2> "$std_err_file") || return_code=$?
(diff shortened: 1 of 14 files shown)


jthegedus:
seems we are using `source` already

Stratus3D:
Looks like the tests are failing because we have one instance of process substitution present in the codebase now. I will fix that in this PR before merging.

jthegedus:
> Looks like the tests are failing because we have one instance of process substitution present in the codebase now. I will fix that in this PR before merging.

I started tracking that in #948 so mark that as completed aswell.

Fixes #948

jthegedus (approved):
Nice improvements :+1: LGTM

lib/commands/command-list-all.bash
```diff
@@ -4,48 +4,42 @@ list_all_command() {
...
-  local std_out
-  local std_err
+  local std_out_file
+  local std_err_file
+  local output
   plugin_path=$(get_plugin_path "$plugin_name")
   check_if_plugin_exists "$plugin_name"
 
   # Capture return code to allow error handling
-  return_code=0 && split_outputs std_out std_err "bash ${plugin_path}/bin/list-all" || return_code=$?
```
Stratus3D, line 13: Rather than relying on this `split_outputs` function that keeps output in memory I opted to just write STDOUT and STDERR to separate temp files. This makes the code here much simpler. The downside is we are relying on the file system. But I figured this is fine because we already use temp files in several other places in the codebase.

jthegedus: Fair enough choice :+1:

PR #897, opened 2021-03-19
asdf shell failing should point out relevant file to source. 

seivan:
# Summary
`asdf shell` will now point out what file to source.

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -61,7 +61,7 @@ asdf_cmd() {
   local ASDF_CMD_FILE args_offset
 
   if [ "shell" == "$1" ]; then
-    printf "Shell integration is not enabled. Please ensure you source asdf in your shell setup." >&2
+    echo "Shell integration is not enabled. Please ensure you source $(asdf_dir)/lib/asdf.sh/fish in your shell setup." >&2
     exit 1
   fi
 
--- a/bin/asdf
+++ b/bin/asdf
@@ -61,7 +61,7 @@ asdf_cmd() {
   local ASDF_CMD_FILE args_offset
 
   if [ "shell" == "$1" ]; then
-    printf "Shell integration is not enabled. Please ensure you source $(asdf_dir)/lib/asdf.sh/fish in your shell setup." >&2
+    printf "Shell integration is not enabled. Please ensure you source $(asdf_dir)/lib/asdf.{sh|fish} in your shell setup." >&2
     exit 1
   fi
 
--- a/bin/asdf
+++ b/bin/asdf
@@ -61,7 +61,7 @@ asdf_cmd() {
   local ASDF_CMD_FILE args_offset
 
   if [ "shell" == "$1" ]; then
-    printf "Shell integration is not enabled. Please ensure you source $(asdf_dir)/lib/asdf.{sh|fish} in your shell setup." >&2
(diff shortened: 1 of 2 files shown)


jthegedus:
I take it this is only really going to be hit by Homebrew installations as to get this far you'd have already set this up via our other installation methods?

seivan:
> I take it this is only really going to be hit by Homebrew installations as to get this far you'd have already set this up via our other installation methods?

Less then current situation, since I fixed both the installation and managed to add a caveat about it, less people should be hit. 


It isn’t exactly clear why you need to source it until you actually do, but by then you’re already forgotten the caveat. 

This should remind you.

jthegedus:
This should have a specific note about Homebrew then since only the users who setup the tool via Homebrew would see this.

No other setups would need to source the `lib/asdf.{sh|fish}`.

It seems to better support Homebrew these changes you're making are going to result in a completely different code path through `asdf` for these users, is that fair to say?

seivan:
> This should have a specific note about Homebrew then since only the users who setup the tool via Homebrew would see this.

How is this specific to Homebrew? This will show up if you run the binary. 


> 
> No other setups would need to source the `lib/asdf.{sh|fish}`.

```
git clone asdf; cd asdf
./bin/asdf
```

The error shows up. 

> It seems to better support Homebrew these changes you're making are going to result in a completely different code path through `asdf` for these users, is that fair to say?

I don't understand this sentence, how is either the error or the explanation related to Homebrew? I didn't add the error message, I explained how to fix it.

jthegedus:
Yes, this is a simple change to add directions.

I am wondering if the case in which this error shows is ever actually needed. `lib/asdf.sh` is sourced in `asdf.sh`, so how are users ending up in the state where they haven't sourced `asdf.sh` and need `lib/asdf.sh`

Perhaps I should raise a different issue to discuss.

jthegedus:
Point being, users shouldn't ever need to source `lib/asdf.sh` themselves. So the error perhaps means `asdf.sh` and this change is further erroneous.

seivan:
> Yes, this is a simple change to add directions.

Better than no directions. 

> I am wondering if the case in which this error shows is ever actually needed. `lib/asdf.sh` is sourced in `asdf.sh`, so how are users ending up in the state where they haven't sourced `asdf.sh` and need `lib/asdf.sh`

I don't source neither files. 

If you're running the binary that's under `/bin/` you will get this error. 
The error didn't explain how to resolve it, and now it does so you can go on with your day.

> Perhaps I should raise a different issue to discuss.

Maybe.

> Point being, users shouldn't ever need to source `lib/asdf.sh` themselves. So the error message perhaps means `asdf.sh` and this change is further erroneous.

The thing is, `asdf.sh` does *other* things as well. Now I can see the value for that where people might not have their paths properly set up, or the priority is skewed to using system executables, that's a valid point, but it's point not related to **this** particular error. 

**TL;DR** 
I **personally** don't source either files, but if I ever do want to run `$ asdf shell` I get reminded that I can source `lib/asdf.sh` to get the feature temporarily activated and move on.

Stratus3D (changes requested):
> Point being, users shouldn't ever need to source lib/asdf.sh themselves. So the error message perhaps means asdf.sh and this change is further erroneous.

I agree with this and I don't feel like this error message change is an improvement. I think we should either close this PR or change the error message to include a URL to relevant documentation.

PR #906, opened 2021-03-23
Update syntax with dash commands

Kosmas:
# Summary

## Other Information

information here.

### DIFF
--- a/docs/core-manage-plugins.md
+++ b/docs/core-manage-plugins.md
@@ -7,30 +7,30 @@ See [Creating Plugins](plugins-create) for the super-simple API for supporting m
 ## Add
 
 ```shell
-asdf plugin add <name>
-# asdf plugin add erlang
+asdf plugin-add <name>
+# asdf plugin-add erlang
 ```
 
 If the plugin you want to install is not part of the plugins repository, you
 can add it using its repository URL:
 
 ```shell
-asdf plugin add <name> <git-url>
-# asdf plugin add elm https://github.com/vic/asdf-elm
+asdf plugin-add <name> <git-url>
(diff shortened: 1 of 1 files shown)


jthegedus (changes requested):
Hi @Kosmas are you also able to include changes to the respective completions files? (or at least the shell you use?)

seivan:
My two cents here, is we should get rid of the dash commands all together, if not remove them from docs and `$ usage`
I find it very confusing and clutters the interface.

As I understand it's there as a "bonus" because of how the commands are structured

jthegedus:
The `-` syntax was the original, the spacing is the new syntax. I agree we should just have the spacing syntax, but consider which workflows you may break and if it is worth it.

As a starting point, we should use the same spacing syntax in all docs and help and completions.

>And oh-my-zsh plugin should be nuked from orbit.
>If you got zsh installed via homebrew it will automatically add asdf completions to your fpath.

I think the docs do a well enough job of identifying this information.

Kosmas:
Hi @jthegedus 

I haven't used asdf for some time, and when I updated it to use it again, I had the option of using the dash command, which at that point I thought it was the new syntax.
That's why I did this PR.
But since the one with the space is the new one, then the documentation is fine and you should close this one here.
By the way I'm using bash in Linux

```
kosmas:$ asdf version
v0.8.0-c6145d0
kosmas:$ asdf 
current          help             list             local            plugin-list      plugin-remove    reshim           update           which            
global           install          list-all         plugin-add       plugin-list-all  plugin-update    uninstall        where   
```

jthegedus:
Thanks for your input @Kosmas Will close and try and remove other references to the `-` commands throughout the rest of the docs

Issue #868, opened 2021-02-11
[Request] Way to revert to the system version?


Honeypot95:
My workflow is usually changing the version of a runtime to try out some stuff, after which I want to revert to the system version. I did not find a way in the documentation or from toying around with `asdf` itself. For now, I just use the `shell` option, after which I restart the shell.

`pyenv` seems to list a "system" when listing versions, which provides a way to revert to the system version. Is such a thing in the cards for `asdf`?

cho-m:
Have you tried using the `system` keyword?
There is some documentation on it in [core-configuration.md](https://github.com/asdf-vm/asdf/blob/master/docs/core-configuration.md#tool-versions):

> - `system` - this keyword causes asdf to passthrough to the version of the tool on the system that is not managed by asdf.

Based on a quick test, it works for me:
```zsh
❯ asdf shell python 3.9.2

❯ python --version
Python 3.9.2

❯ asdf shell python system

❯ python --version
Python 2.7.16

❯ /usr/bin/python --version
Python 2.7.16
```

Currently, ASDF doesn't seem to show it as part of list of available versions.
This could be either be a design decision or a bug.
```zsh
❯ asdf list python
  3.9.2

❯
```

In terms of shell autocompletion, it looks like only fish-shell has been updated to support tab-completing `system` as option:
- Fish Example:
  ```fish
  ❯ asdf shell python <tab>
  3.9.2  system
  
  ❯ asdf shell ruby <tab>
  3.0.0  system
  ```
- Corresponding code:
  https://github.com/asdf-vm/asdf/blob/0f75c2162944f50120906ba61faa62783cecfcb5/completions/asdf.fish#L131

jthegedus:
Thanks for raising this @Honeypot95 and for the thorough answer @cho-m 

Whether `asdf list <name>` should list `system` is open to interpretation, though I would say it should as it increases the usability a lot.

If someone wants to solve rectify these issues, I would say the requirements are:

1. `asdf list <name>` should append `system` to the output
2. all shell completions should be consistent in their tab completion for listing versions. I suspect if 1. is solved, then the Fish completions referenced by @cho-m would need to remove it's inserting of `system`

Honeypot95:
@cho-m Yes, `system` is there as a possible option, albeit not listed, thank you. It was exactly what I needed, and sorry, I did not read thoroughly.

[...]

PR #911, opened 2021-03-31
fix: consistent system completion

Honeypot95:
I am trying to add the "system" parameter for completion, in bash and zsh.

This present solution is similar to how it is done in fish, basically appending the "system" option at the end of the present versions. Because some options (shell, local, global) need this, but others such as uninstall do not need a "system" version in their completions, I was not able to do it elegantly. I had to split the completion code for the commands that do require system, and commands that do not require system.

This solves issue #868.

### DIFF
--- a/completions/_asdf
+++ b/completions/_asdf
@@ -76,6 +76,21 @@ _asdf__installed_versions_of() {
     compadd -a versions
 }
 
+_asdf__installed_versions_of_plus_system() {
+  local plugin_dir="${asdf_dir:?}/installs/${1:?need a plugin version}"
+  if [[ ! -d "$plugin_dir" ]]; then
+    _wanted "asdf-versions-$1" expl "ASDF Plugin ${(q-)1} versions" \
+      compadd -x "no versions installed"
+    return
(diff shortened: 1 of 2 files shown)


jthegedus:
I know the Fish solution was to append `system` to the completions directly, would this be better if we appended `system` to the output of the `list` commands and then removed the appending from the Fish completions. Seems like a more universal solution. Are there reasons you can see why that wouldn't work?

Honeypot95:
@jthegedus The completions for `list` are used indiscriminatively by six commands: `uninstall`, `where`, `reshim`, `shell`, `local`, `global`. Out of these, only for the latter three the `system` completion makes sense (and possibly `where`). This means that if we append `system` to the listing itself, it will appear in places it won't make sense. I suspect that this was the reason it was done like that for `fish`.

I am not saying the code is the best solution, but I personally couldn't see how to do it otherwise.

jthegedus:
Thanks for the investigation and explanation.

jthegedus (approved):
LGTM

PR #800, opened 2020-09-09
fix: <plugin update all> to latest commit on default branch for each plugin

jthegedus:
# Summary

As @joschi pointed out in #799  `asdf plugin update --all` does not respect the set gitref with respect to branch names. Since we cannot provide a `gitref` to each plugin during `update --all` we should update to the latest commit on the ~branch set by the user on `plugin add`~ default branch of the plugin.

Fixes: #799 
Fixes: #813 

# Solution [WIP]

`asdf plugin update <name>`:
- update to latest commit on default branch of plugin repo
 
`asdf plugin update <name> <gitref>`:
- fetch provided gitref
- checkout that gitref

`asdf plugin update --all`:
- for each plugin
- update to latest commit on default branch of plugin repo

## todo

- [x] ~UNTESTED at this time. `--git-dir` usage with `fetch` and `checkout` was not tested by myself~ tested by @joschi 
- [x] create test cases for `asdf plugin update`
  - ensure to capture scenarios where gitref was supplied and then subsequently is not on further updates.
- [x] create test case which covers the initial bug so we get warned on regressions
- [x] update docs and `asdf help` to better describe `asdf plugin update --all` behaviour

## Other Information

- use `--git-dir` and `--work-tree` instead of subshell + `cd`.
	- `--work-tree` has a min git version of `1.7.7.2` according to https://stackoverflow.com/a/9747584/7911479 This shouldn't increase the required git version for `asdf` usage AFAICT. According to wikipedia the latest patch for that minor was `1.7.12.4` released on `2012-10-17`
- use full flag names for `git --prune` and `git --update-head-ok`
- `--git-dir --work-tree` is prefered over `-C` due to the lower supported `git` version

## Caveats

It is worth pointint out, that this code no longer assumes `master` as the default branch, but does assume `origin` as the default name of the remote.

PR #916, opened 2021-04-05
feat: Add versioning to plugin add command

theoretick:
# Summary

Allows user to specify version along with URL when adding plugin to lock
to a specific version instead of always relying on HEAD.

```
❯ git clone https://github.com/theoretick/asdf.git ~/.asdf --branch support-plugin-versions
Cloning into '/Users/theoretick/.asdf'...
remote: Enumerating objects: 8, done.
remote: Counting objects: 100% (8/8), done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 5913 (delta 0), reused 1 (delta 0), pack-reused 5905
Receiving objects: 100% (5913/5913), 1.10 MiB | 12.68 MiB/s, done.
Resolving deltas: 100% (3400/3400), done.

❯ source ~/.asdf/asdf.sh

❯ asdf plugin add java https://github.com/halcyon/asdf-java.git

❯ asdf plugin add java https://github.com/halcyon/asdf-java.git travis-fix

❯ asdf plugin list
java
java-travis-fix

❯
```

## Other Information

Relates to https://github.com/asdf-vm/asdf/pull/234

### DIFF
--- a/lib/commands/command-plugin-update.bash
+++ b/lib/commands/command-plugin-update.bash
@@ -32,8 +32,8 @@ update_plugin() {
   logfile=$(mktemp)
 
   local current_ref
-  current_ref=$(git --git-dir "$plugin_path/.git" rev-parse --abbrev-ref --symbolic-full-name HEAD)
-  if [ "$current_ref" = "HEAD" ]; then
+  current_ref=$(git --git-dir "$plugin_path/.git" branch --show-current)
+  if [ ! "$current_ref" = "$plugin_remote_default_branch" ]; then
     {
       printf "Skipping detached %s\\n" "$plugin_name"
     } >"$logfile" 2>&1
--- a/lib/commands/command-plugin-update.bash
+++ b/lib/commands/command-plugin-update.bash
@@ -30,11 +30,21 @@ update_plugin() {
   plugin_remote_default_branch=$(git --git-dir "$plugin_path/.git" --work-tree "$plugin_path" ls-remote --symref origin HEAD | awk '{ sub(/refs\/heads\//, ""); print $2; exit }')
   local gitref=${3:-${plugin_remote_default_branch}}
   logfile=$(mktemp)
-  {
-    printf "Updating %s to %s\\n" "$plugin_name" "$gitref"
-    git --git-dir "$plugin_path/.git" --work-tree "$plugin_path" fetch --prune --update-head-ok origin "$gitref:$gitref"
-    git --git-dir "$plugin_path/.git" --work-tree "$plugin_path" -c advice.detachedHead=false checkout --force "$gitref"
-  } >"$logfile" 2>&1
+
+  local current_ref
+  current_ref=$(git --git-dir "$plugin_path/.git" rev-parse --abbrev-ref --symbolic-full-name HEAD)
+  if [ "$current_ref" = "HEAD" ]; then
+    {
+      printf "Skipping detached %s\\n" "$plugin_name"
+    } >"$logfile" 2>&1
+  else
+    {
+      printf "Updating %s to %s\\n" "$plugin_name" "$gitref"
+      git --git-dir "$plugin_path/.git" --work-tree "$plugin_path" fetch --prune --update-head-ok origin "$gitref:$gitref"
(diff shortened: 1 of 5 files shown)


theoretick:
@jthegedus would you have a chance to look over this change soon or would there be another review I could pass this one to?

jthegedus:
Sorry mate. I'll be reviewing this the next few days. Trying to get on top of the existing PRs so we can move forward on new changes at greater speed.

Appreciate the work you have contributed.

jthegedus (changes requested):
This is looking good so far.

I would like this change to also include what happens when `asdf plugin update [<name> | --all]` is called on a ref that is a tag. Currently it will ditch the tag and go back to the HEAD of the default branch.

theoretick:
> Is that something you have time to look at @theoretick ?

@jthegedus sorry for the delay but done with db90b80. I don't love the solution but please let me know what you think!

lib/commands/command-plugin-update.bash
```diff
@@ -30,11 +30,21 @@ update_plugin() {
...
   logfile=$(mktemp)
-  {
-    printf "Updating %s to %s\\n" "$plugin_name" "$gitref"
-    git --git-dir "$plugin_path/.git" --work-tree "$plugin_path" fetch --prune --update-head-ok origin "$gitref:$gitref"
-    git --git-dir "$plugin_path/.git" --work-tree "$plugin_path" -c advice.detachedHead=false checkout --force "$gitref"
-  } >"$logfile" 2>&1
+
+  local is_detached=false
+  [ $(git --git-dir "$plugin_path/.git" rev-parse --abbrev-ref --symbolic-full-name HEAD) = "HEAD" ] && is_detached="true"
+  if [[ "$is_detached" == "true" ]]; then
```
theoretick, line 36: FWIW this would probably be much simpler but I chose a more verbose copy to make the command more clear. Or perhaps a simple comment would serve this better

```suggestion
  if [ $(git --git-dir "$plugin_path/.git" rev-parse --abbrev-ref --symbolic-full-name HEAD) = "HEAD" ]; then
```

theoretick:
> This is looking good so far.
> 
> I would like this change to also include what happens when `asdf plugin update [<name> | --all]` is called on a ref that is a tag. Currently it will ditch the tag and go back to the HEAD of the default branch.
> 
> Is that something you have time to look at @theoretick ?

@jthegedus So I attempted this with 58de320 but it seems this approach is at odds with https://github.com/asdf-vm/asdf/pull/800 given https://github.com/asdf-vm/asdf/blob/95f2cdf72c074d72efdb293ae238ef11dbd8a773/test/plugin_update_command.bats#L59 explicitly updates detached HEADs to the default branch 🤔. I'm not sure how best to resolve this change in behavior. Do you have any suggestions?

aabouzaid:
This feature is super crucial to have minimal security measurements.
Otherwise any plugin could be malformed and introduce a security breach.

Stratus3D:
Hi @theoretick , sorry for the late reply here. I agree with @aabouzaid this is important for security reasons.

However, I'm in the process of rewriting asdf in Golang, and the rewrite should be done in a month or two. I don't see much point in getting this merged right now only to be replaced by the Golang implementation. Would you consider contributing these changes to the Golang implementation in a month or two?

theoretick:
@Stratus3D sure, if I have capacity then I'd be happy to port it over and excited to see how the rewrite lands. Thanks!

aabouzaid:
Just in case anyone is waiting for this, you can start using the community project [ASDF Plugin Manager](https://github.com/asdf-community/asdf-plugin-manager) right away and control asdf plugins via `.plugin-versions`.

joeljuca:
Hello. What's the state of this feature (plugin versioning; not the PR itself)? I landed here from a link present in the website (see: https://asdf-vm.com/manage/plugins.html#update), so it might be updated as well.

I'm writing a plugin for CHICKEN and Gambit Schemes and I'm trying to understand if managing plugin versions does make sense – IIUC, asdf always use the latest commit from the default branch of a plugin, am I right?

If that's the case, I'll postpone semantic version management from these plugins.

Issue #919, opened 2021-04-06
`asdf plugin` commands fail to execute


CodingAnarchy:
**Describe the bug**

`asdf plugin` commands return an error indicating that the command is unknown: 

```
Unknown command: `asdf plugin --update all`
No plugin named plugin
```

To update the `asdf-ruby` plugin (the ultimate goal today), it was required to use `asdf plugin-update ruby`, which is not a documented command.

**To Reproduce**

1. `asdf update`
2. `asdf plugin --update all`

**Expected behavior**

`asdf` should have started to update the associated installed plugins.

**Actual behavior**

```
Unknown command: `asdf plugin --update all`
No plugin named plugin
```

### Environment

```shell
OS:
Darwin RCM-YG1H4JGH8 19.6.0 Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64 x86_64 i386 MacBookPro15,1 Darwin

SHELL:
fish, version 3.2.1

ASDF VERSION:
v0.8.0-c6145d0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/Users/mtanous/.asdf

ASDF INSTALLED PLUGINS:
helm                         https://github.com/Antiarchitect/asdf-helm.git
java                         https://github.com/halcyon/asdf-java.git
kubectl                      https://github.com/Banno/asdf-kubectl.git
python                       https://github.com/danhper/asdf-python.git
ruby                         https://github.com/asdf-vm/asdf-ruby.git
shellcheck                   https://github.com/luizm/asdf-shellcheck.git
```

**asdf plugins affected (if relevant)**:

**Screenshots**

**Additional context**

arikai:
I believe this is related: https://github.com/asdf-vm/asdf/issues/899
Seems like portability issue

Issue #899, opened 2021-03-19
"plugin update" broken by recent PR


peaceful-james:
**Describe the bug**

https://github.com/asdf-vm/asdf/pull/800

PR #800 has broken `asdf plugin update` for me.

I can checkout the commit before this PR merge and then my problem goes away. As such, I am pretty sure that this PR is the problem.

I am using Trisquel OS 9.0 and git version 2.29.0

**To Reproduce**

```
asdf plugin update --all
```

**Expected behavior**

Plugins should update

**Actual behavior**

Error message:

```
fatal: couldn't find remote ref f
error: pathspec 'f' did not match any file(s) known to git
```

### Environment

```shell
OS:
Linux my-hostname 5.9.14-gnu #2.0 SMP Tue Sep 27 12:35:59 EST 1983 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
zsh 5.4.2 (x86_64-ubuntu-linux-gnu)

ASDF VERSION:
v0.8.0-456d8e3

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/home/my_home_dir/.asdf
```

jthegedus:
Can you list your entire `asfd info` output? If you run `plugin update` for each of your plugins, can you see which have errors?

peaceful-james:
> Can you list your entire `asfd info` output? If you run `plugin update` for each of your plugins, can you see which have errors?

I get the same error for every plugin.

jthegedus:
I cannot reproduce on the same `asdf` version with the same plugins.

Can you run each of these in a plugin directory:

- `git ls-remote --symref origin HEAD`
- `git ls-remote --symref origin HEAD | awk -F'[[:space:]]*' 'NR == 1 {sub(/refs\/heads\//, "", $2); print $2}'`

How did you add these plugins, `asdf plugin add <name>` or `asdf plugin add <name> <git-url>`?

peaceful-james:
I added all plugins with `asdf plugin-add <name>`

here is what I get in my `elixir` plugin directory:

- `git ls-remote --symref origin HEAD`:
```
ref: refs/heads/master  HEAD
ee5765700c7ff99fa50f2078f2b475ad93384212        HEAD
```
- `git ls-remote --symref origin HEAD | awk -F'[[:space:]]*' 'NR == 1 {sub(/refs\/heads\//, "", $2); print $2}'`: 
```
f
```

jthegedus:
[...]

PR #924, opened 2021-04-18
fix: plugin update portability

jthegedus:
# Summary

Some people are having issues with the implementation of #800 

The issue seems to be the `awk` command, specifically the usage of `-F'[[:space:]]*'`.

>awk already splits by multiple spaces by default so we should not override this option

stated here https://github.com/asdf-vm/asdf/issues/899#issuecomment-816374482

Fixes: #899
Fixes: #919


jthegedus:
Pushing this fix through as it is a minor change and tests pass.

Stratus3D (commented):
Looks good to me!

PR #925, opened 2021-04-20
docs: describe using the system version

jthegedus:
# Summary

Document `system` version usage in the "manage versions" doc page. 

Fixes: #868

## Other Information

This came about as I stated in https://github.com/asdf-vm/asdf/issues/868#issuecomment-787530624:

>We do list `system` on this page - asdf-vm.com/#/core-configuration?id=tool-versions
>But it should probably also be mentioned on this page too - asdf-vm.com/#/core-manage-versions

The OP of the issue PRd the tab completion fix for system on ZSH and Bash completions as the other part of this.

### DIFF
--- a/docs/core-manage-versions.md
+++ b/docs/core-manage-versions.md
@@ -84,6 +84,17 @@ The version format is the same supported by the `.tool-versions` file.
 ASDF_ELIXIR_VERSION=1.4.0 mix test
 ```
 
+## Fallback to system Version
+
+To use the system version of tool <name> instead of an asdf managed version you can set the version for the tool to `system`.
+
+Set system with either `global`, `local` or `shell` as outlined in [Set Current Version](#set-current-version) section above.
+
+```shell
+asdf local <name> system
+# asdf local python system
+```
+
 ## View Current Version
 
 ```shell
--- a/docs/core-manage-versions.md
+++ b/docs/core-manage-versions.md
@@ -84,9 +84,9 @@ The version format is the same supported by the `.tool-versions` file.
 ASDF_ELIXIR_VERSION=1.4.0 mix test
 ```
 
-## Fallback to system Version
+## Fallback to System Version
 
-To use the system version of tool <name> instead of an asdf managed version you can set the version for the tool to `system`.
+To use the system version of tool `<name>` instead of an asdf managed version you can set the version for the tool to `system`.
 
 Set system with either `global`, `local` or `shell` as outlined in [Set Current Version](#set-current-version) section above.
 


docs/core-manage-versions.md
```diff
@@ -84,6 +84,17 @@ The version format is the same supported by the `.tool-versions` file.
 ASDF_ELIXIR_VERSION=1.4.0 mix test
 ```
 
+## Fallback to system Version
```
Stratus3D, line 87: Why is system not capitalized here?

docs/core-manage-versions.md
```diff
@@ -84,6 +84,17 @@ The version format is the same supported by the `.tool-versions` file.
...
+
+To use the system version of tool <name> instead of an asdf managed version you can set the version for the tool to `system`.
+
+Set system with either `global`, `local` or `shell` as outlined in [Set Current Version](#set-current-version) section above.
+
+```shell
+asdf local <name> system
+# asdf local python system
+```
```
Stratus3D, line 96: This seems like a very special case and I'm not sure it makes sense to put this on a popular page above common operations like viewing the current version and uninstalling versions.

markburns:
Tangential to the pull request itself, but I think the existence of this branch with a slash in the name is breaking `asdf update`. Perhaps because there is already a `docs` branch

```
asdf update
error: cannot lock ref 'refs/remotes/origin/docs/system-version-in-manage-versions-section': 'refs/remotes/origin/docs' exists; cannot create 'refs/remotes/origin/docs/system-version-in-manage-versions-section'
From https://github.com/asdf-vm/asdf
 ! [new branch]      docs/system-version-in-manage-versions-section -> origin/docs/system-version-in-manage-versions-section  (unable to update local ref)
```

docs/core-manage-versions.md
jthegedus: I didn't capitalize "system" because it is the value in question which is all lowercase, `system`. I will change.

jthegedus:
@markburns I cannot see that branch under - https://github.com/asdf-vm/asdf/branches/all I guess it must have existed on the remote when the full clone was performed?

Seems an odd error since it is a valid git ref.

docs/core-manage-versions.md
jthegedus: Placing this after "Set Current Version" seemed the most logical place.

I view this page as the comprehensive list rather than a popular subset.

markburns:
Sorry I'm sure this is not the right place for this discussion, but I'll just paste it here so you can debug in case you get any other support issues like this.

But it looks like it points at https://github.com/asdf-vm/asdf/commits/b65fbf98a70f52560b219de14105087b1b91f510 which I determined by looking in `asdf/.git/packed-refs`
which looks like:

```
1da25256b3c88d650c4d15ac20e4de1 refs/remotes/origin/bash-strict-mode
7876fcdd9002f8fc87bb50b15b7ceb8bbf84bc8d refs/remotes/origin/debug-fish-test
b65fbf98a70f52560b219de14105087b1b91f510 refs/remotes/origin/docs
72630a0b10e82b54830be7c2a4b0449e19cbe331 refs/remotes/origin/improve-current-behavior
a33819958f125f7c2a3a5f22782a07af443add9c refs/remotes/origin/improve-multi-versions-support
12c9a9c53439caba5c328ed452bdc9518564f711 refs/remotes/origin/keep-source
ac1a35b85bde049b9e2d531032eb55534e38ffe7 refs/remotes/origin/master
67f0e743be5db9bf0b6ea3173c560393fc6f4834 refs/remotes/origin/pr-108
c11e9f69a392596a05e190fca24cfcf60749ddca refs/remotes/origin/qwer-rename
4d146ea652c5bec09a59bf60c432dfcbcda408ea refs/remotes/origin/speedup-reshim
9eff4b00feb5fca769a4a6f95c227707df040feb refs/remotes/origin/tb/bash-strict-mode
8e085bfc5aacb2c5c32369b0d2fa8d4cc0faaa85 refs/remotes/origin/tb/no-unset-vars
728155cccab965162741b6cf78a7b1cff581b2fe refs/remotes/origin/which-command
ae2eb8b75d07ec787e9e603cc0e89affb9c57c6f refs/tags/v0.1.0
^337b335c4db5949bfa8c61c94c11abbf72378d4f
3310fedb500a7a3916fe94320b8d8e928869f8b4 refs/tags/v0.2.0
^dab7d51f224480620ee02034a0869e33188e57bb
16f8335a41fbe324dbc91ad08169c6a9c355da2e refs/tags/v0.2.1
^2e96bc63ebe8232d3ff675bfd461c089e7a01d2d
324906af1e406bc8976cf74dccc403bc84ebaaa3 refs/tags/v0.3.0
^d3a66d78ee60322609967846142d5fd56ffe6a2f
a099d99cdec6f87132ca03868d925d5a95f35c09 refs/tags/v0.4.0
^d19f6131a6745517070a895e89b9caf2fd912e88
716bc07a5c370ee00990a6a6d33ce84beb9301ba refs/tags/v0.4.1
^12db51e483968d29264b22cd75685682898ccc41
...
```

I've seen this occasionally in other repos. I'm surprised this kind of issue doesn't happen more often. But I think it's caused by git allowing `/` within a branch name but expanding them in the `refs` directory into paths, and then also expecting not to find a reference at any non-leaf node in the tree. So whilst the top-level `docs` is fine. It only becomes a problem later when clashing with a `docs/x`. But I think `docs/x`, and `docs/y` are compatible.

So yes, the creation of the `docs` branch followed by the creation of a `docs/anything` branch is the trigger here.
It seems deleting the reference for me in the `.git/packed-refs` file fixes it.

jthegedus:
Pushing this through as we can change the location in the docs in the future.

jthegedus:
@markburns I would have expected this issue to appear in each repo using gitflow style naming conventions.

For easier discoverability, can you create a new issue with all this information in it, your solution and then close the issue? That way if someone else runs into that problem they will find an actual Issue history, thanks

Issue #763, opened 2020-07-13
`asdf latest` command should defer to plugin to determine latest version


Stratus3D:
**Is your feature request related to a problem? Please describe.**

The `asdf latest` command assumes versions listed by the `list-all` command are semantic versions. This assumption is not necessarily correct for all plugins and could cause problems.

**Describe the solution you'd like**

The `asdf latest` should invoke a plugin callback named `latest-stable` when one is present to compute the latest version. If the `latest-stable` script is missing the `asdf latest` command should fallback to computing the latest version as it does now by assuming the versions are semantic versions.

**Describe workarounds you've considered**

None. This seems to be the most straightforward solution to this problem.

**Additional context**

This is actually what was requested for https://github.com/asdf-vm/asdf/issues/216 but was never properly implemented.

* https://github.com/asdf-vm/asdf/pull/575
* https://github.com/asdf-vm/asdf/issues/216

jthegedus:
Will this also resolve https://github.com/asdf-vm/asdf/issues/648 ?

threkk:
I have the same issue here. I am working on a [plugin](https://github.com/threkk/asdf-age) for [age](https://github.com/FiloSottile/age) and even if age is stable enough for being recommended to use with [sops](https://github.com/mozilla/sops) among other use cases, the command `asdf latests` fails (and therefore, the pipeline) as all the versions are tagged with `rc` or `beta`. Having the change of explicitly declaring which one is the latest stable version would allow the plugin to work.

Issue #648, opened 2020-02-01
Can't get the latest version of miniconda3


goulartdev:
### Steps to reproduce

`asdf latest python miniconda3`

### Expected behavior

Should return the latest version of miniconda 3

### Actual behavior

It returns nothing

### Environment

**OS**: WSL/Ubuntu

**asdf version**: 0.7.6

The problem is with this regex, in asdf/lib/commands/command-latest:

`grep -vE "(^Available versions:|-src|-dev|-latest|-stm|[-\\.]rc|-alpha|-beta|[-\\.]pre|-next|(a|b|c)[0-9]+|snapshot|master)"`

The pattern `(a|b|c)[0-9]+` will remove all anaconda2, anaconda3, miniconda2 and miniconda3 from the result. This could be a solution:

`(a|b|c)(?<!conda)[0-9]+`

Or, since this could happen with other plugins, maybe a more general solution would be better:

`-(a|b|c)[0-9]+`

But I don't know which plugins uses this `(a|b|c)[0-9]+` pattern, so changing this could break another plugin. Perhaps it should be possible for a plugin to overwrite this regex or the entire command, so this can be handled individually.

Stratus3D:
The command you want is `asdf latest python miniconda-3`. The dash is not ignored, and the only version of miniconda 3 is `miniconda-3.18.3`. If you don't care about the miniconda version just `asdf latest python miniconda` should give you all available miniconda versions.

DjouGoulart:
Hey Stratus, thank you for answering. This can be a little confusing, but there is two versions os miniconda. Miniconda/miniconda2 are based on python 2 (in some point, miniconda was renamed to miniconda2), and miniconda3 is based on python 3. So: 

- `asdf latest python miniconda2` should return `miniconda2-4.3.30`
- `asdf latest python miniconda3` should return `miniconda3-4.3.30`

But both return nothing because of that pattern in te regex. Typing just `asdf latest python miniconda` will return `miniconda-3.18.3`, that is a old version of miniconda2, released in 2015. And when I type `asdf install python latest:miniconda`, this old version will be installed, not the latest.  Sorry if that wasn't clear. More about that:

https://github.com/astroconda/astroconda/issues/46

Stratus3D:
[...]

PR #938, opened 2021-05-11
feat: asdf latest defer to plugin to determine the latest version

threkk:
# Summary

If the command `latest-stable` exists in the plugin, `asdf latest` defers to it to determine the latest stable version.

Fixes: #763 
Resolution of #648 now falls to plugin authors.


threkk:
I have opened this pull request which solves an existing an existing issue about three weeks ago, but it has gotten no attention since then. Is there anything missing or anything I can do so it can be taken in consideration? :(

jthegedus:
@threkk Sorry mate, will look at this in the next couple of days. We appreciate the help and this PR isn't going unnoticed. Sometimes I don't look at a PR until I know I have enough time to give it attention. Which these past few weeks hasn't been much time at all.

Will look at this next.

jthegedus:
@Stratus3D I am confident the changes here so will push through.

I will clean up the testing in a future PR as we need to test most of the code against 2 different plugins, ones with the minimum scripts and another with all optional scripts like `latest-stable` as added here. In the meantime I have used `dummy_legacy_plugin` though this is incorrect.

Additionally, I will PR some docs updates about plugin API. We're currently handling calling out to scripts from plugins in different ways, expecting different results, eg: `list-all` does a lot more work to capture plugin reported errors, whereas here (`latest-stable`) we expect something or nothing and no `stderr`s, though we're not guarding against plugin reported errors in these scripts.

jthegedus (approved):
@threkk Thanks for contributing this. I added more testing scenarios and merged with the changes on `master` that came from #633

threkk:
Thank you very much for taking it in consideration. This will also allow me to finish a plug-in I was working on too :)

lib/commands/command-latest.bash
```diff
@@ -5,18 +5,31 @@ latest_command() {
...
-  versions=$(asdf list-all "$plugin_name" "$query" |
-    grep -vE "(^Available versions:|-src|-dev|-latest|-stm|[-\\.]rc|-alpha|-beta|[-\\.]pre|-next|(a|b|c)[0-9]+|snapshot|master)" |
-    sed 's/^\s\+//' |
-    tail -1)
-
-  if [ -z "${versions}" ]; then
-    exit 1
+
+  if [ -f "${plugin_path}/bin/latest-stable" ]; then
+    versions=$(bash "${plugin_path}"/bin/latest-stable "$query")
```
Stratus3D, line 18: FYI at some point hopefully in the near future I'd like to explore allowing callbacks to be written in any language and not just bash. We'd need to change the callback invocations from `bash "${plugin_path}"/bin/<callback> <args>` to just `"${plugin_path}"/bin/<callback> <args>`. This will give plugin creators more control over how they implement callbacks.

lib/commands/command-latest.bash
```diff
@@ -5,18 +5,31 @@ latest_command() {
...
+    if [ -z "${versions}" ]; then
+      # this branch requires this print to mimic the error from the list-all branch
+      printf "No compatible versions available (%s %s)\n" "$plugin_name" "$query" >&2
+      exit 1
+    fi
+  else
+    # pattern from xxenv-latest (https://github.com/momo-lab/xxenv-latest)
+    versions=$(asdf list-all "$plugin_name" "$query" |
+      grep -vE "(^Available versions:|-src|-dev|-latest|-stm|[-\\.]rc|-alpha|-beta|[-\\.]pre|-next|(a|b|c)[0-9]+|snapshot|master)" |
```
Stratus3D, line 27: I know this isn't a new change, but I think `Available versions` could be left out of this list. Plugins should just be printing the list of available versions and nothing else. An `Available versions` heading is not what is specified in the docs.

Stratus3D:
@jthegedus thank you for handling. I am sorry I dropped the ball on this.

@threkk thank you for this PR! I'd been meaning to implement this on my own for months now, this is a great improvement!

lib/commands/command-latest.bash
threkk: Sure! It sounds good. I don't mind to look into it when the time comes.

lib/commands/command-latest.bash
threkk: Makes sense. I decided to leave it as it was purely for backward compatibility, just in case someone relied on it.

PR #944, opened 2021-05-19
chore: comment out instructions in issue/pr templates

jthegedus:
# Summary

Issue and PR reporters often do not remove the instructions from the template before submitting. This fixes that issue by commenting them out. They are visible when creating the issue, but do not require deletion before submission and do not pollute the Issue/PR for the readers.


Stratus3D:
Nice improvement!

Issue #948, opened 2021-05-20
Remove banned process substitution & force PR base branch merge


jthegedus:
**Describe the bug**

Process substitution was banned in #863 but I merged #881 first which included a process substitution. This was not caught because we do not require PRs to be update with `master` before merge.

**To Reproduce**

Reported: https://github.com/asdf-vm/asdf/commit/a7d3661f6c53b24ae1c21e93f94209f3af243349#r51098840

Code: https://github.com/asdf-vm/asdf/blob/156a82cc494e97955ddf44d801472ca105159338/lib/commands/command-list-all.bash#L45

**Expected behavior**

- [x] no process substitution
- [x] PRs should be required to be up to date with master

jthegedus:
I have added the branch protection rule to enforce branches are synced with `master` before merging:

<details>
<summary>Before & After Images</summary>

![asdf-branch-protection-before](https://user-images.githubusercontent.com/20798510/119058625-15a14400-ba12-11eb-8eea-50a817bdb005.png)

![asdf-branch-protections-after](https://user-images.githubusercontent.com/20798510/119058627-18039e00-ba12-11eb-8547-34e532ee8523.png)

</details>

We can disable this if it becomes a nuisance.

Stratus3D:
Closing this as I have merged a PR to remove the process substitution code from the `list-all` command. 🎉

PR #949, opened 2021-05-20
ci: push-event triggers CI only on master branch

jthegedus:
# Summary

Currently our PRs run 2 sets of the workflow for each of the `push` and `pull_request` events. We should only run one set of events on PRs which should be the PR events. Push events can run on `master` only.

I don't think this changes any developer workflows (observing push builds before opening PRs), but please discuss if this impacts your experience.

I also removed the cron execution as it is not really adding value.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -2,13 +2,13 @@ name: Main workflow
 
 on:
   push:
+    branches:
+      - master
     paths-ignore:
       - "**.md"
   pull_request:
     paths-ignore:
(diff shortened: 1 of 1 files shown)


jthegedus:
Probably need to fix #948 before everything is green here.

jthegedus:
Force merging as #948 is being fixed in #890

Stratus3D:
Thanks @jthegedus, I think these are good changes. Cron job wasn't adding any value in my experience.

Issue #465, opened 2019-02-13
Failing tests on windows.


abhishalya:
In reference to https://github.com/asdf-vm/asdf/issues/450 and https://github.com/asdf-vm/asdf/pull/451 I have been preparing appveyor CI to run tests. Initially I found that most of the tests are failing with the usual bash for windows. To get the maximum compatibility, I switched to cygwin bash and now most of the tests pass directly.

But some of the tests are still failing, six are from the file [shim_exec.bats](https://github.com/asdf-vm/asdf/blob/master/test/shim_exec.bats) and one is from [which_commands.bats](https://github.com/asdf-vm/asdf/blob/master/test/which_command.bats). Following are the tests which are failing:

1. `asdf exec should pass all arguments to executable even if shim is not in PATH`
2. `shim exec should fallback to system executable when specified version is system`
3. `shim exec should execute system if set first`
4. `shim exec doest not use custom exec-env for system version`
5. `shim exec should remove shim_path from path on system version execution`
6. `shim exec doesnt execute command if pre-hook failed`
7. `which should show path of system version`

I tried printing the output of each test-command to terminal, and I think that all of the commands fail due to something in the `$PATH` variable. It will be helpful if someone could guide me through the required steps for solving this.

You can see the log [here](https://ci.appveyor.com/project/abhishalya/asdf/builds/22330530).

deiga:
Do we actually care about windows as a CI platform? `BATS` is a `bash` test runner and it's not like windows has its own version of `bash`.
We should probably just delete the windows CI stuff to not confuse people.

jthegedus:
It seems [WSL2 is available on the GitHub Actions Windows environment](https://github.com/actions/virtual-environments/issues/50), so I would like to co-opt this Issue to target adding tests for that.

Related issues to WSL:
- #730 
- #573 
- #558

jthegedus:
GitHub Actions for WSL setup - https://github.com/marketplace/actions/setup-wsl

jthegedus:
>Do we actually care about windows as a CI platform?

We do care about the test runner. We want to execute our tests in all environments/platforms `asdf` can run. WSL1 / WSL2 / Ubuntu / macOS all have system quirks and differences, notably filesystem interactions etc which we interact with and should capture.

#956 has added the WSL1 environment using [setup/wsl](https://github.com/marketplace/actions/setup-wsl) GH Action. We still need to update the tests.

earonesty:
[...]

Issue #942, opened 2021-05-19
Automated Releases


jthegedus:
**Is your feature request related to a problem? Please describe.**

Our release process is currently manual. It should be automated. We should be releasing more often and if errors are found, fix and move forward instead of taking an overly-cautious approach to releases.

People want the releases!! #921 

**Describe the solution you'd like**

Use https://semantic-release.gitbook.io/semantic-release/ to manage our release process.

It uses a semver with `fix`/`feat`/`BREAKING_CHANGE` nomenclature in commit messages to identify which part of semver to increment.

Semantic release will run in a GitHub Action and require a new config file in the repo. It will not require Node.js resources in the repo as previously thought ([guide for non-JS packages](https://semantic-release.gitbook.io/semantic-release/support/faq#can-i-use-semantic-release-to-publish-non-javascript-packages)).

It can run arbitrary shell scripts so we will likley use a slight modification of the current release script.

It can generate additions to the Changelog with links to PRs.

While supporting many release styles, of varying complexity, I would like to keep things simple and just have a new release published on merges to the main repo branch.

**Caveats**

To upload assets of the tagged release to GitHub releases page you need the Semantic Release GitHub plugin which then requires some workarounds for protected branches - https://github.com/semantic-release/github/issues/175

Whether we **need** to upload an archive of assets to GitHub release page is up for discussion.

**Additional context**

I have been using Semantic Release for a few projects of my own with great success. See https://github.com/jthegedus/svelte-adapter-firebase/ for an example.

NeoHsu:
That's good job!
Maybe can use some tools to help this topic,
I usually use [git-cz](https://github.com/commitizen/cz-cli) to do standardized commit message and use [standard-version](https://github.com/conventional-changelog/standard-version) to define release version .
this is my project build & release flow, FYI.

jef:
I prefer https://github.com/google-github-actions/release-please-action. It still uses conventional commits, but I like it's agnostic pattern more. Just my two cents 😁

jthegedus:
I really like [release-please](https://github.com/google-github-actions/release-please-action) as well, except it ends up functionally being the same or ends up making a lot of noise via many PRs if you don't merge immediately.

jef:
> as it closes the release PR to create a new PR instead of updating

[...]

PR #956, opened 2021-05-25
ci: adopt asdf for internal dev. init WSL test environments

jthegedus:
# Summary

Adopt `asdf` for out tests, internal dev. Init WSL test envs. Cleanup Vagrantfile.

- remove unused `Vagrantfile`
- provide `.tool-versions` for `asdf` development and CI workflows
- extract Lint and Format GitHub Workflows and use `.tool-versions` with `asdf/actions`
  - remove `lint.sh` as our CI pipeline now enforces all formatting and linting reliably, so no need to run this script manually before release, nor for it to exist outside the CI Workflow definition.
- extract Bats testing to individual jobs per OS type.
  - nix:
    - install test deps as before
    - run on each of `macos-10.15`, `macos-11` (disabled as it's in preview), `ubuntu-18.04` & `ubuntu-20.04`
  - wsl1:
    - setup WSLv1 with various distributions with additional test deps
    - install bats-core manually
    - run on `windows-2019`. `windows-2016` is not supported with WSL GitHub Action https://github.com/Vampire/setup-wsl
    - `continue-on-error: true` as the failing tests will be addressed later. Tracking in #465


jthegedus:
This doesn't change our existing Ubuntu/macOS tests, only expands them with an `matrix.os` list. It adds the scaffolding for WSL1 tests to be fixed.

I am of the mind we should probably **not** use `continue-on-error: true` in our WSL1 tests. This gives the false impression our tests are working when they are not. We should be using "required status checks" to set which tests are **required** vs not required for PRs to be considered valid.

image

For example, we should require all except WSL1 checks pass for the PR to be valid. We have not been using required status checks as the event trigger filters `paths-ignore: ["**.md"]` would mean PRs with only `.md` changes would not trigger the Jobs which are required, thus getting into annoying PR states. I have removed those in anticipation for others opinions.

jthegedus:
To other asdf core members:

I am merging as there are no negative side-effects I can see and this can easily be reverted.

Enabling Required Status Checks as well. These can be disabled via https://github.com/asdf-vm/asdf/settings/branches

lint.sh
```diff
@@ -1,8 +0,0 @@
-#!/usr/bin/env bash
-exec shellcheck -s bash -x \
-  asdf.sh lint.sh release/tag.sh \
-  bin/asdf bin/private/asdf-exec \
-  lib/utils.bash lib/commands/*.bash \
-  completions/*.bash \
-  test/test_helpers.bash \
-  test/fixtures/dummy_plugin/bin/*
```
Stratus3D, line 8: I use this file. Please restore it.

Stratus3D:
@jthegedus how do we perform the necessarily shellcheck linting and shfmt formatting when working locally? I don't rely on GitHub actions for validating my changes. I liked the `lint.sh` script because it ran shellcheck for me locally. I'd like another script for running shfmt locally.

Recently I've been creating PRs with builds that fail because I don't have a good way to run shfmt locally.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,95 @@
...
+          $HOME/bin/bats --version
+
+      - name: Run tests
+        shell: wsl-bash {0}
+        run: |
+          $HOME/bin/bats test
+        env:
+          GITHUB_API_TOKEN: ${{ github.token }}
+
+  nix:
```
Stratus3D, line 59: Why is this job named `nix`?

lint.sh
jthegedus: Will do.

release/README.md
```diff
@@ -1,28 +1,19 @@
...
-tagging as well as a dev version things that come after the tag (e.g. a heading
-with the format `<next-version>-dev`).
-2. Run the tests and the linter - `bats test` and `lint.sh`.
-3. Run the release script. The new version must be in the format `0.0.0-rc0`.
-For example: `release/tag.sh 0.0.0-rc0`.
-4. If the release script succeeds, push to GitHub. Make sure to use the correct
-remote to push to the official repository
+
+1. Update the CHANGELOG. Make sure it contains an entry for the version you are tagging as well as a dev version things that come after the tag (e.g. a heading with the format `<next-version>-dev`)
+2. Run the tests and the linter - `bats test`
```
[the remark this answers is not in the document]
jthegedus: I will put the lint script back.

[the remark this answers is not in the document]
jthegedus: IMO the release process should be entirely separate from running tests and linting. At the point of release the code should already have passed our CI which performs these assertions.

#942 would address the need for this entire file and process with the description of the release process moved somewhere like the contribution docs

.github/workflows/tests.yml
jthegedus: It is the job that runs our Bats tests on the *nix OSs. `*` is invalid in the job name when I tried naming it `*nix`. Didn't think the job name was particularly important.

The other job is called `WSL1` because though it is a Windows OS, the environment we want to test in is different distros on WSL1 environment.

jthegedus: We can rename the job to whatever, I just thought this was the most descriptive at the time.

lint.sh
jthegedus: See #961

release/README.md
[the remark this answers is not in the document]
jthegedus: resolved in #961

jthegedus:
See #961 for the revert of Shellcheck and addition of Shfmt scripts for local execution.

With the addition of `.tool-versions` in this PR, we should all be using the same versions of all these tools.

Stratus3D:
Thanks @jthegedus !

jthegedus:
Sorry for being impatient and breaking your workflow. My distraction with CI took me away from my initial goal of adding `.tool-versions` so we are all dev-ing with the same versions locally.

PR #957, opened 2021-05-26
feat: configurable plugin repo last check time

NeoHsu:
# Summary
Add update plugins repo function to `asdf plugin-update` command that can be get latest asdf plugins list

Hi everyone,
I tried to update the asdf plugins today,
but I didn't get latest plugins list.

I can't find a way to update asdf plugin list, 
so I implement this function in plugin update command.

I think about plugin update command for user flow,
When user want to update plugin, 
they maybe want to get latest plugin list,
so I add trigger function on plugin update command.

I hope to help this project to be better.

Thanks.

### DIFF
--- a/docs/core-configuration.md
+++ b/docs/core-configuration.md
@@ -45,6 +45,7 @@ legacy_version_file = yes
 - `legacy_version_file` - defaults to `no`. If set to yes it will cause plugins that support this feature to read the version files used by other version managers (e.g. `.ruby-version` in the case of Ruby's `rbenv`).
 - `use_release_candidates` - defaults to `no`. If set to yes it will cause the `asdf update` command to upgrade to the latest release candidate release instead of the latest semantic version.
 - `always_keep_download` - defaults to `no`. If set to `yes` it will cause `asdf install` always keep the source code or binary it downloads. If set to `no` the source code or binary downloaded by `asdf install` will be deleted after successful installation.
+- `plugin_repository_last_check_duration` - defaults to `60` mins (1 hrs). It will set asdf plugins repository last check duration. When `asdf plugin add <name>`, `asdf plugin list all` command be executed, it will check last update duration to update repository. If set to `0` it will update asdf plugins repository every time.
 
 ## Environment Variables
 
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -384,9 +384,11 @@ repository_needs_update() {
   update_file_dir="$(asdf_data_dir)/tmp"
   local update_file_name
   update_file_name="repo-updated"
-  # `find` outputs filename if it has not been modified in the last day
+  # `find` outputs filename if it has not been modified in plugin_repository_last_check_duration setting.
+  local plugin_repository_last_check_duration
+  plugin_repository_last_check_duration="$(get_asdf_config_value "plugin_repository_last_check_duration")"
   local find_result
-  find_result=$(find "$update_file_dir" -name "$update_file_name" -type f -mtime +1 -print)
+  find_result=$(find "$update_file_dir" -name "$update_file_name" -type f -mmin +"${plugin_repository_last_check_duration:-60}" -print)
   [ -n "$find_result" ]
 }
 
--- a/test/plugin_list_all_command.bats
+++ b/test/plugin_list_all_command.bats
@@ -26,6 +26,18 @@ foo                           http://example.com/foo"
   [[ "$output" =~ "$expected_plugins_list" ]]
 }
 
+@test "plugin_list_all no immediate repo sync expected because check_duration is greater than 0" {
+  echo 'plugin_repository_last_check_duration = 10' > $HOME/.asdfrc
+  run asdf plugin-list-all
+  local expected="\
+bar                           http://example.com/bar
+dummy                        *http://example.com/dummy
+foo                           http://example.com/foo"
+
+  [ "$status" -eq 0 ]
+  [ "$output" = "$expected" ]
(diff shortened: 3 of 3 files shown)


jthegedus:
Hi @NeoHsu thanks for proposing this change, I think we do need to communicate this better. I have been thinking about making a change like this. Some background on this:

- asdf plugin repo is updated once per every **two** days
  - code: https://github.com/NeoHsu/asdf/blob/7d836cdcdec443830656f9f0bd2f8dcd94c63241/lib/utils.bash#L389
- the update is triggered by either command: `asdf plugin add <tool>` or `asdf plugin list all`
  - we **need to document this**
- adding this to `asdf plugin update` is fine, but would also need to be documented.

Your proposal splits the functionality of the plugin repo update into two:
1. always check when running `asdf plugin update` 
2. use the time-based check in `asdf plugin list all` and `asdf plugin add`

I think there is a more simple and user-configurable solution that gives us both results without the split in functionality:

- use `find -mmin` instead of `find -mtime`. This changes our check resolution to minutes from days.
- set the `-mmin` time value from a config value in `.asdfrc`. This lets user's customise the frequency of the update check. Default to 3600.
- no need extract new function as we can then just use `initialize_or_update_repository`
- no need to print the updates out as people can expect and know when they are on the latest.

What do you think of this suggestion?

NeoHsu:
That's cool !!
I didn't notice this command `find -mtime`.

But I think print plugins updates is very important.
The asdf be designed very easy to build new plugin.
Many develpors can build plugin into asdf by self.
I will want to look which new plugin have be join to asdf.

Thanks for getting back to me.

NeoHsu:
Hi @jthegedus ,
I modify this PR to follow your suggestion,
thanks

lib/utils.bash
```diff
@@ -384,9 +384,12 @@ repository_needs_update() {
   update_file_dir="$(asdf_data_dir)/tmp"
   local update_file_name
   update_file_name="repo-updated"
-  # `find` outputs filename if it has not been modified in the last day
+  # `find` outputs filename if it has not been modified in the update plugins repository duration.
+  # Default 1 day = 1440 mins
```
jthegedus, line 388: Can we move this to the documentation?

defaults to `1440`mins (24hrs)

docs/core-configuration.md
```diff
@@ -45,6 +45,7 @@ legacy_version_file = yes
 - `legacy_version_file` - defaults to `no`. If set to yes it will cause plugins that support this feature to read the version files used by other version managers (e.g. `.ruby-version` in the case of Ruby's `rbenv`).
 - `use_release_candidates` - defaults to `no`. If set to yes it will cause the `asdf update` command to upgrade to the latest release candidate release instead of the latest semantic version.
 - `always_keep_download` - defaults to `no`. If set to `yes` it will cause `asdf install` always keep the source code or binary it downloads. If set to `no` the source code or binary downloaded by `asdf install` will be deleted after successful installation.
+- `update_plugins_repository_duration` - defaults is `1440` mins. It will set asdf plugin repository updated duration. When `asdf plugin add`, `asdf plugin list all` command be executed, it will check last update duration to update repository.
```
jthegedus, line 48: can we name this something like: `plugin_repository_last_check_duration` or `plugin_repository_last_check_duration`?

I am trying to think of something more descriptive but naming this is hard :thinking:

lib/utils.bash
jthegedus: I am also happy to change the default to something lower, like 60mins if 1 day is too high

jthegedus:
This is looking like a good improvement @NeoHsu

@Stratus3D do you have thoughts here? I am not sure how portable the `-mmins` option is. I can't find much online.

NeoHsu:
Hi @jthegedus, I'm finish now,
Thanks for your suggestions.

jthegedus:
I'm going to write a couple of tests for this:
1. for the `plugin_repository_last_check_duration = 0` case. Should expect sync.
2. for the default case. Should expect no sync.

jthegedus (approved):
This looks good to me now. Thanks again

Stratus3D:
> This is looking like a good improvement @NeoHsu
> 
> @Stratus3D do you have thoughts here? I am not sure how portable the `-mmins` option is. I can't find much online.

Sorry I missed your comment. I am not sure either, but I also don't see folks online reporting any issues with the flag not being supported, so I guess we will see how it goes.

I agree this is a nice improvement. Is there a way to turn off checking completely? e.g. is there another configuration option for that? Or can I somehow set this last check time to infinity?

jthegedus:
> Is there a way to turn off checking completely? e.g. is there another configuration option for that? Or can I somehow set this last check time to infinity?

Good questions. I would prefer a solution that does not require another config option. I believe the numeric value for `mmin` is represented by gnulib timespec which has a max value of `999999999` which equates to 1902 years. We could introduce a special val like `plugin_repository_last_check_duration = never`.

Stratus3D:
> > Is there a way to turn off checking completely? e.g. is there another configuration option for that? Or can I somehow set this last check time to infinity?
> 
> Good questions. I would prefer a solution that does not require another config option. I believe the numeric value for `mmin` is represented by gnulib timespec which has a max value of `999999999` which equates to 1902 years. We could introduce a special val like `plugin_repository_last_check_duration = never`.

I would like something like `plugin_repository_last_check_duration = never` that tells asdf not to make any unnecessary network connections (I like to work offline so checking for updates is something I'd like to disable completely.

Issue #946, opened 2021-05-20
Document `asdf plugin list all`/`asdf plugin add <name>` perform plugin repo sync based on asdfrc `plugin_repository_last_check_duration` option


jthegedus:
**Is your feature request related to a problem? Please describe.**

People have questioned why the plugin they added (to the plugin repo) is not installable via the short-name from the plugin repo moments after the plugin has been added to the repo.

Usually, the user is aware that the plugin repo syncs, however it's not clear **when** ~(I believe once daily) and most aren't aware they can force the sync via the `asdf plugin list all` command.~

`asdf plugin add <name>` and `asdf plugin list all` will perform a check at most once per hour (since #957). This value is also configurable via `.asdfrc` since #957

**Describe the solution you'd like**

Document this in the docs site page: https://asdf-vm.com/#/core-manage-plugins?id=update

It is currently documented in the config section https://asdf-vm.com/#/core-configuration?id=homeasdfrc but is asked often enough we should mention it elsewhere.

**Describe similar asdf features and why they are not sufficient**

**Describe workarounds you've considered**

**Additional context**

Any question we are asked via an issue should be added to the docs site or the functionality in question have its first point of user contact improved.

PR #965, opened 2021-06-02
docs: explain plugin repo sync details

jthegedus:
# Summary

- explain plugin repo syncing and that it is configurable
- add missing explanations for each add, update, remove plugin ops as they are listed in `asdf help` but not in the docs

I also started refering to the plugin repo as the "short-name" repo as it seems people are confusing this as the only way to install plugins. This is going as far as requests to customise the plugin repo location which is completely unnecessary when you can add via Git URL. Refering to it as the short-name repo, hopefully will indicate this isn't the sole plugin repo. Though, this is just a fix of a symptom and not the cause IMO.

I also swapped the docs order so adding via Git URL comes first as it is the better method for people to use.

Fixes: #946

### DIFF
--- a/docs/core-manage-plugins.md
+++ b/docs/core-manage-plugins.md
@@ -1,24 +1,24 @@
-Plugins are how asdf knows how to handle different packages.
+Plugins are how asdf knows how to handle different tools like Node.js, Ruby, Elixir etc.
 
-See [Creating Plugins](plugins-create) for the plugin API used to support more languages.
+See [Creating Plugins](plugins-create) for the plugin API used to support more tools.
 
 ## Add
 
-Add via the short-name association in the plugins repository:
-
-```shell
-asdf plugin add <name>
-# asdf plugin add erlang
-```
-
-If not in the short-name repo, you can add it using its repository URL:
+Add plugins via their Git URL:
 
 ```shell
 asdf plugin add <name> <git-url>
 # asdf plugin add elm https://github.com/vic/asdf-elm
 ```
 
-When scripting, prefer this longer `git-url` method as it is independent of the short-name repo.
+or via the short-name association in the plugins repository:
+
+```shell
+asdf plugin add <name>
+# asdf plugin add erlang
+```
+
+When scripting, prefer the longer `git-url` method as it is independent of the short-name repo.
 
 ## List Installed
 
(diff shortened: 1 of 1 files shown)


docs/core-manage-plugins.md
```diff
@@ -1,24 +1,25 @@
-Plugins are how asdf understands handling different packages.
+Plugins are how asdf knows how to handle different packages.
```
Stratus3D, line 1: "packages" here might be a little ambigous. It might be good to elaborate, something like "how to handle different tools, like Ruby, Elixir, Node.JS, etc..."

docs/core-manage-plugins.md
```diff
@@ -1,24 +1,25 @@
...
 
-See [All plugins](plugins-all) for the plugins repository which lists all asdf plugins we know.
-
-See [Creating Plugins](plugins-create) for the super-simple API for supporting more languages.
+See [Creating Plugins](plugins-create) for the plugin API used to support more languages.
```
Stratus3D, line 3: Above we say "packages" and here we use the word "languages". Not some of the things asdf versions are not languages. "support more tools" or "support more tools and languages" might be better.

docs/core-manage-plugins.md
```diff
@@ -1,24 +1,25 @@
...
+Add via the short-name association in the plugins repository:
+
 ```shell
 asdf plugin add <name>
 # asdf plugin add erlang
 ```
 
-If the plugin you want to install is not part of the plugins repository, you
-can add it using its repository URL:
+If not in the short-name repo, you can add it using its repository URL:
```
Stratus3D, line 14: nitpick: you can add it via URL regardless, and the plugins repository displays URLs for this purpose. It might be better to just say:

"Alternatively, you can add a plugin using its repository URL:"

Stratus3D:
Left a few comments. Otherwise looks good to me!

jthegedus:
Sorry for the lacks PR title, GitHub went down yesterday right when I had time to work on OSS so I didn't get to it :sweat_smile:

docs/core-manage-plugins.md
jthegedus: I swapped the order so the Git URL method is first and clarified the language based on your feedback, thanks.

Stratus3D (approved):

PR #966, opened 2021-06-02
chore: question Issue type & direct to GH Discussions

jthegedus:
# Summary

This will add an option in the New Issue selection (https://github.com/asdf-vm/asdf/issues/new/choose) to select a "question" type which will direct people to our GitHub Discussions

The goal being reduction in Issues that are merely questions/clarifications.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,2 +1,6 @@
 # Force users to use one of our predefined issue templates
 blank_issues_enabled: false
+contact_links:
+  - name: Question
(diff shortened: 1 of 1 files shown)


jthegedus:
Perhaps this should link out to StackOverflow instead? :thinking:

jthegedus:
Merging, we can change to SO later if desired.

jthegedus:
This is the result.

image

Stratus3D:
> Perhaps this should link out to StackOverflow instead? 🤔

it would be nice if we could link to both. Some users may prefer StackOverflow, and I follow asdf questions on both sites.

PR #967, opened 2021-06-08
fix shell config command for Linux,Fish,Pacman

lukad:
# Summary

Fixes the instruction for sourcing asdf for Linux,Fish,Pacman.

It currently says to source the `.sh` file instead of the correct `.fish` one.


jthegedus (approved):
Good catch, thanks :)

PR #971, opened 2021-06-15
chore: add Issue link for questions on SO

jthegedus:
# Summary

As discussed in #966 we should also link out to StackOverflow since we have an active Q&A presence there.

Fixes: #966 

Looks like this (Edited HTML):

image


Stratus3D (commented):
Looks great!

PR #973, opened 2021-06-16
docs: homebrew autoremove after asdf uninstall

DanielKehoe:
The linked guide shows how to remove dependencies installed by Homebrew for asdf. Goes into more detail than the asdf docs.


jthegedus:
Hi @DanielKehoe I appreciate your work on the blog. Doesn't this all boil down to adding

```
brew autoremove
```

to the current steps?

For brevity we would rather just include all instructions here and not link to external sources.

DanielKehoe:
Yes, for brevity it can be reduced to `brew autoremove`. I like to give diagnostics and some explanations to give inexperienced users help if they need it. Not a problem if you want to keep external links to a minimum. They can always Google :-)

jthegedus:
Add the `brew autoremove`. Thanks for suggesting this change.

PR #987, opened 2021-07-07
ci: automated release workflow

jthegedus:
# Summary

Automated release **PRs** created by [`release-please`](https://github.com/google-github-actions/release-please-action) which uses [`conventional commits`](https://www.conventionalcommits.org/en/v1.0.0/) to determine the semver release version. **Merge the PR to create the release!**

This removes all manual steps to release making it highly repeatable, actionable by any maintainer, and actionable without a workstation or git (I can merge from my phone while out and about).

A `release-please` PR will:
- look at git history since last release
- analyze commit messages
- determine the semver change
- update the changelog
- update the version number
- if new commits are made to the default branch, the PR is updated accordingly, recalculating the above
- on merge, creates:
  - the Git Tag
  - the GitHub Release

Closes: #942

## Details

- `.github/workflows/release.yml`: release-please workflow which creates release PRs
- `.github/workflows/semantic-pr.yml`: enforces PR titles (since the title is used on squash merge) are Conventional Commit compliant
- `.github/workflows/docs.yml`: updates the asdf version in the install instructions when a new release is published (replacing `release/tag.sh` functionality). *This workflow will likely be superseded by our static docs site generator once I get to that.*
- replace `VERSION` file with `version.txt` as release-please currently requires use of this file

### Release PRs

A release PR looks like this:

image

This shows the new additions to the Changelog. This image is taken from my fork - https://github.com/jthegedus/asdf/pull/1

image

Changelog PR numbers and SHAs are actual links too!

### Documentation Version Update

I ran the workflow on my fork to test:

image

- GitHub Action execution: https://github.com/jthegedus/asdf/runs/3006470600?check_suite_focus=true
- Result commit: https://github.com/jthegedus/asdf/commit/3f2fd1e9bec35493b9d673dfff96b5326e1c919b

### What is lost

From what I understand Release Candidates are not possible using [`release-please`](https://github.com/google-github-actions/release-please-action), I may be mistaken. However useful there are other solutions to RCs, our support for `asdf update --head` being one but also enabling feature flags with a toggle command like `asdf --release-candidate` which could set a flag and opt the user into other code paths.

Nonetheless, I believe the benefits of the rest of this system outweigh the loss of RCs.

### Additional info

[...]

PR #990, opened 2021-07-07
chore: rm unused changelog heading

jthegedus:
# Summary

Clean up unused changelog heading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Changelog
 
-## 0.8.2-dev
-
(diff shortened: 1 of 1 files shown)


Stratus3D:
The idea was the PR creators or maintainers would add things under the dev heading, and then when the release was tagged the `-dev` suffix would just be removed. Nobody has been updating the changelog so I have to populate it from scratch before I tag a new release.

jthegedus:
Ah okay, well, please review #987 as it resolves this issue automatically

PR #995, opened 2021-07-13
fix: allow plugin callbacks to be in any language

Stratus3D:
The original code invoked all plugin callback scripts with Bash.
While all plugin callback scripts are currently implemented in
Bash it is conceivable that another a better shell language will
emerge and become popular. I dont think mandating Bash is the
right choice anymore. I do think that for nearly all plugins
Bash is the right choice right now and for compatibility
purposes Bash should be encouraged.

Note to self:

* Write unit tests for this.
* See if these changes have any effect on https://github.com/asdf-vm/asdf/pull/817


jthegedus:
Looking good. Would it be worth adding `bash`/`sh` to banned list?

Stratus3D:
> Looking good. Would it be worth adding `bash`/`sh` to banned list?

We're still using bash in one place, and of course we use it in the shebang line because asdf written in Bash, so I don't think we are quite ready for that yet. But possibly soon, assuming we can ignore the shebang lines.

Stratus3D:
@jthegedus I think this PR is ready to be merged. Should I merge it now or wait until the next release?

jthegedus:
I'd just merge it now

jthegedus (approved):
LGTM

Stratus3D:
@jthegedus I overlooked the WIP label. I meant to remove it. The only thing I didn't address here was unit tests, but being that I only removed code in this PR, I figured it wasn't critical. I do want to write basic unit tests for this however. I will do so in a later PR.

PR #996, opened 2021-07-14
fix: ensure shims get created when data dir has spaces

mook-as:
# Summary

If `ASDF_DATA_DIR` has a space in the path (by default, this means `asdf` itself was installed to a path with a space), then the shims will not be created correctly in bash.  This PR fixes the issue.

## Other Information

I have only checked up to the actual executable (in my case, node) running — I do not yet know if there are other issues related to spaces in the path.

I'm using extra variables for the lists (which contain newlines) because POSIX shell wouldn't support `while read … done; while < <(command)` (and is caught by the tests).

### DIFF
--- a/lib/commands/command-reshim.bash
+++ b/lib/commands/command-reshim.bash
@@ -103,7 +103,9 @@ generate_shim_for_executable() {
 generate_shims_for_version() {
   local plugin_name=$1
   local full_version=$2
-  for executable_path in $(plugin_executables "$plugin_name" "$full_version"); do
+  local all_executable_paths
+  IFS=$'\n' read -rd '' -a all_executable_paths <<<"$(plugin_executables "$plugin_name" "$full_version")"
+  for executable_path in "${all_executable_paths[@]}"; do
     write_shim_script "$plugin_name" "$full_version" "$executable_path"
   done
 }
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -576,7 +576,9 @@ with_plugin_env() {
 plugin_executables() {
   local plugin_name=$1
   local full_version=$2
-  for bin_path in $(list_plugin_exec_paths "$plugin_name" "$full_version"); do
+  local all_bin_paths
+  IFS=$'\n' read -rd '' -a all_bin_paths <<<"$(list_plugin_exec_paths "$plugin_name" "$full_version")"
+  for bin_path in "${all_bin_paths[@]}"; do
     for executable_file in "$bin_path"/*; do
       if is_executable "$executable_file"; then
         printf "%s\\n" "$executable_file"
--- a/test/reshim_command.bats
+++ b/test/reshim_command.bats
@@ -3,11 +3,11 @@
 load test_helpers
 
 setup() {
-  setup_asdf_dir
+  ASDF_BATS_SPACE_IN_PATH=true setup_asdf_dir
   install_dummy_plugin
 
-  PROJECT_DIR=$HOME/project
-  mkdir $PROJECT_DIR
+  PROJECT_DIR="$HOME/project"
+  mkdir "$PROJECT_DIR"
 }
 
 teardown() {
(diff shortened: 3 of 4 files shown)


lib/commands/command-reshim.bash
```diff
@@ -103,9 +103,12 @@ generate_shim_for_executable() {
 generate_shims_for_version() {
   local plugin_name=$1
   local full_version=$2
-  for executable_path in $(plugin_executables "$plugin_name" "$full_version"); do
+  local executable_path
+  local all_plugin_executables
+  all_plugin_executables="$(plugin_executables "$plugin_name" "$full_version")"
```
Stratus3D, line 108: Aren't these executable paths delimited by newlines?

Couldn't we just use the existing code with `IFS=$'\n'` set?

lib/utils.bash
```diff
@@ -576,13 +576,15 @@ with_plugin_env() {
 plugin_executables() {
   local plugin_name=$1
   local full_version=$2
-  for bin_path in $(list_plugin_exec_paths "$plugin_name" "$full_version"); do
+  local bin_path all_bin_paths
+  all_bin_paths="$(list_plugin_exec_paths "$plugin_name" "$full_version")"
+  while read -r bin_path; do
```
Stratus3D, line 581: The same thing could be done here as well. All we need to do is read `all_bin_paths` line by line since paths are delimited by newlines.

Stratus3D:
Thanks for the PR @mook-as ! Thank you for catching this, it's definitely something we need to get fixed. Do you think you'd be able to write a basic unit tests for these two functions to prevent this bug from being introduced again in the future?

lib/commands/command-reshim.bash
mook-as: We can't, because `for` is a reserved word and not a real command; I've changed this to use the form taken [elsewhere in `utils.bash`](https://github.com/asdf-vm/asdf/blob/60e0e05a9f289e10514890a905c05f8573eb2b7c/lib/utils.bash#L692) instead, though.

mook-as (commented):
Thanks for the fast review!

I've made changes to make it match existing code better, as well changed the tests to just default to having a space in `ASDF_DIR` to catch any issues.  This _does_ involve changing a lot of the tests to quote things properly; sorry that the PR is so invasive.

I couldn't figure out a nice way to fix `shim_exec.bats`, though, so I've opted that one out of having spaces for now.  This is because it tests having a tool version of `path:…`, but it's unclear from the [documentation](https://asdf-vm.com/#/core-configuration?id=tool-versions) what should happen if there's a space in the path, given that it's a file with space-delimited fields.

lib/commands/command-reshim.bash
jthegedus: @mook-as to clarify, is `for` a reserved word in POSIX shells?

mook-as: I didn't know this and had to look it up, but yes, `for` is a [reserved word](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_04) for POSIX shells.  See also the [definition of a for loop](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_04_030).

I did know that it wasn't working with `FOO=bar for …`, at least :)

jthegedus:
> This _does_ involve changing a lot of the tests to quote things properly; sorry that the PR is so invasive.

Is there a way to add specific test cases for the code paths you modified to support the space, leaving the large existing set of tests for without a space in place?

mook-as:
Okay, I've changed things so that having spaces in the test is opt in (per test setup), and only do it for the `reshim` tests. Would that be better? I still fixed all the test within the one file, but it's not quite as big (and I can start doing follow up PRs _slowly_ until the whole thing is covered).

jthegedus:
>Okay, I've changed things so that having spaces in the test is opt in (per test setup), and only do it for the reshim tests. Would that be better? 

Yes, I think so. We don't need to have every single asdf test cover both cases, just as long as the modified codepaths in

```
lib/commands/command-reshim.bash 
lib/utils.bash 
```

are covered with the existing cases and your new cases.

Stratus3D:
> > This _does_ involve changing a lot of the tests to quote things properly; sorry that the PR is so invasive.
> 
> Is there a way to add specific test cases for the code paths you modified to support the space, leaving the large existing set of tests for without a space in place?

I had the exact same thought as @jthegedus.

Checking the updated code now.

Stratus3D (approved):

jthegedus:
@mook-as are you happy with this to be merged?

mook-as:
@jthegedus I am; I can't figure out what's wrong with the WSL tests, though.

jthegedus:
They do not work. We do not expect them to either. It is my fault they are there. I will be removing those test cases shortly. We do want to test in other envs like wsl2 eventually.

PR #997, opened 2021-07-14
chore: migrate to issue forms

jthegedus:
# Summary

GitHub has recently release [Issue Forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/) which expand on the capabilities on Issues allowing the definition of custom web forms with required fields. Full spec available here - https://gh-community.github.io/issue-template-feedback/structured/

This updates our Bug & Feature templates.


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,60 @@
...
+      render: shell
+    validations:
+      required: true
+  - type: textarea
+    id: plugins
+    attributes:
+      label: asdf plugins affected (if relevant)
+      description: The plugins you list here should appear in the list from `asdf info`
+  - type: textarea
+    id: additional-information
```
Stratus3D, line 57: Do we really need this field? It seems like the user could type everything in the description field.

.github/ISSUE_TEMPLATE/feature_request.yaml
```diff
@@ -0,0 +1,54 @@
...
+      required: true
+  - type: textarea
+    id: workarounds
+    attributes:
+      label: "Describe other workarounds you've considered"
+      description: "A clear and concise description of any alternative solutions or features you've considered."
+    validations:
+      required: true
+  - type: dropdown
+    id: importance
```
Stratus3D, line 40: Do you think this information will be valuable to us? A lot of users tend to think there feature request is more important than it really is.

.github/ISSUE_TEMPLATE/feature_request.yaml
```diff
@@ -0,0 +1,54 @@
...
+      label: Importance
+      description: How important is this feature to you?
+      options:
+        - nice to have
+        - would make my life easier
+        - i cannot use asdf without it
+    validations:
+      required: true
+  - type: textarea
+    id: additional-context
```
Stratus3D, line 51: Same thought as for the other template. All the context needed should go in the problem field.

Stratus3D (approved):

.github/ISSUE_TEMPLATE/feature_request.yaml
jthegedus: I agree most people do overestimate the importance. I think it would be nice to know whether the lack of the feature is stopping someone using `asdf` entirely or just a quality of life issue. Most quality of life issues we generally recommend people script around, but knowing the difference in the request upfront may be helpful.

.github/ISSUE_TEMPLATE/feature_request.yaml
jthegedus: Not really, I was just copying over the existing templates

.github/ISSUE_TEMPLATE/bug_report.yaml
jthegedus: Not really, I was just copying over the existing templates

PR #1000, opened 2021-07-19
docs: rewrite using Vuepress

jthegedus:
# Summary

Rewrite the documentation site using a different tool to generate no/disabled-JS compatible solution.

- Live Demo: https://jthegedus.github.io/asdf/
	- disable JS and test as well, all but search & light/dark toggle works as expected
- Build output raw files: https://github.com/jthegedus/asdf/tree/gh-pages
- Built with Vuepress. docs: https://v2.vuepress.vuejs.org/

**Todo**:

- [x] replace CNAME file. It was ommitted for deploying to testing env.
- [x] fix: `CNAME` and `.nojekyll` are in source dir not hosted dir. `.nojekyll` is probably not needed as it's not used in the demo. Need to update how `CNAME` is handled
  - removed `.nojekyll` as demo did not require it.
  - `docs/CNAME` is copied to the site dist dir on build and therefore deployed to the GitHub Pages source location and should be read from the dist location by the GitHub Pages config
- [ ] Post-merge:
	- [ ] update GitHub Pages config. This may mean the docs site is temporarily down while that is configured. See in screenshot section at end of PR for image of required config.
	- [ ] enable branch protection for `gh-pages` branch to ensure it cannot be deleted accidentally or pushed to by anyone other than the bot.
	- [ ] validate that the `.vuepress/config.js` `base:"/"` config is correct and that it doesn't need to change to `base: "/asdf/"`.
 
**Content Changes**:

- better landing page
- Introduction - which explains asdf and how it works and stacks up to similar tools
- Getting Started guide - which walks through the installation clearly
- existing pages have been slightly tweaked. Check the [Reference/Manage](https://jthegedus.github.io/asdf/manage/core.html) section
- dedicated [plugin section](https://jthegedus.github.io/asdf/plugins/create.html). To be expanded upon re-write of plugin authorship. Currently adds links to all the popular plugin places
- [contribution section](https://jthegedus.github.io/asdf/contribute/core.html) with updated guides for contributing to each aspect of `asdf`
- [FAQ](https://jthegedus.github.io/asdf/learn-more/faq.html) with links to Issues/Discussions/SO

To replace the Dropdown selection menus for a JavaScript-free solution we have opted for expandable HTML details tags for each option. See this in the [add to your shell](https://jthegedus.github.io/asdf/guide/getting-started.html#_3-add-to-your-shell) section of the Getting Started guide or the [core uninstall](https://jthegedus.github.io/asdf/manage/core.html#uninstall) section.

Fixes: #862

## Feedback

I really think Vuepress is the write choice here. I would urge not to get stuck on the specific tool used here but what it allows. Most files here are just Markdown as with our previous solution, with Vuepress Config, navbar and sidebars all being JS objects in JS files used during dev & build time (not shipped).

I would like to draw specific attention to these pages for content review:

- [Introduction](https://jthegedus.github.io/asdf/guide/introduction.html) 
- [Getting Started](https://jthegedus.github.io/asdf/guide/getting-started.html) 

Collapsible HTML details elements and Markdown tables have been used to show information in a vertically compact manner reducing clutter when a user is only wanting one of piece of information for their system. Grepping a table is less cluttered than scrolling endlessly through unneeded/unuseful content.

## Other Information

**Why not Jekyll?**

I tried using the proposed Jekyll theme in #862 but was let down by a few things. Firstly, no support for i18n out of the box. I tried using the two most common Jekyll plugins for i18n with the theme but could not get them to work properly without significant effort, which goes against the purpose of these tools. The light/dark mode in the theme is still very early in support. It seemed, to rectify most of these issues I was going to have to customise the theme heavily, which again goes against the purpose of using a theme. And while the user base seemed large for this theme, there appears to be a lack of support with a slow bug fix/release cycle.

**Why Vuepress?**

Vuepress is a static-site generator, the fact it supports writing pages in Vue.js is not why it was chosen. Of all options tested, Vuepress came with the most sane defaults. Importantly, it enables us to just write Markdown. We do not have to deal with templates or content processing pipelines as some of the other suggested solutions require.

Vuepress also has an extremely large user base. This means it has bugs fixed and released in a timely manner. It also means it has:

- internationalization
- mobile responsive theme
- light & dark mode
- naive search
- JavaScript-free fallback

all built in to the core or default theme. This is unlike all other solutions which relegated these features to plugins of various maintenance or age.

**Why Vuepress 2?**

Vuepress 1 is no longer supported with bug fixes and updates suspended (as far as I can tell). Vuepress 2 is almost feature complete (and out of beta) with one item remaining on their Roadmap for features we do not use. Most of the difference between v1 & v2 are in the internal tooling, with minor changes to the default theme or API.

**Caveats to Vuepress**

We must now host the built files for GitHub Pages as it only builds Jekyll sites from source. This is unfortunate for us as it means the `gh-pages` branch will be cloned when performing a full repo clone. This is somewhat mitigated by the GitHub Action which builds the site overwriting those files so there is only ever one copy of the build outputs in the `gh-pages` branch. See the built files here - https://github.com/jthegedus/asdf/tree/gh-pages

This is a small price to pay for the out of the box experience of Vuepress.

Unfortunately this also adds a `package-lock.json` file to the `docs` dir so makes the diff line count huge!

**Other Vuepress info**

While outputting HTML, when enabled, Vuepress will load minimal JavaScript to init a router to enable SPA style routing with transitions etc. This does not affect initial page load and is not required. When JS is disabled you have a standard static HTML site experience.

<details>
<summary>Screenshots</summary>

Desktop:

image
---

Mobile:

image
---

GitHub Pages config update required after merge:

image

</details>


Lewiscowles1986:
I think it's good for me as my primary ask was dark mode.

Stratus3D:
Sorry for the late review @jthegedus. The new website looks great! I like the layout and worked great for me without enabling JS. I am reviewing the code now.

Stratus3D:
> We must now host the built files for GitHub Pages as it only builds Jekyll sites from source. This is unfortunate for us as it means the gh-pages branch will be cloned when performing a full repo clone. 

Most users as they clone asdf with the `--branch` flag so this shouldn't be an issue for most users of asdf right?

CONTRIBUTING.md
```diff
@@ -1,43 +1,39 @@
-# How to Contribute to asdf
+# How to Contribute to `asdf`
```
Stratus3D, line 1: Why is asdf in backticks? This seems inconsistent with some exist content on the documentation site.

docs/.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 16.4.2
```
Stratus3D, line 1: Using asdf to build asdf.

docs/.vuepress/config.js
```diff
@@ -0,0 +1,55 @@
+const navbar = require("./navbar");
+const sidebar = require("./sidebar");
+
+module.exports = {
+  base: "/",
+  head: [],
+  locales: {
+    "/": {
+      lang: "en-US",
+      title: "asdf",
```
Stratus3D, line 10: Is this the page title? If so it could probably be longer. e.g. `asdf - Manage multiple runtime versions with a single CLI tool`

docs/contribute/core.md
```diff
@@ -0,0 +1,111 @@
...
+git clone https://github.com/asdf-vm/asdf.git
+```
+
+The tools for core development are managed with `asdf` in this repo's `.tool-versions`. Add the plugins with:
+
+```shell:no-line-numbers
+asdf plugin add bats https://github.com/timgluz/asdf-bats.git
+asdf plugin add shellcheck https://github.com/luizm/asdf-shellcheck.git
+asdf plugin add shfmt https://github.com/luizm/asdf-shfmt.git
+```
```
Stratus3D, line 22: It might be good to note that these tools don't have to be installed via asdf. And if you are new to asdf and going to be breaking things there is an advantage to not using asdf-installed tools to work on asdf.

docs/contribute/core.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Test: all tests
+bats test/
+# Test: for specific command
+bats test/list_commands.bash
+```
+
+::: tip
+
+**Add tests!** - Please add new tests to cover new code paths before you create a Pull Request. See [bats-core documentation](https://bats-core.readthedocs.io/en/stable/index.html)
```
Stratus3D, line 55: Tests are **required** for new features and speed up review of bug fixes.

docs/contribute/documentation.md
```diff
@@ -0,0 +1,119 @@
...
+
+Format the code before committing:
+
+```shell:no-line-numbers
+npm run format
+```
+
+## Pull Requests, Releases & Conventional Commits
+
+`asdf` is using an automated release pipeline which relies on Conventional Commits in PR titls. Detailed documentation found in the [core contribution guide](./core.md).
```
Stratus3D, line 58: `titls` => `titles`

docs/manage/core.md
```diff
@@ -0,0 +1,338 @@
+# Core
+
+The core `asdf` command list is rather small, but can facilitate many workflows.
+
+## Installation & Setup
```
Stratus3D, line 5: Is this heading even needed on this page?

docs/guide/introduction.md
```diff
@@ -0,0 +1,62 @@
+# Introduction
+
+`asdf` is a tool version manager. All tool version definitions are contained within one file (`.tool-versions`) which you can check-in to your project's Git repository to share with your team, ensuring everyone is using the **exact** same versions of tools.
```
Stratus3D, line 3: I like the fact that you've mentioned the `.tool-versions` file in the first paragraph. It is just as much an API as the CLI commands are, but it isn't prominently documented on our existing site.

docs/guide/introduction.md
```diff
@@ -0,0 +1,62 @@
...
+
+`asdf` was built to address two core issues:
+
+1. different CLI tools to manage each runtime/tool in a repository/project
+2. different CLI commands & configuration files (`.*-version` files) per manager for each of the runtime/tools
```
Stratus3D, line 8: To me this is just one issue. If I had to sum this up I'd say something like this:

The old way of working required multiple CLI version managers, each with their distinct API, implementation (e.g. `$PATH` manipulation, shims, environment variables, etc...) and configuration files. asdf provides a single interface and configuration file to simplify development workflows, and can be extended to all tools and runtimes via a simple plugin interface.

What do you think?

Stratus3D (commented):
I left a few comments, but overall I think this looks great! One thing I would say is that the simpler we can keep things, the less chance there will be to confuse people. I noticed a couple places were it seemed like there were headings or sentences that were similar to those on other pages.

Also, why was the Português translation deleted?

jthegedus:
> > We must now host the built files for GitHub Pages as it only builds Jekyll sites from source. This is unfortunate for us as it means the gh-pages branch will be cloned when performing a full repo clone.
> 
> Most users as they clone asdf with the `--branch` flag so this shouldn't be an issue for most users of asdf right?

Correct, using `--branch` to perform initial checkout will not grab the `gh-pages` branch.

---

>I left a few comments, but overall I think this looks great! One thing I would say is that the simpler we can keep things, the less chance there will be to confuse people. I noticed a couple places were it seemed like there were headings or sentences that were similar to those on other pages.

I agree simple is best. A significant issue with the existing docs is that people do not follow through the setup process completely as it is not clear where to stop, it is cluttered on the initial page and then spreads across many pages. I added the `/guide/getting-started.html` page to be a single page to cover the basic setup end to end. This **requires** duplication as want to focus on the mimimum required, explain how/why for first time readers and explain the basic caveats, while also having a complete reference with more detail (each of the `/manage/*.html` pages) for repeat/advanced users.

To improve for beginners requires some level of duplication. Though I tried to keep it to a minimum.

---

>Also, why was the Português translation deleted?

Because I restructured, rewrote some pages, wrote new pages and cannot perform the translation myself, I thought it best to leave them off until @Rogerio-Viana had a chance to fill in the gaps and launch the pt-br translations when complete instead of with gaps.

I can certainly add the existing pages as best I can and leave a comment like "we've recently added some new pages, if you would like to help translate please see [link to docs]". Thoughts?

---

Thanks for taking the time to review thoroughly. I will spend some time addressing/implementing your more particular suggestions.

CONTRIBUTING.md
jthegedus: I removed them from this page. I tend to ref `asdf` the name in backticks in most places. I have done so on many pages throughout the docs. Perhaps we can discuss the correct usage after the new site is live.

docs/.tool-versions
jthegedus: Yes! I have this nested as to not require `nodejs` when developing the core code.

docs/guide/introduction.md
jthegedus: That's much better, thanks. I found this introduction page the most difficult to write.

docs/guide/introduction.md
jthegedus: Great, hopefully with this here and the guide people will find it more clear it needs to be set for `asdf` to work.

docs/manage/core.md
jthegedus: Not really. Though "Uninstall" is on this page as well, so it made sense to bookend the "core" section with the Install / Uninstall. Technically neither Install or Uninstall should live here, but they don't belong anywhere else more than here and creating a page just for Install/Uninstall didn't seem right or fit anywhere. Update is here because it is a core command, but it only is if you installed with Git.

docs/.vuepress/config.js
jthegedus: From the Vuepress docs
>This will be the suffix for all page titles, and displayed in the navbar in the default theme.

This is what the social preview looks like: https://socialsharepreview.com/?url=https://jthegedus.github.io/asdf/

image

Vuepress doesn't support further customisation regarding this.

docs/guide/introduction.md
jthegedus: I updated with your wording but with the "implementation" and parens as the last item in the first sentence.

CONTRIBUTING.md
Stratus3D: I guess I prefer asdf when referring to the whole project, and only using `asdf` when specifically referring to the command-line executable.

docs/guide/introduction.md
Stratus3D: > That's much better, thanks. I found this introduction page the most difficult to write.

Yeah, I've spent too much time on asdf to be good at describing it to people who aren't familiar with it. This introduce page looks great thought!

Stratus3D:
> I can certainly add the existing pages as best I can and leave a comment like "we've recently added some new pages, if you would like to help translate please see [link to docs]" on the to be translated pages. Thoughts @Stratus3D & @Rogerio-Viana ?

I can go either way, but I guess I'd rather have older pages present even if they don't match the english pages. Either way I won't hold up this PR. I think it's a big improvement and I'm really happy we now have a documentation site that doesn't rely on JS for everything.

Stratus3D (approved):

jthegedus:
I will merge this in, complete the "post-merge" list at the top of the PR, then create a new PR with the translations from @Rogerio-Viana as a starting point, then people can add to them over time.

docs/manage/core.md
Stratus3D: One downside to this is that it moves this page up in the Google search results when someone searches for installation and setup instructions.

Stratus3D:
> I agree simple is best. A significant issue with the existing docs is that people do not follow through the setup process completely as it is not clear where to stop, it is cluttered on the initial page and then spreads across many pages. I added the /guide/getting-started.html page to be a single page to cover the basic setup end to end. This requires duplication as want to focus on the mimimum required, explain how/why for first time readers and explain the basic caveats, while also having a complete reference with more detail (each of the /manage/*.html pages) for repeat/advanced users.

That makes sense. Although I think it might just be better to have an install page, as "install" is more specific than "getting started" and is more tightly scoped to just installation. Any other "getting started" info can be linked to at the bottom of the install page. Another benefit to having the page named "install" is that some users already use asdf and just want to get it setup on a new computer or OS and just want the install commands for their new system.

jthegedus:
> That makes sense. Although I think it might just be better to have an install page, as "install" is more specific than "getting started" and is more tightly scoped to just installation. Any other "getting started" info can be linked to at the bottom of the install page. Another benefit to having the page named "install" is that some users already use asdf and just want to get it setup on a new computer or OS and just want the install commands for their new system.

I am confused.

Getting Started is meant as a guide for users to setup a machine to get `asdf` working as expected (addressing the prior docs issue of people bailing before stepping through 3 pages to get `asdf` in a functional state).

Are you suggesting splitting out Getting Started into two pages, Install and Getting Started? Or are you suggesting we add a new page entirely just to feature the first two (or three?) sections of Getting Started elsewhere?

I think it's reasonable for us to expect users to either look at the "Getting Started" page for install instructions, or search for "install" which will take them to "Getting Started" which has 4 of it's 6 headings starting with "Install".

Alternately, we could rename the "Getting Started" Navbar link and homepage link to read "Guide" or "Install Guide" or "Setup Guide".

Stratus3D:
Sorry for the confusion @jthegedus. I'm not sure how exactly the content should be broken up among pages, but I find  "Install" be a more specific than "Getting Started" and I typically look for "install" instructions rather than "Getting Started" when I know I want to try out a new piece of software.

The first half of our "Getting Started" page is just the install instructions for asdf. And the second half of the page goes into installing a plugin, installing a version, setting a version etc... We could split the page into two pages, or we can keep it as it is. There are however several things I find misleading the page.

1. "2. Install asdf" This heading isn't correct, as it only explains how to download asdf. After this step is complete the user will not have successful installed asdf.
2. "3. Add to your Shell" heading is also confusion because sounds like adding asdf to the shell is some sort of optional step outside of the installation.
3. "Setup Complete!" heading is also confusing because setup was actually complete after "3. Add to your Shell" section but the user is not informed that installation has actually been completed.

I suggest at least the following changes:

* Change "2. Install asdf" heading to "2. Download asdf"
* Change "3. Add to your Shell"  heading to something like "3. Install on system" (not sure what a good title would be)
* Either move "Setup Complete!" to after the "3. Add to your Shell" section, or add a "Installation Complete!" section immediately after the "3. Add to your Shell" section.

jthegedus:
>The first half of our "Getting Started" page is just the install instructions for asdf. And the second half of the page goes into installing a plugin, installing a version, setting a version etc... We could split the page into two pages, or we can keep it as it is.

I agree that the Getting Started page is two different sections in one, but I think they're two halves of the same coin so should remain as a single page

Re the rest, I agree with your suggested changes.

> find "Install" be a more specific than "Getting Started" and I typically look for "install" instructions rather than "Getting Started" when I know I want to try out a new piece of software.

I will rename "getting started" to "install guide" and see how it feels.

docs/guide/getting-started.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Setup Complete!
+
+That completes the initial setup of `asdf` :tada: You can now manage `nodejs` versions for your project. Follow similar steps for each type of tool in your project!
+
+`asdf` has a many more commands to become familiar with, you can see them all by running `asdf --help` or `asdf`. The core of the commands are broken into three categories:
+
+- [core `asdf`](/manage/core.md)
+- [plugins](/manage/plugins.md)
+- [versions (of tools)](/manage/versions.md)
```
MichaelDimmitt, line 295: for example this link no longer works on github

but does work for the site
https://github.com/asdf-vm/asdf/blob/c98d306a9ee8356a7bb7c61048decec2387f1836/docs/guide/getting-started.md#setup-complete

PR #1008, opened 2021-07-23
fix: don't override existing ASDF_DIR

ericbn:
# Summary

Allow users to define their own value of ASDF_DIR before sourcing asdf.sh, e.g. when it was installed with homebrew.

For example, I can use this in my shell startup script:

```
ASDF_DIR=/usr/local/opt/asdf
source "$ASDF_DIR/asdf.sh"
```

Also, similar test for `[ -z "$ASDF_DIR" ]` is already being done in `lib/utils.bash`.

## Other Information

Also unset locally used variables at the end.

### DIFF
--- a/test/version_commands.bats
+++ b/test/version_commands.bats
@@ -21,6 +21,9 @@ setup() {
   mkdir -p $CHILD_DIR
 
   cd $PROJECT_DIR
+
+  # asdf lib needed to run asdf.sh
+  cp -rf $BATS_TEST_DIRNAME/../{bin,lib} $ASDF_DIR/
 }
 
 teardown() {
--- a/asdf.sh
+++ b/asdf.sh
@@ -3,16 +3,18 @@
 # For Bash, ${BASH_SOURCE[0]} will be used to obtain this script's path.
 # For Zsh and others, $0 (the path to the shell or script) will be used.
 _under="$_"
-if [ "${BASH_SOURCE[0]}" != "" ]; then
-  current_script_path="${BASH_SOURCE[0]}"
-elif [[ "$_under" == *".sh" ]]; then
-  current_script_path="$_under"
-else
-  current_script_path="$0"
-fi
+if [ -z "$ASDF_DIR" ]; then
+  if [ -n "${BASH_SOURCE[0]}" ]; then
+    current_script_path="${BASH_SOURCE[0]}"
+  elif [[ "$_under" == *".sh" ]]; then
+    current_script_path="$_under"
+  else
+    current_script_path="$0"
+  fi
 
+  ASDF_DIR="$(dirname "$current_script_path")"
+fi
 export ASDF_DIR
-ASDF_DIR="$(dirname "$current_script_path")"
 # shellcheck disable=SC2016
 [ -d "$ASDF_DIR" ] || echo '$ASDF_DIR is not a directory'
 
@@ -31,3 +33,5 @@ PATH="${ASDF_USER_SHIMS}:$PATH"
 # shellcheck source=lib/asdf.sh
 # Load the asdf wrapper function
 . "${ASDF_DIR}/lib/asdf.sh"
+
+unset _under current_script_path ASDF_BIN ASDF_USER_SHIMS


jthegedus:
@ericbn is this related to or independent of https://github.com/Homebrew/homebrew-core/pull/81664?

ericbn:
Hi @jthegedus. The changes here still don't fix the problems I've raised in that PR in homebrew-core. So, still unrelated in this sense.

ericbn:
@jthegedus this change allows users to set `ASDF_DIR` before sourcing `asdf.sh`, as this is one of the environment variables mentioned in https://asdf-vm.com/manage/configuration.html#environment-variables

jthegedus:
@Stratus3D this change looks good to me. What are your thoughts?

test/version_commands.bats
```diff
@@ -21,6 +21,9 @@ setup() {
   mkdir -p $CHILD_DIR
 
   cd $PROJECT_DIR
+
+  # asdf lib needed to run asdf.sh
+  cp -rf $BATS_TEST_DIRNAME/../{bin,lib} $ASDF_DIR/
```
jthegedus, line 26: Do the tests fail without this change? Which new codepath requires this as I don't see any test changes.

ericbn: Yes, tests fail without this change because `setup_asdf_dir` creates an "empty" ASDF_DIR. It's more like a valid ASDF_DATA_DIR than a valid ASDF_DIR, because the {bin,lib} files are not there. Since with this change the asdf.sh will respect any previously set ASDF_DIR variable, that ASDF_DIR will be used, and the tests fail.

Maybe we should move this `cp -rf $BATS_TEST_DIRNAME/../{bin,lib} $ASDF_DIR/` (which is repeated may times after calling `setup_asdf_dir` in individual tests, so this function guarantees that a complete ASDF_DIR was created. WDYT?

ericbn:
I don't know any fish script, but if you agree with this change, maybe the same should be applied to the `asdf.fish` script, for symmetry.

test/version_commands.bats
jthegedus: I think I understand after looking through some code. `setup_asdf_dir` makes sense, but I question why it would be needed even after your change. As I understand, it will not overwrite `ASDF_DIR` if it is already set. `ASDF_DIR` is set during the `setup_asdf_dir` func, so why would these `version_commands.bats` tests start to fail?

I would expect your change in `asdf.sh` to only affect tests in `asdf_sh.bats`, but I don't think there's a test case for this scenarion, so a test should be added there.

ericbn: The `version_commands.bats` tests fail because:
1. They call `setup_asdf_dir` in the setup step.
    1. `setup_asdf_dir` creates a temp directory with no bin or lib files
    2. and sets the `ASDF_DIR` env variable to be the path to that directory.
2. The tests call `. $(dirname "$BATS_TEST_DIRNAME")/asdf.sh`.
3. `asdf.sh` will use the value of the existing `ASDF_DIR` env variable, that was set by  `setup_asdf_dir`
4. and fail because there's no `lib/asdf.sh` file in that `ASDF_DIR`.

This is what I've mentioned in my previous comments, and that's why I've said that "setup_asdf_dir creates an 'empty' ASDF_DIR", and suggested that it should guarantee that the `ASDF_DIR` it creates is complete.

KlausEverWalkingDev (approved):
Looks fine to me.

Stratus3D (approved):

n-rodriguez:
It seems that this change break something : https://github.com/jbox-web/asdf-squarectl/actions

ericbn:
@n-rodriguez, this works for me:
```
unset ASDF_DIR
git clone --depth 1 --branch master https://github.com/asdf-vm/asdf.git /tmp/.asdf
/tmp/.asdf/bin/asdf plugin-test squarectl https://github.com/jbox-web/asdf-squarectl --asdf-tool-version latest --asdf-plugin-gitref 9967cd047af5fcd74e926bb1abf6d4aecffd90ec squarectl --version
```

Maybe you already have ASDF_DIR set to something else before running `.asdf/bin/asdf`...

Stratus3D:
@n-rodriguez did you figure out what your issue was? I don't see any failures for your CI action.

n-rodriguez:
@Stratus3D sorry I did some clean but I triggered one for you : https://github.com/jbox-web/asdf-squarectl/actions/runs/1426185845

n-rodriguez:
Actually I can't really tell if it's related. The only thing is synchronicity of events.

Stratus3D:
I'm pretty sure it is due to https://github.com/asdf-vm/asdf/commit/69ff2d0c9a4fd273c9dac151345f60f7b146e569. See our discussion here: https://github.com/asdf-vm/asdf/pull/1078 https://github.com/asdf-vm/asdf/issues/1083 and here: https://github.com/asdf-vm/asdf/issues/1083, and here: https://github.com/asdf-vm/asdf/pull/1084. Please follow those issues for the resolution of this. Your issue is not caused by this PR.

Issue #789, opened 2020-08-24
Consider regex validation of plugin names


jthegedus:
**Is your feature request related to a problem? Please describe.**
Currently we allow users to specify anything as a plugin name. We should probably validate plugin names against an allowed list of characters, as in the future with tabular output the parsing may be effected by POSIX character classes appearing within plugin names.

- asdf plugin add nodejs : is valid
- asdf plugin add :rocket::champagne::tada: : is invalid

**Describe the solution you'd like**
Simple regex validation of plugin name in [`lib/commands/command-plugin-add.bash`](https://github.com/asdf-vm/asdf/blob/8ca2af3316be03f980598c5dbcb8152b6b56e96a/lib/commands/command-plugin-add.bash#L9)

**Additional context**
`[:space:]` is probably already one character we should not allow

danhper:
I agree this would be a good idea. I think pretty much any printable character except from spaces should be ok. Or should we stick with ascii?

jthegedus:
>I think pretty much any printable character except from spaces should be ok. Or should we stick with ascii?

I had the same thoughts. People will try using unicode and emojis 😆 

ASCII is the safe bet IMO, especially when considering the environments `asdf` would be run in

danhper:
Agreed, as much as I like emojis, asdf-:rocket::champagne::tada: is probably not a great plugin name :joy: 
So I guess the regexp would be something along the lines of `asdf-[a-zA-Z0-9-]` ?
Do we want to allow `.`,  `_`, `$` or other symbols. I kind of feel dashes are enough and we currently don't have any plugin using such symbols.

jthegedus:
I've no idea of what downstream effects specific chars like `$` ir `.` could have in the codebase, but `[a-zA-Z0-9-]` is plenty IMO.

PR #1010, opened 2021-07-25
fix: regex validate plugin names on plugin add cmd

jthegedus:
# Summary

Apply regex validation to plugin names when `asdf plugin add` is used.

Regex is `^[a-zA-Z0-9_-]+$` and applied via `grep --extended-regexp`.

Fixes: #789 

## Other Information

`^[a-zA-Z0-9-]$` was the original discussion. I have added `_` as well as the regex will then cover all popular naming conventions (camelCase, snake_case, kebab-case, PascalCase and UPPER_SNAKE).

We may restrict this regex further in future because of platform independent handling of filesystem paths with varying cases as discussed in https://github.com/asdf-vm/asdf-plugins/issues/28


test/test_helpers.bash
```diff
@@ -46,7 +46,7 @@ install_mock_broken_plugin() {
 install_mock_plugin_repo() {
   local plugin_name=$1
   local location="${BASE_DIR}/repo-${plugin_name}"
-  cp -r "$BATS_TEST_DIRNAME/fixtures/dummy_plugin" "${location}"
```
jthegedus, line 49: The `dummy` plugin was hardcoded into this helper function. It needed to be changed to test a new plugin with different name and subsequently broke the `plugin update` tests for updating to the remote default branch.

jthegedus: nvm, been staring at screens too long

Stratus3D:
Is `grep --extended-regexp` supported on all platforms?

jthegedus:
I am not sure of the portability of `grep --extended-regexp` (`grep -E`, I prefer the clearer extended flag names), but we're already using it in a number of places, so this change doesn't increase the surface area of tools or flags used.

I think we just wait until someone flags `grep -E` as being an issue for their platform before attempting to remove the dependence.

Stratus3D:
@jthegedus sounds good. I didn't realize core was already using `grep -E`.

PR #1026, opened 2021-08-19
docs: Update the link to the Plugin Shortname Index

hiljusti:
The previous link was to a plugin template. Updated to point to a list of plugins. (I think this is the right place?)


jthegedus (approved):
Great catch, thanks!

Issue #574, opened 2019-09-24
Missing Plugin Doesn't Generate a warning


delmendo:
I updated a project. It now includes a tool version for lua.  I haven't installed the lua plugin because I haven't needed it before.  I did an asdf install and it didn't warn me that I was missing the plugin.

### Steps to reproduce

$ cat .tool-versions 
nodejs 10.16.3
python 3.7.3
lua 5.2

$ asdf plugin-list
elixir
erlang
nodejs
python

$ asdf install
nodejs 10.16.3 is already installed
python 3.7.3 is already installed

### Expected behavior
$ asdf install
nodejs 10.16.3 is already installed
python 3.7.3 is already installed
lua  plugin not installed

### Actual behavior

nothing

It doesn't install or warn me about missing plugin

### Environment

**OS**: Linux - Ubuntu 19.04

**asdf version**: v0.7.4

Stratus3D:
Not sure if we should consider this a bug or not, but it is important enough that we should address it as soon as possible. Thank you for reporting.

jthegedus:
Related #968 

As I expressed in #968 I think we should consider this a bug. I think we should at least warn that a plugin is missing, but could be convinced we should hard error.

threkk:
What could be the treatment? If you attempt to install a plugin that is not currently added, it should:

1. Install the missing plugin if available.
2. Give a warning and do nothing.
3. Give an error and stop.

jthegedus:
>     3\. Give an error and stop.

2 or 3.

threkk:
Option 2 is more user friendly: if a plug in is missing, just keep going and fix it later.
Option 3 is more pipeline friendly: if something is wrong, just stop it.

There are good arguments for either of them. Regarding a fix, I have briefly looked into it and I think I can fix it in short time once I have some free time (hopefully soon).

PR #1033, opened 2021-08-29
feat: Displays a warning when a plugin from the tools-version list does not exist

threkk:
# Summary

When calling the install command, asdf tries to look for versions for all the plugins available and installs them. With this change, it will attempt to find versions for all the installed plugins and plugins defined in the `.tool-versions`.

## Other Information

Fixes https://github.com/asdf-vm/asdf/issues/574

### DIFF
--- a/lib/commands/command-install.bash
+++ b/lib/commands/command-install.bash
@@ -71,6 +71,7 @@ install_local_tool_versions() {
   search_path=$(pwd)
 
   local some_tools_installed
+  local some_plugin_not_installed
 
   local tool_versions_path
   tool_versions_path=$(find_tool_versions)
@@ -96,10 +97,15 @@ install_local_tool_versions() {
   if [ -f "$tool_versions_path" ]; then
     tools_file=$(strip_tool_version_comments "$tool_versions_path" | cut -d ' ' -f 1)
     for plugin_name in $tools_file; do
-      if [[ ! "${plugins_installed[*]}" =~ $plugin_name ]]; then
+      if ! printf '%s\n' "${plugins_installed[@]}" | grep -P -q "^$plugin_name\$"; then
         printf "%s plugin is not installed\n" "$plugin_name"
+        some_plugin_not_installed='yes'
       fi
     done
+  fi
+
+  if [ -n "$some_plugin_not_installed" ]; then
+    exit 1
   fi
 
   if [ -n "$plugins_installed" ]; then
--- a/lib/commands/command-install.bash
+++ b/lib/commands/command-install.bash
@@ -62,6 +62,7 @@ install_one_local_tool() {
     exit 1
   fi
 }
+
 install_local_tool_versions() {
   local plugins_path
   plugins_path=$(get_plugin_path)
@@ -71,11 +72,33 @@ install_local_tool_versions() {
 
   local some_tools_installed
 
+  local tool_versions_path
+  tool_versions_path=$(find_tool_versions)
+
+  # Locate all the plugins defined in the versions file.
+  local tools_file
+  if [ -f "$tool_versions_path" ]; then
+    tools_file=$(strip_tool_version_comments "$tool_versions_path" | cut -d ' ' -f 1)
+  fi
+
+  # Locate all the plugins installed in the system
+  local plugins_installed
   if ls "$plugins_path" &>/dev/null; then
     for plugin_path in "$plugins_path"/*; do
       local plugin_name
       plugin_name=$(basename "$plugin_path")
+      plugins_installed="$plugins_installed $plugin_name"
+    done
+    plugins_installed=$(echo "$plugins_installed" | tr " " "\n")
+  fi
 
+  # Combine both lists into one
+  local tools
+  tools="${tools_file[*]} ${plugins_installed[*]}"
+  tools=$(echo $tools | sort | uniq)
+
+  if [ -n "$tools" ]; then
+    for plugin_name in $tools; do
       local plugin_version_and_path
       plugin_version_and_path="$(find_versions "$plugin_name" "$search_path")"
 
--- a/lib/commands/command-install.bash
+++ b/lib/commands/command-install.bash
@@ -75,12 +75,6 @@ install_local_tool_versions() {
   local tool_versions_path
   tool_versions_path=$(find_tool_versions)
 
-  # Locate all the plugins defined in the versions file.
-  local tools_file
-  if [ -f "$tool_versions_path" ]; then
-    tools_file=$(strip_tool_version_comments "$tool_versions_path" | cut -d ' ' -f 1)
-  fi
-
   # Locate all the plugins installed in the system
   local plugins_installed
   if ls "$plugins_path" &>/dev/null; then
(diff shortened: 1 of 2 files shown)


threkk:
btw, is it normal that all the test run successfully but the pipeline still fails?

jthegedus:
If multiple plugins are not present on the system, will this error for the first it finds in the list, with a rerun required to then error on the next missing plugin? Assuming that is true, is there a way we could log/warn of all missing plugins as opposed to the first not found?

threkk:
I have rewritten the plugin installation. This approach preserves the original installation and at the same time it will report plugin's not available. It will also not stop in case there is a missing plugin.

jthegedus:
Great, thanks for the update. Can you write test cases to cover these scenarios to prevent regressions?

threkk:
Sure! I will try to find time this week(end) and come back with an update :)

Stratus3D:
> I have rewritten the plugin installation. This approach preserves the original installation and at the same time it will report plugin's not available. It will also not stop in case there is a missing plugin.

I think we should stop installing when a missing plugin is encountered. Some projects define plugins in the order they must be installed in, because certain plugins depend on others. Proceeding to install subsequent versions after a plugin is not found could cause problems.

lib/commands/command-install.bash
```diff
@@ -71,11 +72,38 @@ install_local_tool_versions() {
...
+    printf "Install plugins first to be able to install tools\\n"
+    exit 1
+  fi
 
+  # Locate all the plugins defined in the versions file.
+  local tools_file
+  if [ -f "$tool_versions_path" ]; then
+    tools_file=$(strip_tool_version_comments "$tool_versions_path" | cut -d ' ' -f 1)
+    for plugin_name in $tools_file; do
+      if [[ ! "${plugins_installed[*]}" =~ $plugin_name ]]; then
```
Stratus3D, line 99: Is this the same logic we use elsewhere in the codebase? I think this might cause problems if two plugins share matching prefixes. For example, `ag` and `age` are both existing plugins right now. With this logic, `age` could match the plugin name `ag`.

jthegedus:
>I think we should stop installing when a missing plugin is encountered.

Fine with me, but I would like the output to show all missing plugins, as opposed to just the first encountered missing so it is a showstopper only once.

lib/commands/command-install.bash
threkk: No, that is the first occurrence of that comparison. I will turn it into an exact comparison.

threkk:
The changes are then;
- Fix the comparison for the plugin name (`~=` -. `==`)
- Exit if at least one plugin is missing.
- Add unit tests to avoid future regresions.
Do I forget anything?

Stratus3D:
> The changes are then;
> 
>     * Fix the comparison for the plugin name (`~=` -. `==`)
> 
>     * Exit if at least one plugin is missing.
> 
>     * Add unit tests to avoid future regresions.
>       Do I forget anything?

I think that is all. Thanks again for your work on this. @jthegedus do you have anything to add?

jthegedus:
Just for clarity

* Fix the comparison for the plugin name (`~=` -. `==`)
* **Compute list of all missing plugins.**
* Exit if at least one plugin is missing **, warn with the whole list of all missing plugins.**
* Add unit tests to avoid future regresions.

Thanks for tackling our feedback @threkk

threkk:
I made the changes discussed before. The test run green ubuntu, but osx tests look broken in general. Is this a known issue or I need to change the code?

Stratus3D:
> I made the changes discussed before. The tests run successfully in Ubuntu, but OSX tests look broken in general. Is this a known issue or do I need to change the code?

@threkk, not sure about the tests. They are passing on the latest commit of the master branch. You could try rebasing your PR branch on top of latest master, although I'm not sure that will fix it.

Stratus3D:
When I clone down your branch and run it locally (on OSX 10.15.7 (19H1323)) all but two tests pass. The two failures were:

```
 ✗ install_command deletes the download directory
   (in test file test/install_command.bats, line 261)
     `[ ! -d $ASDF_DIR/downloads/dummy/1.1.0 ]' failed
 ✗ asdf update --head should checkout the master branch
   (in test file test/update_command.bats, line 33)
     `[ $(git rev-parse --abbrev-ref HEAD) = "master" ]' failed
   Cloning into '/var/folders/7m/9bq62tmj60zf14d788w0lk8c0000gp/T/asdf.XXXX.4Z9epbdy/home/.asdf'...
```

The first test always fails for me locally (even on master) and the second only failed because I was on your branch and not the master branch when I ran the tests. So I don't see any issues here with your code. Perhaps try re-running the tests on GH or try rebasing onto latest master?

threkk:
I rebased the branch and sadly the issues still persist on OSX. This seems to be some difference between OSX and Ubuntu. I will try to reproduce on a OSX machine to pin point the issue but I can't tell exactly when I will be able. If anyone has any idea about what can be wrong, I will appreciate the help :) I have my suspicions on `grep` behaving differently, but cannot prove it.

threkk:
It took me some time, but finally found the problem: one of the flags was a *nix only flag and was causing the errors. Once removed, everything works as expected :)

jthegedus:
Thanks @threkk !! We will add `grep -P` to our list of banned commands/flags so we don't hit this type of issue again :pray:

jthegedus (approved):
LGTM. Feel free to merge if you're happy @Stratus3D

MPV:
Cool stuff! Would it be possible to cut a new release including this change?

jthegedus:
We will release `0.9.0` once we finish items in the [milestone](https://github.com/asdf-vm/asdf/milestone/2) and fix bugs introduced in `head`

cmur2:
> I think we should stop installing when a missing plugin is encountered. Some projects define plugins in the order they must be installed in, because certain plugins depend on others. Proceeding to install subsequent versions after a plugin is not found could cause problems.

This change breaks my workflow as I intentionally do not install all plugins on all my machines (depending on which I need). Is there any way to disable this feature (looking at the code I guess no) or make it so it can be disabled?

Stratus3D:
@cmur2 I think you make a valid point. Specifically, how do these changes break your current workflow?

We may want to provide a flag or configuration option so that missing plugins do not cause the install command to exit with a status of 1.

cmur2:
@Stratus3D I use a shared dotfiles repo between several of my machines (e.g. work and private ones) to share a `.tool-versions` (symlinked from the dotfiles git checkout to my home directory) among those machines so updating tools is less of a hassle (once instead of once per machine).

For work I might need `kubectl` which I don't use privately so it's in `$HOME/.tool-versions` and on the work machine I've added the asdf plugin for that but not on the private machine. This e.g. saves download bandwidth and disk storage on my private machine.

I also have other tools like `nodejs` which I might need at work and privately so I've added the asdf plugin on both machines.

After the update to asdf-vm v0.9.0, I cannot update nodejs on my private machine anymore since asdf complains about the missing `kubectl` plugin and exits with 1. It'd be cool if asdf could just ignore this state (optionally).

For now I used `sed -i 's/-n "$some_plugin_not_installed"/-n ""/g' "$ASDF_DIR/lib/commands/command-install.bash"` after updating asdf to workaround this.

PR #1035, opened 2021-09-02
fix: incorrect usage of grep

yacchi:
# Summary
Message displayed when executing a command that does not specify a version, and the error message is output at the same time because of incorrect usage of grep.

Extract from macOS man:
> -F, --fixed-strings
>        Interpret pattern as a set of fixed strings (i.e. force grep to behave as fgrep).
> -f file, --file=file
>        Read one or more newline separated patterns from file.  Empty pattern lines match every input line.  Newlines are not considered part of a pattern.  If file is empty, nothing is matched.

Expected:
```shellscript
> go
No version set for command go
Consider adding one of the following versions in your config file at 
go-sdk 1.17
```

Actual(in head branch):
```shellscript
> go
grep: go-sdk 1.17
go-sdk system: No such file or directory
No version set for command go
Consider adding one of the following versions in your config file at 
go-sdk 1.17
```

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -681,7 +681,7 @@ select_from_preset_version() {
   shim_versions=$(get_shim_versions "$shim_name")
   if [ -n "$shim_versions" ]; then
     preset_versions=$(preset_versions "$shim_name")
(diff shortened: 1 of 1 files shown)


Stratus3D (approved):

Stratus3D:
Thanks for the PR @yacchi !

PR #1037, opened 2021-09-03
fix: wait until the plugin update are finished

yacchi:
# Summary

In `asdf plugin update --all`, `wait` is not working and the command exits before the update of each plugin are finished.

Executing `while` after a pipe will run it in a subshell, so you can't use `wait` on the caller to wait for each command to finish.

In such a case, *Process Substitution* is often used, but it seems to be prohibited in this project.
However, for the usage here, I think *Here string* is fine.

### DIFF
--- a/lib/commands/command-plugin-update.bash
+++ b/lib/commands/command-plugin-update.bash
@@ -8,12 +8,14 @@ plugin_update_command() {
 
   local plugin_name="$1"
   local gitref="${2}"
+  local plugins=
 
   if [ "$plugin_name" = "--all" ]; then
     if [ -d "$(asdf_data_dir)"/plugins ]; then
-      find "$(asdf_data_dir)"/plugins -mindepth 1 -maxdepth 1 -type d | while IFS= read -r dir; do
+      plugins=$(find "$(asdf_data_dir)"/plugins -mindepth 1 -maxdepth 1 -type d)
+      while IFS= read -r dir; do
         update_plugin "$(basename "$dir")" "$dir" "$gitref" &
-      done
+      done <<<"$plugins"
       wait
     fi
   else
--- a/test/plugin_update_command.bats
+++ b/test/plugin_update_command.bats
@@ -112,3 +112,10 @@ teardown() {
   [ "$status" -eq 0 ]
   [ -f $ASDF_DIR/shims/dummy ]
 }
+
+@test "asdf plugin-update done for all plugins" {
+  local command="asdf plugin-update --all"
+  # Count the number of update processes remaining after the update command is completed.
+  run bash -c "${command} >/dev/null && ps -o 'ppid,args' | awk '{if(\$1==1 && \$0 ~ /${command}/ ) print}' | wc -l"
+  [[ 0 -eq "$output" ]]
+}


Stratus3D:
Thanks for the PR!

Is there any way we can write a test for this fix to verify we don't introduce this bug at a later time?

yacchi:
@Stratus3D 

I'm currently writing tests in bats, and I found that bats wait for all commands to finish in a different way.
So even if there is a problem with the use of wait, the tests with bats always seem to succeed.

I haven't found a way to work around this problem, do you know of any good information?

If you write such a test code and debug output along with the date and time, you will see that the waiting areas are different.
(To make the example easier to understand, I have added `sleep 2` to `update_plugin`.)

Added test code in plugin_update_command.bats (echo is a debug print for bats)
```bats
@test "asdf plugin-update for all plugins" {
  install_mock_plugin_repo "dummy2"
  run asdf plugin add "dummy2" "${BASE_DIR}/repo-dummy2"
  run asdf plugin-update --all
  echo "# end $(date -u) in $(basename "${BASH_SOURCE[0]}:$LINENO")" >&3
  {
    [[ "$output" = *"Updating dummy to master"* ]]
    [[ "$output" = *"Updating dummy2 to master"* ]]
  } || (
    echo "output = ${output}"
    false
  )
}
```

Add debug output to the code before modification in plugin_update_command.bats:
```shellscript
    if [ -d "$(asdf_data_dir)"/plugins ]; then
      echo "# update start $(date -u) in $(basename "${BASH_SOURCE[0]}:$LINENO")" >&3
      find "$(asdf_data_dir)"/plugins -mindepth 1 -maxdepth 1 -type d | while IFS= read -r dir; do
        update_plugin "$(basename "$dir")" "$dir" "$gitref" &
      done
      echo "# wait start $(date -u) in $(basename "${BASH_SOURCE[0]}:$LINENO")" >&3
      wait
      echo "# wait end $(date -u) in $(basename "${BASH_SOURCE[0]}:$LINENO")" >&3
    fi
  else
```

Output:
```shellscript
> env LANG=C bats test/plugin_update_command.bats -f "all plugins"
    update start Sun Sep  5 23:30:32 UTC 2021 in command-plugin-update.bash:14                 1/1
   wait start Sun Sep  5 23:30:32 UTC 2021 in command-plugin-update.bash:18
   wait end Sun Sep  5 23:30:32 UTC 2021 in command-plugin-update.bash:20
   end Sun Sep  5 23:30:34 UTC 2021 in bats.60856.src:120
 ✓ asdf plugin-update for all plugins

1 test, 0 failures
```

Add debug output to the modified code:
```shellscript
    if [ -d "$(asdf_data_dir)"/plugins ]; then
      echo "# update start $(date -u) in $(basename "${BASH_SOURCE[0]}:$LINENO")" >&3
      local plugins
      plugins=$(find "$(asdf_data_dir)"/plugins -mindepth 1 -maxdepth 1 -type d)
      while IFS= read -r dir; do
        update_plugin "$(basename "$dir")" "$dir" "$gitref" &
      done <<<"$plugins"
      echo "# wait start $(date -u) in $(basename "${BASH_SOURCE[0]}:$LINENO")" >&3
      wait
      echo "# wait end $(date -u) in $(basename "${BASH_SOURCE[0]}:$LINENO")" >&3
    fi
  else
```

Output:
```shellscript
> env LANG=C bats test/plugin_update_command.bats -f "all plugins"
    update start Sun Sep  5 23:38:39 UTC 2021 in command-plugin-update.bash:14                 1/1
   wait start Sun Sep  5 23:38:39 UTC 2021 in command-plugin-update.bash:20
   wait end Sun Sep  5 23:38:41 UTC 2021 in command-plugin-update.bash:22
   end update Sun Sep  5 23:38:41 UTC 2021 in bats.63388.src:120
 ✓ asdf plugin-update for all plugins

1 test, 0 failures
```

Stratus3D:
Thanks for the detailed examples @yacchi! Would it not be possible to do something like this to check if the updates are finished? 

```
[[ "$(jobs | head | cut -f4 -d' ')" -eq "running" ]
```

I'm pretty sure this would not be a fool proof way of checking - the job could always finish after the command exited but before than line of test code ran. But it might be better than nothing if it works.

yacchi:
@Stratus3D 
Thank you for the test example.
I have examined various methods and added the test code.


I found out that the processes that couldn't wait and remained had a Parent PID of 1
This is because the parent process (`asdf plugin-update --all`) exits while the update job for each plugin is running in the background.

The `jobs` command was not able to detect this, but I was able to detect the remaining processes by using the ps command instead.

> I'm pretty sure this would not be a fool proof way of checking - the job could always finish after the command exited but before than line of test code ran. But it might be better than nothing if it works.

It is true that this method cannot detect the problem if all the jobs are terminated before the end of the process list.
However, as far as I have tried, the code before the fix is always an error, so I think it works reasonably well.

Stratus3D (approved):

Stratus3D:
Thanks for the test!

PR #1042, opened 2021-09-08
Use more idiomatic fish

maxnordlund:
# Summary

The use of `.` is deprecated in fish, unlike POSIX shells. We can avoid calling `psub` and pipe directly into source which does the same thing.

## Other Information

See https://fishshell.com/docs/current/cmds/source.html


maxnordlund:
I've added a comment in the test that failed, which forbids `source`, and disabled the check using another comment in the function. I think this is good to go now. 🐟

Stratus3D:
@danhper you are familiar with Fish right? Can you take a look at these changes?

test/banned_commands.bats
```diff
@@ -23,6 +23,7 @@ banned_commands=(
     # Process substitution isn't POSIX compliant and cause trouble
     "<("
     # source isn't POSIX compliant. . behaves the same and is POSIX compliant
+    # Except in fish, where . is deprecated, and will be removed in the future.
     source
```
jthegedus, line 27: This is interesting to know, thanks @maxnordlund Do you have suggestions for how we should handle this in future once `.` is removed in Fish?

maxnordlund: Just keep using `source` I guess? It's only `.` that's deprecated, which might not be that obvious now that I read it again.

jthegedus: Our plan was to deprecate use of `source` internally because `.` is more POSIX compliant. If Fish is dropping `.`, then this will affect all usage of `.` in our codebase.

maxnordlund: Fish isn't a POSIX shell at all, so I think it would be better to keep them separated. But most of asdf is written in bash, right? There it would be fine to enforce `.` over `source`.

If you want to go full POSIX, you could switch to `#!/bin/sh` and see what shellcheck finds. 🤔

jthegedus: I think it is a bit soon for us to go "FUll POSIX"

Stratus3D: @jthegedus we could just ignore `.fish` files in our banned commands script. That would probably be the best way to handle since fish isn't POSIX.

jthegedus: Ah, of course! Yes, sounds like a good solution.

jthegedus:
I think we should merge and those testing with `--head` will let us know if there's issues

Issue #1046, opened 2021-09-12
bug: sed RE error: trailing backslash (\) on FreeBSD


hadret:
### Describe the Bug

`asdf install ruby latest:2.7` doesn't work as expected on FreeBSD, it spits the following error message:

```
sed: 1: "s/^\s\+//": RE error: trailing backslash (\)
```

I traced down the error to the `.asdf/lib/commands/command-latest.bash` and this line in particular:

```
[...]
    sed 's/^[[:space:]]\+//' |
[...]
```

### Steps to Reproduce

Run install for any plugin using latest + : on FreeBSD. Or this:

```
asdf list-all ruby 2.7 | grep -vE "(^Available versions:|-src|-dev|-latest|-stm|[-\\.]rc|-alpha|-beta|[-\\.]pre|-next|(a|b|c)[0-9]+|snapshot|master)" | sed 's/^\s\+//
```

### Expected Behaviour

It should just work like on Linux and macOS.

### Actual Behaviour

Error message: `sed: 1: "s/^\s\+//": RE error: trailing backslash (\)`

### Environment

```shell
OS:
FreeBSD calcifer.io 13.0-RELEASE-p4 FreeBSD 13.0-RELEASE-p4 #0: Tue Aug 24 07:33:27 UTC 2021     root@amd64-builder.daemonology.net:/usr/obj/usr/src/amd64.amd64/sys/GENERIC  a
md64

SHELL:
fish, version 3.3.1

ASDF VERSION:
v0.8.1-a1ef92a

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/home/hadret/.asdf

ASDF INSTALLED PLUGINS:
python                       https://github.com/danhper/asdf-python.git
ruby                         https://github.com/asdf-vm/asdf-ruby.git
```

PR #1047, opened 2021-09-12
fix: sed re error trailing backslash on FreeBSD (#1046).

hadret:
# Summary

This fixes the following error with sed on FreeBSD:

```
sed: 1: "s/^\s\+//": RE error: trailing backslash (\)
```

Fixes: #1046

## Other Information

I confirmed the command still works as expected on Linux and macOS.


Stratus3D:
Thanks for the detailed bug report and PR @hadret ! These changes look good and everything still works for me.

Stratus3D (approved):

Issue #859, opened 2021-02-03
Allow re-installing old/broken tool installations following a plugin update, e.g. via post-plugin-update hook


nirvana-msu:
**Is your feature request related to a problem? Please describe.**
Sometimes you make updates to your plugin that fix something that was broken in the previous installs. In these cases, it would be great to have a way to force re-installation of the old (broken) installs.

**Describe the solution you'd like**
One way to achieve that would be by adding a `post-plugin-update` hook - so that once plugin is updated, it can trigger re-installation of old (broken) installs. One tricky part would be to ensure re-install is only triggered when plugin is updated to a certain revision - e.g. next update may not require re-install. It would have been easy if plugins themselves were versioned, but I suppose it could also be done if we know which git-ref we're upgrading from.

Alternatively, this could be addressed by allowing a plugin to "declare" that a change breaks certain install versions - and then asdf itself could take care of re-installing those versions (if installed).

The hook is more general so can probably cover some other use-cases not described here.

**Describe similar asdf features and why they are not sufficient**
There are already similar hooks `post-plugin-add` and `post-plugin-remove`, but in this case an action needs to be taken on plugin update.

**Describe workarounds you've considered**
At the moment I have to warn users that they need to manually re-install affected tool versions after updating plugin.

Stratus3D:
Hi @nirvana-msu , currently there is no way to indicate that an installed version depends on, or is affected by, a plugin version change. Currently we don't even support plugin versions yet, but we will eventually (see https://github.com/asdf-vm/asdf/issues/166 and related issues for that). As it is currently we don't have a way to support this. Closing this issue now.

PR #1049, opened 2021-09-18
feat: add post update plugin support

yacchi:
# Summary

Currently, asdf does not have a hook that is called when a plugin is updated, so it is not possible for plugins to perform migrations.

For example, if a plugin has a problem and needs to be fixed under "installs", the only way to do this is to have the user do it after installation.

With the post-install-hook, it will be possible to migrate when updating, which will make it easier to develop plug-ins and improve the usability for users.

## Other Information

This may also answer the request of #859 .

### DIFF
--- a/docs/plugins/create.md
+++ b/docs/plugins/create.md
@@ -138,7 +138,7 @@ See also the related hooks:
 
 #### bin/post-plugin-update
 
-This can be used to run any post-plugin-update actions after the plugin has been updated to asdf.
+This can be used to run any post-plugin-update actions after asdf has downloaded the updated plugin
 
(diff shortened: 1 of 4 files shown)


docs/plugins/create.md
```diff
@@ -136,6 +136,19 @@ See also the related hooks:
 - `post_asdf_plugin_add`
 - `post_asdf_plugin_add_${plugin_name}`
 
+#### bin/post-plugin-update
+
+This can be used to run any post-plugin-update actions after the plugin has been updated to asdf.
```
Stratus3D, line 141: I'm not sure the `... to asdf.` ending adds anything meaningful here. I think `This can be used to run any post-plugin-update actions after asdf has downloaded the updated plugin` might be better wording here.

Stratus3D (approved):
These changes look good to me. @jthegedus what do you think?

jthegedus:
Yes, this looks good. A very useful hook.

PR #1052, opened 2021-09-25
docs: fix brew upgrade from latest commit on master command

texpert:
# Summary

Currently, the brew upgrade from latest commit on master command is specified like `brew upgrade asdf --HEAD`, which is incorrect, failing with 

`Error: invalid option: --HEAD`.

The correct option is [--fetch-HEAD](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/upgrade.rb#L57)

Fixes: display the correct command:

`brew upgrade asdf --fetch-HEAD`


KlausEverWalkingDev (approved):
Fine.

jthegedus (approved):
Thanks!

PR #1057, opened 2021-09-27
fix: don't generate on error if backup file doesn't exists

JrCs:
Signed-off-by: JrCs <90z7oey02@sneakemail.com>

# Summary

If the replacement (sed) is not done the backup file isn't create and the rm command return on error.  
With `rm -f` no error is generated even if the file doesn't exists.

### DIFF
--- a/lib/commands/command-reshim.bash
+++ b/lib/commands/command-reshim.bash
@@ -70,7 +70,7 @@ write_shim_script() {
   if [ -f "$shim_path" ]; then
     if ! grep -x "# asdf-plugin: ${plugin_name} ${version}" "$shim_path" >/dev/null; then
       sed -i.bak -e "s/exec /# asdf-plugin: ${plugin_name} ${version}\\"$'\n''exec /' "$shim_path"
-      rm "$shim_path".bak
+      rm -f "$shim_path".bak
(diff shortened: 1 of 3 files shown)


JrCs:
@jthegedus if you call asdf with a bash script that use a `set -u` asdf will stop if it can't remove the backup file.
I don't have another test for this.

Stratus3D:
@JrCs can you share a little more about how you invoked asdf after setting `set -e`? I'm curious how this setting outside of the asdf codebase is affect it.

JrCs:
@Stratus3D easy: do a `set -eu` in your current shell then call asdf.  
For my use case the `set -e` is done by my CI pipeline for the current shell.

Stratus3D:
Hmm... I am unable to reproduce this issue on OSX using Zsh:

```
$ set -e
$ asdf local lua 5.3.3
$ asdf reshim
```

Both commands succeed without any issues with this fix.

This is not a big deal to me, as I'd eventually like to move to having `set -e` set inside asdf, and these changes are one step closer to making that possible. Thanks for the PR @JrCs ! I am merging now.

Stratus3D (approved):

PR #1058, opened 2021-09-27
'asdf current --as-tool-versions' outputs a format suitable for export to `.tool-versions`

teruokun:
# Summary

Add in an option for `asdf current` that outputs a format compatible to use as a `.tool-versions` file.  This is useful for replicating a local configuration across one or more configuration files to be used by another host/environment and also outputs more controlled formats for other tools that may want to query what is currently active.

### DIFF
--- a/lib/commands/command-current.bash
+++ b/lib/commands/command-current.bash
@@ -4,6 +4,7 @@
 plugin_current_command() {
   local plugin_name=$1
   local terminal_format=$2
+  local tool_versions=$3
 
   check_if_plugin_exists "$plugin_name"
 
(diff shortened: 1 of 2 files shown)


offbyone (approved):

jthegedus:
I think this type of change warrants a larger discussion before a PR.

Stratus3D:
Thanks for the PR @teruokun, I do think we need to discuss this before preparing to merge. If you have any relevant issues/discussions related to this can you please link to them? Thanks!

offbyone:
I know for my part, I want the ability to test out tool version setups in a subshell and then persist them to a project configuration. Sort of a "lock" approach to the process.

teruokun:
So, the main reasoning is to allow for easy duplication of the asdf current environment used locally on another host.  As it may be composed of multiple layers, tracking through all of the named configurations and merging them seems like a lot of repeated logic when what we care about is how `asdf` specifically will interpret the configuration.  This also allows for programmatic reading of the configuration in a more-controlled format as there is no consistent delineation of the standard text output when multiple versions are surfaced (for example, `python 3.9.7 3.8.10 3.10.0`) to allow for a named default (e.g. `python`) that can be used side-by-side with version-specific binaries (e.g. `python3.10`).

So in other words, I want to avoid weird behavior around parsing this output when all I intend for is to use it with another `asdf` installation

Stratus3D:
Overall these changes look good to me, and I think this is a nice addition. @jthegedus what do you think?

Stratus3D:
Sorry for the late response here. I think this is a feature worth adding. However asdf has been rewritten in Go (https://github.com/asdf-vm/asdf/releases/tag/v0.16.0) so this PR cannot be merged. I'm going to close it now.

@teruokun if you still want this feature can you open a feature request issue? Thanks!

PR #1066, opened 2021-10-08
feat: Elvish Shell support

elijahr:
# Summary

Add support for [Elvish Shell](https://elv.sh/).


elijahr:
@jthegedus I've reworked this considerably, updated documentation, improved test coverage, and added shell completions. This is ready for review.

jthegedus:
@elijahr I would like another maintainer to review this before merge. Unfortunately we're all rather busy all the time. Thanks for your patience.

.github/workflows/tests.yml
```diff
@@ -28,15 +28,21 @@ jobs:
...
+          sudo add-apt-repository -y ppa:fish-shell/nightly-master
+          sudo add-apt-repository -y ppa:longsleep/golang-backports
           sudo apt-get update
-          sudo apt-get -y install fish
+          sudo apt-get -y install fish golang-go
+
+          # Build elvish and add to path
+          git clone --depth 1 --single-branch https://github.com/elves/elvish
+          cd elvish
+          make get
```
Stratus3D, line 44: Is this building elvish from source? Or is this just downloading a pre-built binary?

asdf.fish
```diff
@@ -2,9 +2,9 @@ set -x ASDF_DIR (dirname (status -f))
 
 set -l asdf_user_shims (
   if test -n "$ASDF_DATA_DIR"
-    echo $ASDF_DATA_DIR/shims
+    printf $ASDF_DATA_DIR/shims
```
Stratus3D, line 5: The first argument to `printf` should be the format string. The format string should not contain variables, only placeholders. For example, in this case `printf "%s\n" "$ASDF_DATA_DIR/shims"` to replace `echo`.

elijahr: Changed in e697aaba1d15ea53efbddb31b382a7df0de379bb

.github/workflows/tests.yml
elijahr: Building from source - didn't take too long locally, but looks like there is a linux binary available, so using that instead. Changed in 25382798d9b2d65a3867f548732a5f625f195d85

jthegedus:
Going to merge this as it is a large change and the longer we leave it the more annoying it will become to merge. We can hotfix `head` with any major issues. Thanks @elijahr

Stratus3D (approved):

Stratus3D:
Thanks @jthegedus and @elijahr ! Looks great!

PR #1084, opened 2021-11-04
fix: Fix plugin-test arg parsing

### LINKED ISSUE #1083: We may have broken plugin test
@tsuyoshicho yeah it does look like we may have broken something here. @raxod502 would you be willing to take a look at @tsuyoshicho builds? Output shows things like:

```
asdf-awscli: Could not download https://github.com/aws/aws-cli/archive/--asdf-plugin-gitref.tar.gz
```

clearly `--asdf-plugin-gitref` was a flag for the tag, not the tag itself.

_Originally posted by @Stratus3D in https://github.com/asdf-vm/asdf/issues/1078#issuecomment-959141878_

raxod502:
Should fix #1083.

### DIFF
--- a/lib/commands/command-plugin-test.bash
+++ b/lib/commands/command-plugin-test.bash
@@ -9,24 +9,26 @@ plugin_test_command() {
   local plugin_gitref="master"
   local tool_version
   local interpret_args_literally
+  local skip_next_arg
 
   for arg; do
     shift
-    if [ -n "${interpret_args_literally}" ]; then
+    if [ -n "${skip_next_arg}" ]; then
+      skip_next_arg=
+    elif [ -n "${interpret_args_literally}" ]; then
       set -- "$@" "${arg}"
     else
       case "${arg}" in
       --asdf-plugin-gitref)
-        plugin_gitref="$2"
-        shift 2
+        plugin_gitref="$1"
+        skip_next_arg=true
         ;;
       --asdf-tool-version)
-        tool_version="$2"
-        shift 2
+        tool_version="$1"
+        skip_next_arg=true
         ;;
       --)
         interpret_args_literally=true
-        shift
         ;;
       *)
         set -- "$@" "${arg}"
--- a/test/fixtures/dummy_plugin/LICENSE
+++ b/test/fixtures/dummy_plugin/LICENSE
@@ -1,0 +1,20 @@
+The MIT License (MIT)
+
+Copyright (c) 2014 Akash Manohar J
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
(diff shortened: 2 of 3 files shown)


jthegedus:
can we capture this scenario as a test for regression purposes?

lib/commands/command-plugin-test.bash
```diff
@@ -9,24 +9,26 @@ plugin_test_command() {
...
       case "${arg}" in
       --asdf-plugin-gitref)
-        plugin_gitref="$2"
-        shift 2
+        plugin_gitref="$1"
+        skip_next_arg=true
         ;;
       --asdf-tool-version)
-        tool_version="$2"
-        shift 2
```
Stratus3D, line 25: I don't understand why these lines didn't work. Can you explain what was wrong with the previous code? New code looks good, but it seems like it should be equivalent to the old code.

raxod502: Well, I feel less guilty for failing to realize the bugs last time :)

Two issues with the old code:
* We `shift` at the beginning of the loop, so although `$1` and `${arg}` are equivalent at the beginning of the loop, they are no longer equivalent after `shift`. Therefore we should be using `$1` instead of `$2`.
* Even if you fix that, modifying the args array in place does not affect the iteration, so even if you `shift` a second time to skip over the option argument, it still shows up as the next value for `${arg}`. Hence the requirement to use a bool to track whether to skip the next argument.

raxod502: You'll notice that all of this is horribly obtuse, this is why writing pure POSIX compatible scripts is not so popular these days!

jthegedus: > You'll notice that all of this is horribly obtuse, this is why writing pure POSIX compatible scripts is not so popular these days!

Indeed 😅

Stratus3D: @raxod502 just curious, where does `arg` get set? I don't see it.

Stratus3D: Would a pattern like this be better? https://github.com/Stratus3D/dotfiles/blob/master/scripts/tools/to_server#L101-L143 (this is a pattern I've repeatedly used because it doesn't require setting any temporary variables. Only `$1`, `$2`, and `shift` are used.

raxod502: * `arg` gets set by the `for arg` at the top of the loop. It's equivalent to `for arg in "$@"`.
* No, unfortunately, we can't do it the reasonable way, because we need to save an array of the command to run, for execution later, and unfortunately POSIX sh only supports having one array variable in your whole program, namely `$@`. So we have to do all of the manipulation in place.

raxod502:
@jthegedus Added a regression test and verified that it fails before the changes from this PR are applied!

The license file is needed because `plugin-test` fails without a license file in the plugin.

Stratus3D (approved):

jthegedus:
Thanks @raxod502

Issue #686, opened 2020-03-28
Verify sed commands are correct


Stratus3D:
*This is issue is so I don't forget to work on this myself, but if anyone else wants to take a stab at this they are welcome to*

### Expected behavior

`sed` find and replace commands in asdf should handle paths with all valid characters. `sed` find and replace commands use delimiter characters, so unless those delimiters are escaped in the input value, the find and replace expression may not be able to handle replace any string in any input string.

See https://unix.stackexchange.com/questions/129059/how-to-ensure-that-string-interpolated-into-sed-substitution-escapes-all-metac for the proper command.

Stratus3D:
Going to try to get this finished in the next couple days.

Stratus3D:
I opened a PR that I'm pretty happy with based on my work today on this. Unfortunately I wasn't able to write a test to validate that our remaining `sed` commands are correct, but as a general rule, `sed` substitutions that contain un-escaped shell variables ought to be avoided and may contain bugs.

PR #1087, opened 2021-11-05
fix: Sed improvements

Stratus3D:
# Summary

* fix: Use bash instead of sed for updating PATH environment variable
* fix: Remove unnecessary ls and sed commands

Fixes: #686

### DIFF
--- a/lib/commands/command-help.bash
+++ b/lib/commands/command-help.bash
@@ -14,20 +14,17 @@ EOF
 }
 
 asdf_extension_cmds() {
-  local plugins_path ext_cmd_path ext_cmds plugin
+  local plugins_path plugin_path ext_cmd_path ext_cmds plugin
   plugins_path="$(get_plugin_path)"
-  # use find instead of ls -1
-  # shellcheck disable=SC2012
-  for plugin in $(ls -1 "$plugins_path" 2>/dev/null | sed "s#^$plugins_path/##"); do
-    ext_cmd_path="$plugins_path/$plugin/lib/commands"
-    ext_cmds=$(
-      ls -1 "$ext_cmd_path"/command*.bash 2>/dev/null |
-        sed "s#$ext_cmd_path/##"
-    )
+  for plugin_path in "$plugins_path"/*; do
+    plugin="$(basename "$plugin_path")"
+    ext_cmd_path="$plugin_path/lib/commands"
+    ext_cmds="$(find "$ext_cmd_path" -name "command*.bash")"
     if [[ -n $ext_cmds ]]; then
       printf "\\nPLUGIN %s\\n" "$plugin"
       for ext_cmd in $ext_cmds; do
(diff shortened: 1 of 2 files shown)


lib/commands/command-help.bash
```diff
@@ -14,20 +14,17 @@ EOF
 }
 
 asdf_extension_cmds() {
-  local plugins_path ext_cmd_path ext_cmds plugin
+  local plugins_path plugin_path ext_cmd_path ext_cmds plugin
   plugins_path="$(get_plugin_path)"
-  # use find instead of ls -1
-  # shellcheck disable=SC2012
-  for plugin in $(ls -1 "$plugins_path" 2>/dev/null | sed "s#^$plugins_path/##"); do
```
Stratus3D, line 21: I was able to remove both `ls` and `sed` from this line and replace the `sed` command with a simpler `basename` command. I'm considering adding `ls` to the banned commands list as I see it used in some anti-patterns in our codebase. To me it seems like listing files is done better with commands like `find`, `grep`, and the occasional globbing.

lib/commands/command-help.bash
```diff
@@ -14,20 +14,17 @@ EOF
...
-    ext_cmd_path="$plugins_path/$plugin/lib/commands"
-    ext_cmds=$(
-      ls -1 "$ext_cmd_path"/command*.bash 2>/dev/null |
-        sed "s#$ext_cmd_path/##"
-    )
```
Stratus3D, line 26: Same thing here. Was able to replace a `ls` and `sed` pipeline with just `basename`.

lib/utils.bash
```diff
@@ -806,3 +806,20 @@ with_shim_executable() {
...
+  printf "%s" "${input//"$find_str"/"$replace"}"
+}
+
+remove_path_from_path() {
+  # A helper function for removing an arbitrary path from the PATH variable.
+  # Output is a new string suitable for assignment to PATH
+  local PATH=$1
+  local path=$2
+  substitute "$PATH" "$path" "" | sed -e "s|::|:|g"
+}
```
Stratus3D, line 825: I noticed we did this in a couple places so I decided to extract this logic out into it's own function.

lib/utils.bash
```diff
@@ -806,3 +806,20 @@ with_shim_executable() {
...
 }
+
+substitute() {
+  # Use Bash substituion rather than sed as it will handle escaping of all
+  # strings for us.
+  local input=$1
+  local find_str=$2
+  local replace=$3
+  printf "%s" "${input//"$find_str"/"$replace"}"
+}
```
Stratus3D, line 817: A bit more code now, but this is much more robust than `sed` because it can handle any character in the three input strings and will handle escaping for us.

jthegedus: Is this Bash specific?

jthegedus (approved):
LGTM!

lib/utils.bash
Stratus3D: Yes I believe so. But this is a Bash file and should only be executed by Bash.

PR #1088, opened 2021-11-08
fix: clarify the wording when no version is set

offbyone:
# Summary

"no version set" can read as a noun, this confused someone I work with
recently :)


jthegedus (approved):
Thanks

jthegedus:
`shfmt` made some changes for some reason, I was using the `shfmt` version in the repo's `.tool-versions`.

PR #1094, opened 2021-11-15
Add missing sudo to apt install command

kerolloz:


Stratus3D:
We've intentionally left off as much as possible because we leave it up to the end user to know and use `sudo` when necessary. `sudo` is typically required on most systems for apt-get but if the user doesn't use when they should, they will get a error that will prompt them to use it, so there is no downside to missing it other than having to re-run the command as root.

PR #1095, opened 2021-11-15
docs: Add missing `sudo` to apt install command

kerolloz:


Stratus3D:
We've intentionally left off as much as possible because we leave it up to the end user to know and use `sudo` when necessary. `sudo` is typically required on most systems for apt-get but if the user doesn't use when they should, they will get a error that will prompt them to use it, so there is no downside to missing it other than having to re-run the command as root.

kerolloz:
@Stratus3D Oh yeah, I understand your point of view. I just think it would be more consistent with the previous commands 
image
they all have `sudo` which makes more sense if this command have sudo as well so the user won't have to get an error. just copy and paste the command like the previous ones. Let me know what you think...

Stratus3D:
Good point. I guess I'm inclined to get rid of those uses of sudo though. What are your thoughts? I like keeping the docs as concise as possible, and most folks who use a package manager know whether or not they need to run the install operation as root.

Stratus3D:
FYI @jthegedus, do you have any opinion on this?

kerolloz:
@Stratus3D that's great. I agree with you. It doesn't really matter getting rid of all `sudo`s or using `sudo` everywhere, what matters is keeping the docs concise.

Stratus3D:
I'll let @jthegedus decide this one, but I'm in favor of removing all uses of `sudo`.

jthegedus:
It is hard to balance between ergonomics and ease-of-use and forcing people to understand and be responsible for their own systems security. Given we er on the side of caution removing `sudo` seems appropriate.

@kerolloz Can you update this PR to remove all `sudo`s?

Perhaps a generic note in the "getting started" section noting that `sudo` may be required depending on the user's system configuration

jthegedus (changes requested):
Given we er on the side of caution removing `sudo `seems appropriate

Stratus3D:
Thanks for the PR @kerolloz , but I think we need to make the docs consistent by removing `sudo` rather than adding it. I am going to close this PR, but feel free to re-open or create another PR that removes uses of `sudo` in the docs.

kerolloz:
I'm sorry for being late on this one. Done in #1116
@Stratus3D  @jthegedus

Issue #1036, opened 2021-09-02
asdf list could return the latest version available vs what's installed locally


scalp42:
### Is your feature request related to a problem? Please describe

While setting versions in `.tool-versions` is awesome, it'd be great to have a way to know when a new version is available.

### Describe the proposed solution

When typing `asdf list`, it'd print the latest version available:

```
$> asdf list

ruby (latest is 3.0.2)
  2.7.3
  3.0.1
terraform (latest is 1.0.5)
  0.14.4
  0.15.5
  1.0.4
```

### Describe similar `asdf` features and why they are not sufficient

`asdf latest X` work but it becomes cumbersome when the list starts growing.

### Describe other workarounds you've considered

Just scripting using `asdf list`, `asdf info` and `asdf latest`

PS: thank you very much and no matter what, `asdf` is awesome!

jthegedus:
Hmm, this would add a call to `asdf latest <plugin>` per plugin in the list, slowing down the currently local-only command. We want to avoid network calls if we can.

Perhaps if we capture this under a flag `asdf list --show-latest` or an `.asdfrc` config value it would preserve existing local-only behaviour.

scalp42:
Yes, totally agree re keeping it local.

jthegedus:
Just thinking out loud, perhaps a command `asdf latest --all` which calls `asdf latest <plugin>` for each plugin you have installed and a marker for if you have that latest version installed?

```
$> asdf list

ruby
  2.7.3
  3.0.1
terraform
  0.14.4
  0.15.5
  1.0.4

$> asdf latest --all

ruby      3.0.1    installed
terraform 1.0.5    missing
```

Would this grant you the information you are after?

scalp42:
yes, it'd be perfect!

PR #1096, opened 2021-11-16
feat(latest): adds the flag --all to the latest command

threkk:
# Summary
As discussed in #1036, this patch adds a `--all` flag to the latest command which will show the list of plugins, their latest version and if this one is installed or not.

Fixes: #1036

## Other Information
This patch basically reuses the logic used in the list command and the latest command. The main different is the removal of the query parameter as different plugins might need different queries.

### DIFF
--- a/lib/commands/command-latest.bash
+++ b/lib/commands/command-latest.bash
@@ -7,6 +7,10 @@ latest_command() {
   local query=$2
   local plugin_path
 
+  if [ "$plugin_name" == "--all" ]; then
+    latest_all
+  fi
+
   [[ -z $query ]] && query="$DEFAULT_QUERY"
 
   plugin_path=$(get_plugin_path "$plugin_name")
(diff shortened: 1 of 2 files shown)


jthegedus (approved):
Looks good to me, thanks @threkk

jthegedus:
@threkk this has a bug. When a plugin is installed, but no versions installed, this prints "installed". EG:

```
~/projects 
❯ asdf latest --all
deno    1.16.4  installed
firebase        9.23.0  installed
gcloud  365.0.1 installed
nodejs  17.2.0  installed
rust    1.57.0  installed

~/projects took 5s 
❯ asdf list gcloud
  No versions installed
 ```
 
 Are you able to update a fix with additional test cases for the permutations?

threkk:
Hi! I am sorry for this, I will look into it ASAP. You can expect a patch this week.

jthegedus:
> Hi! I am sorry for this, I will look into it ASAP. You can expect a patch this week.

No worries mate 😄  We should also document this new flag in the `help.txt`.

threkk:
Hi! The issue has been fixed at https://github.com/asdf-vm/asdf/pull/1118 :)

Issue #1098, opened 2021-11-26
Cleanup invalid shim metadata with `asdf reshim`


Kaylebor:
### Is your feature request related to a problem? Please describe

In recent versions, some unknown plugin interaction caused certain shim metadata to be corrupted; namely, a ghost `..` plugin appeared in these shims, which caused the following warning to appear every time they are run:

```
$ASDF_DIR/lib/utils.bash: line 234: ASDF_.._VERSION: invalid variable name
```

See #1029

The current workaround is to delete the offending shims, and calling `asdf reshim` afterwards. Since the number of shims affected is likely to be unknown, it is often better to call `rm $ASDF_DIR/shims/*` instead.

### Describe the proposed solution

Implement a function `cleanup_invalid_shim_metadata`, to be called inside `reshim_command`, to remove invalid metadata from shims (only invalid plugins for now).

Instead of checking for valid characters, I propose that any non-installed plugin be deleted from shim metadata instead, so it catches potential future issues.

### Describe similar `asdf` features and why they are not sufficient

`asdf reshim` does check for obsolete shims, but only on currently existing plugins.

While this usually is not a problem, if this metadata has invalid values (like the above `..`), there can be some unexpected behaviours. Fully cleaning any non-installed plugins means that any non-valid value will be removed, as problematic plugin names would be caught and fixed in other parts of `asdf` or the plugins themselves.

### Describe other workarounds you've considered

The ideal solution would be for these "ghost" shims not to appear in the first place, but I lack the knowledge of `bash` and `asdf` needed to find where it originated from. Plus, it might happen again.

An alternate option is not to have this built-in into `reshim`, but as a separate command. The main reason to do this could be speed concerns, but [my local prototype of this feature](https://github.com/Kaylebor/asdf/tree/cleanup-invalid-shims) doesn't increase the execution time noticeably.

Stratus3D:
Hi @Kaylebor , I'm going to close this issue due to inactivity, but I do want to say I think we've improved the behavior of asdf in this regard. I've not observed invalid shim metadata myself in recent years. Additionally, I'm preparing to replace all the Bash code for asdf with Go (PR https://github.com/asdf-vm/asdf/pull/1805). Go is much faster than Bash, so it deletes and re-generates all shims every time something changes. You are welcome to give it a try if you want to test it out.

PR #1099, opened 2021-11-26
feat: Cleanup invalid shims

Kaylebor:
# Summary

Adds a new function `cleanup_invalid_shim_metadata`.

This function looks for references to non-installed plugins for the given plugin, and removes this metadata.

Also, adds a check in `check_if_plugin_exists` that filters out '..' as a plugin name.

Related: #1029
Closes: #1098

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -111,7 +111,9 @@ check_if_plugin_exists() {
     exit 1
   fi
 
-  if [ ! -d "$(asdf_data_dir)/plugins/$plugin_name" ]; then
+  # ".." is evaluated to the parent plugins folder, which should not be valid
+  if [ ".." == "${plugin_name}" ] \
+    || [ ! -d "$(asdf_data_dir)/plugins/$plugin_name" ]; then
     display_error "No such plugin: $plugin_name"
     exit 1
   fi
--- a/lib/commands/command-reshim.bash
+++ b/lib/commands/command-reshim.bash
@@ -42,7 +42,7 @@ reshim_command() {
       asdf_run_hook "post_asdf_reshim_$plugin_name" "$full_version_name"
     done
   fi
-
+  cleanup_invalid_shim_metadata "$plugin_name"
 }
 
 ensure_shims_dir() {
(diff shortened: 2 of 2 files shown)


Stratus3D:
Thanks for the PR @Kaylebor, however, I do not believe anyone has figured out what is causing this bug. Somewhere in the asdf codebase there is an expression that is introducing `..` as a plugin name. We need to find that code and correct it. Your changes are nice, and I appreciate you taking the time to create this PR, but I'd like fix the root cause rather than just removing `..` whenever it appears. Thoughts?

lib/utils.bash
```diff
@@ -111,7 +111,9 @@ check_if_plugin_exists() {
     exit 1
   fi
 
-  if [ ! -d "$(asdf_data_dir)/plugins/$plugin_name" ]; then
+  # ".." is evaluated to the parent plugins folder, which should not be valid
+  if [ ".." == "${plugin_name}" ] \
```
Stratus3D, line 115: Were you actually seeing this function invoked with a plugin name of `..`? If so this a good clue.

Kaylebor: I'm still not sure how exactly I managed to get `..` into the metadata, so this is more of a conjecture after looking over the code for a while.

At the very least, without this check, I'm pretty sure that if `..` comes as an argument to `check_if_plugin_exist`, it would be recognized as valid, since `..` always exists in the plugin folder (even when empty).

I left this in a separate commit in case the function I made was rejected, since I thought it's worthwhile to investigate this further.

Kaylebor:
> Thanks for the PR @Kaylebor, however, I do not believe anyone has figured out what is causing this bug. Somewhere in the asdf codebase there is an expression that is introducing `..` as a plugin name. We need to find that code and correct it. Your changes are nice, and I appreciate you taking the time to create this PR, but I'd like fix the root cause rather than just removing `..` whenever it appears. Thoughts?

@Stratus3D no problem, I'm a newcomer here so I don't really have the full picture of the code in mind when making this suggestion. And I also agree that fixing the root cause would be the preferred approach, as this shouldn't happen in the first place.

Still, I'd argue that having some way to clean shim metadata would be useful for fixing/troubleshooting issues like this in the future; that's why this function doesn't explicitly look for `..`, but instead checks if the plugin exists in the plugins folder.
In order to keep `reshim` simple, maybe locking such functionality behind a new flag or command would be better.

Do you see value in any of these approaches?

lib/utils.bash
Stratus3D: That's a good observation. I am going to do some testing locally to try to figure out what code is invoking this function with `..`, I tested it with the `reshim` command and it doesn't appear to ever pass in `..` for me.

Stratus3D:
Hi @Kaylebor , I'm sorry about my lack of reply here. Thanks for your effort on this, I was away from asdf for a couple years and came back this year to rewrite it in Golang. I've got the Go code merged this week (PR #1805) (dev binaries here - https://github.com/asdf-vm/asdf/releases/tag/54d15e3) and am preparing the first release of the Go code. The Go code is strongly typed and I've not observed it ever generating invalid shims, so I believe it will fix this issue. I'm going to close this PR as it no longer makes sense to merge Bash changes.

PR #1100, opened 2021-12-03
fix: sort listing by version number in natural order

kianmeng:
# Summary

This PR fixes the issue where the installed versions from `asdf list` or `asdf list <plugin` command are not sorted in natural order.

## Other Information

Before:

```bash
$ asdf list elixir
   ...
  1.8.0-otp-20
  1.8.1
  1.8.1-otp-21
  1.8.1-otp-22
  1.8.2
  1.8.2-otp-21
  1.8.2-otp-22
  1.9
  1.9.1
  1.9.1-otp-22
  1.9.2-otp-22
  1.9.4
  1.9.4-otp-22
  master
  master-otp-24
```

After:

```bash
$ asdf list elixir
  ...
  1.12
  1.12.0
  1.12.1
  1.12.1-otp-24
  1.12.3-otp-23
  1.12.3-otp-24
  1.13.0-rc.0-otp-24
  1.13.0-rc.1-otp-24
  master
  master-otp-24
```

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -95,8 +95,7 @@ list_installed_versions() {
   plugin_installs_path="$(asdf_data_dir)/installs/${plugin_name}"
 
   if [ -d "$plugin_installs_path" ]; then
-    for install in "${plugin_installs_path}"/*/; do
-      [[ -e "$install" ]] || break
+    for install in $(ls -1v ${plugin_installs_path}); do
       basename "$install" | sed 's/^ref-/ref:/'
     done
   fi


jthegedus:
Hi @kianmeng are you able to share resources on how portable `ls -1v` is? We would need to ensure this works on all major shells and OSs we support.

kianmeng:
> Hi @kianmeng are you able to share resources on how portable `ls -v` is? We were previously using `-1` elsewhere so not concerned with that. We would need to ensure `-v` works on all major shells and OSs we support.

I will assume the supported OSes are GNU/Linux and macOS and supported Shells are Bash, Zsh, and Fish? Is there any minimum version requirement for these Shells?

> Tests would also need to be updated, ideally with versions that exhibit this behaviour should a regression occur.

Noted.

jthegedus:
>I will assume the supported OSes are GNU/Linux and macOS and supported Shells are Bash, Zsh, and Fish? Is there any minimum version requirement for these Shells?

This is mostly it. OSs: GNU/Linux, macOS. Shells: Bash, ZSH, Fish, Elvish.

>Is there any minimum version requirement for these Shells?

We don't have official version targets (something we wish to capture), but as low as possible.

lib/utils.bash
```diff
@@ -95,8 +95,7 @@ list_installed_versions() {
   plugin_installs_path="$(asdf_data_dir)/installs/${plugin_name}"
 
   if [ -d "$plugin_installs_path" ]; then
-    for install in "${plugin_installs_path}"/*/; do
-      [[ -e "$install" ]] || break
+    for install in $(ls -1v ${plugin_installs_path}); do
```
Stratus3D, line 98: Is there a way to achieve this without `ls`? I've been considering banning `ls` from the codebase because it's caused a couple bugs in the past.

kianmeng: None I can think of or found. Perhaps we can't or shouldn't implement this feature since using `ls` is not universal applicable for all platforms.

Your thoughts? @jthegedus @Stratus3D

Stratus3D: Would it not be possible to do this with `sort`?

Stratus3D: I guess part of the problem here is that when we list all versions of a plugin, we call out to the plugin's `list-all` callback, which is responsible for returning versions in the correct order. Here we are dealing with a subset of those same versions, but we don't have a way of calling out to the plugin to sort them.

Stratus3D: I suppose we could add another plugin callback that would receive a list of versions and would be expected to sort them, but I'm not sure it would be worth the hassle.

Stratus3D:
Thanks for your PR @kianmeng . I'm sorry I did not give you feedback sooner. We decided to move forward with banning `ls` in the codebase as its output is intended for human use and not machine. There are also some portability issues as not every platform supports every flag. Instead we recommend using `find` to achieve the same thing.

If you can achieve this with `find` we may consider an alternative approach. Closing this PR for now.

kianmeng:
@Stratus3D Noted, and thanks for the explanation.

PR #1106, opened 2021-12-05
fix: shorthand grep options for alpine support

jthure:
# Summary

The version of grep that is shipped with alpine doesn't seem to support "long name" options


jthegedus (approved):
looks good to me

PR #1117, opened 2021-12-08
fix: ban grep long flags

### LINKED ISSUE #1109: bug: ban longhand grep options as they are not supported on Alpine
### Describe the Bug

#1106 fixed an current issue, however we should guard against this arising in future.

### Steps to Reproduce

NA

### Expected Behaviour

Block usage of longhand `grep` options in CI

### Actual Behaviour

NA

### Environment

```shell
NA
```

NA

Stratus3D:
Fixes #1109

Rather than just having one array of both literal commands and regex patterns, I split the banned commands into two arrays - one of literal commands and one of regex. This allows us to use the `-E` (extended regex) option when looking for banned commands via regex.

### DIFF
--- a/test/banned_commands.bats
+++ b/test/banned_commands.bats
@@ -14,20 +14,26 @@ banned_commands=(
     # It's best to avoid eval as it makes it easier to accidentally execute
     # arbitrary strings
     eval
-    # grep -y does not work on alpine and should be "grep -i" either way
-    "grep.* -y"
-    # grep -P is not a valid option in OSX.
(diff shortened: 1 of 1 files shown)


jthegedus (approved):
Looks good to me

jthegedus:
It was just a semantic-pr lint issue. PRs with a single commit have their commit used as the merge name, not the PR title. So the commit must match the semantic-pr naming convention or have more than one commit to use the PR title instead.

PR #1118, opened 2021-12-08
fix: latest --all correctly report plugins as missing

threkk:
# Summary

Fixes the bug reported at #1096. Uninstalled plug-ins are now marked as missing. The plug-in will be marked as installed only if the latest version is among the installed plug-ins.

Fixes: #1096


jthegedus (approved):
LGTM. Tested locally, works as expected. Thanks

threkk:
Thank you very much for catching the bug! I didn't test carefully enough.

Issue #1120, opened 2021-12-13
bug: funcion get_asdf_config_value_from_file() strips everything after second '='


systemmonkey42:
### Describe the Bug

For a particular plugin (vim in this case), it supports cumtomising the build through environment variables.

Adding the following to ~/.asdfrc
```
pre_install_asdf_install_vim = ASDF_VIM_CONFIG=--with-x
```
does not work, because the function `get_asdf_config_value_from_file()` uses `awk -F=`, and always returns only the command
`ASDF_VIM_CONFIG` which fails.

### Steps to Reproduce

asdf plugin add vim

Edit ~/.asdfrc add
```
pre_install_asdf_install_vim = ASDF_VIM_CONFIG=--with-x
```

```
$ asdf install vim latest
/home/me/.asdf/lib/utils.bash: line 662: ASDF_VIM_CONFIG: command not found
```

### Expected Behaviour

Sets the `ASDF_VIM_CONFIG` variable, and build.

### Actual Behaviour

generates a shell error;
```
/home/me/.asdf/lib/utils.bash: line 662: ASDF_VIM_CONFIG: command not found
```

### Environment

```shell
# asdf info
OS:
Linux vm 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

ASDF VERSION:
v0.9.0-9ee24a3

ASDF ENVIRONMENT VARIABLES:
ASDF_DATA_DIR=/home/me/.cache/asdf
ASDF_DIR=/home/me/.asdf

ASDF INSTALLED PLUGINS:
dust                         https://github.com/looztra/asdf-dust.git master f918e74
vim                          https://github.com/tsuyoshicho/asdf-vim.git master 8001a3a

```
```


### asdf plugins affected (if relevant)


vim plugin in this case, but essentially all hooks.

Note,. the following fixes the issue
```
lib/utils.bash|360| get_asdf_config_value_from_file() {
[...]

PR #1121, opened 2021-12-13
fix: Fix for issue 1120

systemmonkey42:
# Summary

Fixes `get_asdf_config_value_from_file()` to handle '=' in values
Fixes `run_hook()` to exit cleanly on hook failure.

#1120


jthegedus:
Can you add tests to this. 1 capturing the successful execution with the new code and 1 with an expected error with the usage that caused you to come and fix this?

systemmonkey42:
> Can you add tests to this. 1 capturing the successful execution with the new code and 1 with an expected error with the usage that caused you to come and fix this?

More than happy to add the tests.

> It seems the usage outlined in #1120 is incorrect as the `asdfrc` file is only supposed to have values handled by the asdf-core, not the plugins.

Only asdf-core reads and executes the `pre_asdf_install_<plugin>` script.

The fact that it sets up the environment prior to the plugin installation process is arguably the intention of having a `pre_asdf_install_<plugin>` feature in the first place.  No plugin I'm aware of will read ~/.asdfrc directly.

systemmonkey42:
I just found most tests 
```
325 tests, 216 failures
```
will fail if `ASDF_DATA_DIR` is set.
I might see if I can test in a docker container.

systemmonkey42:
Test added.

lib/utils.bash
```diff
@@ -658,7 +658,10 @@ asdf_run_hook() {
       unset asdf_hook_fun
       ev'al' "$hook_cmd" # ignore banned command just here
     }
-    asdf_hook_fun "${@:2}"
+    asdf_hook_fun "${@:2}" || {
+      display_error "Hook \"${hook_name}\" failed."
```
Stratus3D, line 662: Is this consistent with other places in the codebase where hooks are invoked? I don't think we should print errors here but not in other places.

Stratus3D (approved):
Other than the one question I had, the code looks good to me. Build is running now.

lib/utils.bash
systemmonkey42: A quick survey of the code reveals that *all* hooks appear to end up in the `asdf_run_hook` function, and as such the error message and immediate exit _is_ consistent between all types of hooks.

However, I also recognize that a hook failure (like a post-reshim-plugin hook) which returns non-zero exit code might want to be ignored.

I am starting to feel that *not* modifying the asdf_run_hook, but rather checking the exit code on 
every call would be the better way to do it.

Eg, 
`lib/commands/command-exec.bash` already validates the exit code, and passes it on if the hook fails
```
      asdf_run_hook "pre_${plugin_name}_${shim_name}" "${shim_args[@]}"
      pre_status=$?
      if [ "$pre_status" -ne 0 ]; then
        return "$pre_status"
      fi
      exec "$executable_path" "${shim_args[@]}"
```
while the exit codes are ignored for 
`lib/commands/command-install.bash`
```
        asdf_run_hook "pre_asdf_download_${plugin_name}" "$full_version"
        "${plugin_path}"/bin/download
...
        asdf_run_hook "pre_asdf_install_${plugin_name}" "$full_version"
        "${plugin_path}"/bin/install
```

I suggest reverting the change to asdf_run_hook, and looking at checking the return code handling as a separate task

systemmonkey42:
> Other than the one question I had, the code looks good to me. Build is running now.

Let me know how you feel about reverting the `asdf_run_hook` change.  The more I think about it, the more I suspect it will break existing behaviour

lib/utils.bash
jthegedus: >I suggest reverting the change to asdf_run_hook, and looking at checking the return code handling as a separate task
I concur

Stratus3D: If you are saying change this:

```
 asdf_hook_fun "${@:2}" || {
      display_error "Hook \"${hook_name}\" failed."
```

back to this:

```
 asdf_hook_fun "${@:2}"
```

Then yes, I agree 👍

systemmonkey42: Done

Stratus3D (approved):
@jthegedus any comments here? I'd like to get this merged.

Stratus3D:
@systemmonkey42 I'm going to merge this PR shortly since it looks like @jthegedus isn't available right now. Can you update the PR to target the master branch instead of the `release-v0.10.0` branch? Thanks!

Stratus3D:
I rebased and manually merged this branch into master. For some reason Github didn't detect the merge but I didn't modify your comments so you'll see them in the master branch now. Thanks for the PR!

PR #1132, opened 2021-12-16
docs: Document the latest-stable callback plugin creation docs

### LINKED ISSUE #1110: docs: missing documentation for plugin bin/latest-stable
### Describe the Bug

There is currently no documentation for the `${plugin}/bin/latest-stable` script which is used in the `asdf latest` command.

This script exists to defer to the plugin which version is the latest as performing `tail` on the entire `list-all` results can give undesirable results when tools like `python` and `java` have many different sources sorted alphabetically.

We need to document this for plugin authors.

Related https://github.com/asdf-vm/asdf/issues/1028

### Steps to Reproduce

NA

### Expected Behaviour

NA

### Actual Behaviour

NA

### Environment

```shell
NA
```

NA

Stratus3D:
Fixes #1110

### DIFF
--- a/docs/plugins/create.md
+++ b/docs/plugins/create.md
@@ -74,6 +74,12 @@ Each of these scripts should tailor their output to the current operating system
 
 The help callback script MUST NOT output any information that is already covered in the core asdf-vm documentation. General asdf usage information must not be present.
 
+#### bin/latest-stable
+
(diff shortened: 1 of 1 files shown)


Stratus3D:
Merging this without review because we can always improve docs later.

PR #1139, opened 2021-12-26
feat: case-insensitive filtering of unstable versions in `latest`

joschi:
# Summary

The `asdf latest` command currently only filters unstable versions with lowercase version suffixes (`alpha`, `pre`, `snapshot`, etc.) but fails if there are version suffixes using uppercase.

This is currently happening with the [asdf-kcctl](https://github.com/joschi/asdf-kcctl) plugin.


Stratus3D (approved):
Thanks for the fix! I don't see any harm in merging this.

Stratus3D:
Thanks for the tests!

PR #1140, opened 2021-12-27
docs: update git/bash install docs to use 0.9.1

myersg86:
Branch 0.8.1 no longer exists, leading to detached head upon checkout

# Summary

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -37,7 +37,7 @@ We recommend using Git, though there are other platform specific methods:
 
 | Method   | Command                                                                                                                                                             |
 | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| Git      | `git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.8.1`                                                                                             |
+| Git      | `git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.9.1`                                                                                             |
 | Homebrew | `brew install asdf`                                                                                                                                                 |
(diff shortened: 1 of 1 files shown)


jthegedus:
Thanks for raising this. I noticed this issue but haven't had the time to debug why https://github.com/asdf-vm/asdf/blob/master/.github/workflows/docs-version.yml didn't work on the `0.9.x` releases and update itself. It would be better to investigate why that didn't trigger than manually update the version.

Taytay:
To be honest, I found it a bit strange that the recommended way to install asdf was to install a particular branch in the first place. I dutifully followed it, but then I ran `asdf update` and discovered that there was actually a more recent version (0.9.0. This made me wonder if our machine setup scripts would be out of date if they referred to a particular branch of asdf as the docs recommended. Shouldn't the docs just tell people to install the latest from the default branch, since presumably that is the most recent stable version?

Stratus3D:
> To be honest, I found it a bit strange that the recommended way to install asdf was to install a particular branch in the first place. I dutifully followed it, but then I ran `asdf update` and discovered that there was actually a more recent version (0.9.0. This made me wonder if our machine setup scripts would be out of date if they referred to a particular branch of asdf as the docs recommended. Shouldn't the docs just tell people to install the latest from the default branch, since presumably that is the most recent stable version?

No, the latest default branch is not considered stable. The standards for a tagged release are a bit higher, and we expect most users to be using a tagged release and not latest default (though us maintainers often do use latest default most of the time). The docs should instruct users to install the latest tag, and as seen here we've got a bug somewhere that isn't bumping the version in the docs. 

We instruct users installing via git to install a specific **tag**, not a specific **branch**, the git flag for doing so happens to be `--branch`, which is a bit confusing.

Stratus3D (approved):

Stratus3D:
I went ahead and merged this to prevent users from continuing to install the old version.

ksmolnicki:
But the v0.9.1 branch/tag is still missing

PR #1141, opened 2021-12-28
fix: Ban `ls` command

Stratus3D:
# Summary

`ls` should be banned because it's seldom the right tool for the job. Often plain old shell globbing, or `find` are more reliable ways of listing directories. 

* Only grep sh/bash files for banned commands
* Ban `ls` command
* Replace instances of `ls` with `find`

Related: https://github.com/asdf-vm/asdf/pull/1100 https://github.com/asdf-vm/asdf/issues/1029

Fixes: Uncertain whether this will fix any outstanding issues.

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -454,7 +454,7 @@ resolve_symlink() {
   # This seems to be the only cross-platform way to resolve symlink paths to
   # the real file path.
   # shellcheck disable=SC2012
-  resolved_path=$(ls -l "$symlink" | sed -e 's|.*-> \(.*\)|\1|')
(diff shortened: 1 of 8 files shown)


lib/utils.bash
```diff
@@ -454,7 +454,7 @@ resolve_symlink() {
   # This seems to be the only cross-platform way to resolve symlink paths to
   # the real file path.
   # shellcheck disable=SC2012
-  resolved_path=$(ls -l "$symlink" | sed -e 's|.*-> \(.*\)|\1|')
+  resolved_path=$(ls -l "$symlink" | sed -e 's|.*-> \(.*\)|\1|') # asdf_allow: ls '
```
Stratus3D, line 457: I've not yet found a better cross-platform solution to this.

PR #1158, opened 2022-01-13
fix: Prevent unbound variable error with nounset in asdf.sh

joehorsnell:
# Summary

`asdf` [v0.9.0][1] included a [bug fix][2] to not override an existing `ASDF_DIR`.

However, if `ASDF_DIR` is not set at all, then this causes an error when using bash `set -u`, or `set -o nounset` - see [here][3] for additional info.

`asdf.sh: line 6: ASDF_DIR: unbound variable`

Edit: see also #160 and #817

[1]: https://github.com/asdf-vm/asdf/releases/tag/v0.9.0
[2]: https://github.com/asdf-vm/asdf/pull/1008
[3]: https://mywiki.wooledge.org/BashFAQ/112

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -3,7 +3,7 @@
 # For Bash, ${BASH_SOURCE[0]} will be used to obtain this script's path.
 # For Zsh and others, $0 (the path to the shell or script) will be used.
 _under="$_"
-if [ -z "$ASDF_DIR" ]; then
+if [ -z "${ASDF_DIR:-}" ]; then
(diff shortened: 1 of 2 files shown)


Stratus3D:
Sorry for the late review here @joehorsnell ! As a general rule we require issues to be reported prior to contributing a fix. However I see that this a fairly trivial one-line fix, so assuming tests pass I will go ahead and merge it. Thanks for the contributing the fix and the tests to prevent a regression!

Stratus3D (approved):

PR #1159, opened 2022-01-13
feat: upgrade elvish to 0.17.0

brunoroque06:
# Summary

With version 0.17 of elvish some new syntax was introduced (https://elv.sh/blog/0.17.0-release-notes.html). The one we currently use is deprecated, and will be removed in version 0.18. I just ran the official upgrader script (https://github.com/elves/upgrade-scripts-for-0.17) on the Elvish code.


asdf.elv
```diff
@@ -87,20 +87,20 @@ for path [
...
       $path
     ]
   }
 }
 
 # Setup argument completions
-fn arg-completer [@argz]{
-  argz = $argz[1:-1]  # strip 'asdf' and trailing empty string
+fn arg-completer {|@argz|
+  set argz = $argz[1:-1]  # strip 'asdf' and trailing empty string
```
cherryblossom000, line 99:
```suggestion
  set argz = $argz[1..-1]  # strip 'asdf' and trailing empty string
```
One thing the Elvish v0.17.0 upgrade script doesn't pick up on is the slice syntax. The legacy `a:b` syntax doesn't work anymore.

The same legacy slice syntax appears in lines 193 and 199.

asdf.elv
```diff
@@ -31,31 +31,31 @@ fn asdf [command @args]{
...
+        set matched = $false
         break
       }
     }
   }
   put $matched
 }
 
-fn ls-shims []{
+fn ls-shims {||
```
cherryblossom000, line 51:
```suggestion
fn ls-shims {
```

I'm aware the Elvish v0.17.0 upgrade script converts `[]{ body }` into `{|| body }` and I guess this is a matter of stylistic preference, but the `||` here isn't needed (the `[]` wasn't required in the legacy syntax as well). The [Elvish docs](https://elv.sh/ref/language.html#function) don't use `{|| body }` and instead use `{ body }`, and IMO `{ body }` looks cleaner.

asdf.elv
```diff
@@ -31,31 +31,31 @@ fn asdf [command @args]{
...
   ls $asdf_data_dir'/shims'
 }
 
-fn ls-executables []{
+fn ls-executables {||
```
cherryblossom000, line 55:
```suggestion
fn ls-executables {
```

cherryblossom000 (changes requested):

docs/guide/getting-started.md
```diff
@@ -152,8 +152,8 @@ Add `asdf.elv` to your `~/.elvish/rc.elv` with:
 
 ```shell:no-line-numbers
 mkdir -p ~/.elvish/lib; ln -s ~/.asdf/asdf.elv ~/.elvish/lib/asdf.elv
-echo "\n"'use asdf _asdf; fn asdf [@args]{_asdf:asdf $@args}' >> ~/.elvish/rc.elv
-echo "\n"'edit:completion:arg-completer[asdf] = $_asdf:arg-completer~' >> ~/.elvish/rc.elv
+echo "\n"'use asdf _asdf; fn asdf {|@args| _asdf:asdf $@args}' >> ~/.elvish/rc.elv
```
cherryblossom000, line 155: In v0.17.0, the [Elvish RC file](https://elv.sh/ref/command.html#rc-file) location was changed.

> - If the legacy `~/.elvish/rc.elv` exists, it is used (this will be ignored in a future version).
> - Otherwise:
>   - On UNIX (including macOS), `$XDG_CONFIG_HOME/elvish/rc.elv` is used, defaulting to `~/.config/elvish/rc.elv` if `$XDG_CONFIG_HOME` is unset or empty.
>   - On Windows, `%AppData%\elvish\rc.elv` is used.

The docs should be updated to reflect these changes, but I'm not sure what to replace `~/.elvish/rc.elv` with as the RC file varies on different systems and configurations. Maybe it could be replaced with `~/.config/elvish/rc.elv` and the RC file docs could be linked above the script?

cherryblossom000:
This isn't related to the Elvish v0.17.0 upgrade but `fn asdf [@args]{_asdf:asdf $@args}` in the installation docs could be replaced with `var asdf~ = $_asdf:asdf~`. A bit out of scope for this PR though.

Stratus3D (commented):
Sorry for the late review @brunoroque06, thanks for the PR!

I think @cherryblossom000 has some great comments that should be addressed. Otherwise looks good to me!

asdf.elv
brunoroque06: Done! Thanks

asdf.elv
brunoroque06: Yes, looks better without the `||`. Done

docs/guide/getting-started.md
brunoroque06: I changed everything to `~/.config/elvish`. What to you think?

brunoroque06:
> This isn't related to the Elvish v0.17.0 upgrade but `fn asdf [@args]{_asdf:asdf $@args}` in the installation docs could be replaced with `var asdf~ = $_asdf:asdf~`. A bit out of scope for this PR though.

I changed this as well. I don't think we need to delay this to another commit. What do you think?

cherryblossom000:
There are also some references to the old config location here:

https://github.com/asdf-vm/asdf/blob/2a538fe63144c3d0710df23738c1f06363b10fa6/test/asdf_elvish.bats#L5-L13

Also, as no one seems to have mentioned it, this PR is technically a breaking change as it will drop support for Elvish versions prior to v0.17.0.

docs/guide/getting-started.md
cherryblossom000: There are also some references to the old config location here:

https://github.com/asdf-vm/asdf/blob/2a538fe63144c3d0710df23738c1f06363b10fa6/test/asdf_elvish.bats#L5-L13

cherryblossom000:
> I changed this as well. I don't think we need to delay this to another commit. What do you think?

LGTM!

Also, as no one seems to have mentioned it yet, this PR is technically a breaking change as it will drop support for Elvish versions prior to v0.17.0.

docs/guide/getting-started.md
brunoroque06: Oh thanks, missed that. It's done.

test/asdf_elvish.bats
```diff
@@ -20,7 +20,7 @@ cleaned_path() {
   output=$(elvish -norc -c "
     unset-env ASDF_DIR
     paths = [$(cleaned_path)]
```
cherryblossom000, line 22:
```suggestion
    set paths = [$(cleaned_path)]
```

test/asdf_elvish.bats
```diff
@@ -31,7 +31,7 @@ cleaned_path() {
   output=$(elvish -norc -c "
     set-env ASDF_DIR "/path/to/asdf"
     paths = [$(cleaned_path)]
```
cherryblossom000, line 33:
```suggestion
    set paths = [$(cleaned_path)]
```

test/asdf_elvish.bats
```diff
@@ -42,7 +42,7 @@ cleaned_path() {
   output=$(elvish -norc -c "
     set-env ASDF_DATA_DIR "/path/to/asdf-data"
     paths = [$(cleaned_path)]
```
cherryblossom000, line 44:
```suggestion
    set paths = [$(cleaned_path)]
```

test/asdf_elvish.bats
```diff
@@ -53,7 +53,7 @@ cleaned_path() {
   result=$(elvish -norc -c "
     unset-env ASDF_DIR
     paths = [$(cleaned_path)]
```
cherryblossom000, line 55:
```suggestion
    set paths = [$(cleaned_path)]
```

test/asdf_elvish.bats
```diff
@@ -66,7 +66,7 @@ cleaned_path() {
   output=$(elvish -norc -c "
     unset-env ASDF_DIR
     paths = [$(cleaned_path)]
```
cherryblossom000, line 68:
```suggestion
    set paths = [$(cleaned_path)]
```

test/asdf_elvish.bats
```diff
@@ -77,7 +77,7 @@ cleaned_path() {
...
```
cherryblossom000, line 79:
```suggestion
    set paths = [$(cleaned_path)]
```

test/asdf_elvish.bats
```diff
@@ -89,7 +89,7 @@ cleaned_path() {
   result=$(elvish -norc -c "
     set-env ASDF_DIR $(pwd)
     paths = [$(cleaned_path)]
```
cherryblossom000, line 91:
```suggestion
    set paths = [$(cleaned_path)]
```

cherryblossom000 (changes requested):
Sorry, I just noticed the legacy assignment syntax in the tests

cherryblossom000 (approved):
LGTM!

Stratus3D (approved):

PR #1166, opened 2022-01-25
docs(website): add simplifed chinese

zhonger:
# Summary

Docs(website): Add simplified chinese


docs/contribute/core.md
```diff
@@ -94,10 +94,10 @@ feat!: feature with breaking change
...
 The full list of `<types>` are: `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `build`, `ci`, `chore`, `revert`.
 
-* `!`: indicates a breaking change
-* `fix`: will create a new SemVer `patch`
-* `feat`: will create a new SemVer `minor`
-* `<type>!`: will create a new SemVer `major`
+- `!`: indicates a breaking change
+- `fix`: will create a new SemVer `patch`
+- `feat`: will create a new SemVer `minor`
+- `<type>!`: will create a new SemVer `major`
```
Stratus3D, line 100: Why were these changed?

Stratus3D (approved):
This is great! I can't read chinese so I can't say if everything here is correct, but scanning through it all I don't see any obvious issues with the changes. Thanks for the PR @zhonger !

Stratus3D:
@asdf-vm/core can anyone else take a look at these changes? I see no reason not to merge this now, so we are currently lacking a chinese translation of the docs.

docs/contribute/core.md
zhonger: I am so sorry for that. It should be not changed. 
In order to distinguish which is changed by me and the `format` tool, I have made two commits.
We can see easily that this change is because of the `format` tool.

Stratus3D:
Since these changes are mostly additive I'm going to go ahead and merge this. I tried google translate on some of the docs and validated that the text in the docs translates to roughly the same thing as what we have for the english docs, but I would love another review from someone who can read the language.

PR #1168, opened 2022-01-30
feat(elvish)!: remove deprecated syntax for Elvish v0.17.0

cherryblossom000:
# Summary

Some fixes and improvements for the Elvish support.

- Updates Elvish script to prevent deprecation warnings on Elvish v0.17.0. This is breaking as it won't work for Elvish versions prior to v0.17.0 (released in December 2021).
- Support calling `asdf` with no arguments. Previously, this would happen:
  ```
  ~> asdf
  Exception: arity mismatch: arguments must be 1 or more values, but is 0 values
  ```
- Add `set-env ASDF_DIR (brew --prefix asdf)/libexec; set-env ASDF_DATA_DIR ~/.asdf` to the `rc.elv` for Homebrew (the Elvish script sets `ASDF_DIR` to `~/.asdf` by default and `ASDF_DATA_DIR` to `ASDF_DIR`, which is incorrect for Homebrew installations)
- Update docs to reflect changes to the [Elvish RC file location](https://elv.sh/ref/command.html#rc-file) in v0.17.0. The new location depends on `$XDG_CONFIG_HOME` on Unix and `%AppData%` on Windows, so I linked to the docs and used `~/.config/elvish/rc.elv` in the installation script. I haven't updated the Chinese docs properly as I don't know Chinese and instead replaced `~/.elvish/rc.elv` with `~/.config/elvish/rc.elv` without the link to the Elvish docs.

These are some more subjective and minor changes as well (please let me know I should undo the commit):
- Use [tilde expansion](elv.sh/ref/language.html#tilde-expansion) instead of `$E:HOME`
- Set `asdf~` variable instead of creating a `fn asdf` and explicitly passing arguments (see [docs on the `~` variable suffix](elv.sh/ref/language.html#variable-suffix))
- Use some destructuring in the `asdf` function using a [rest variable](elv.sh/ref/language.html#var)

## Other Information

Sorry if this PR does too many things at once; I just noticed a few things I could improve with the Elvish support when updating it to work for the latest Elvish version. I'm very happy to split up this PR into smaller ones.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,11 +38,11 @@ jobs:
           sudo apt-get -y install fish curl
 
           # Download elvish binary and add to path
-          curl https://dl.elv.sh/linux-amd64/elvish-v0.16.3.tar.gz -o elvish-v0.16.3.tar.gz
-          tar xzf elvish-v0.16.3.tar.gz
-          rm elvish-v0.16.3.tar.gz
+          curl https://dl.elv.sh/linux-amd64/elvish-v0.17.0.tar.gz -o elvish-v0.17.0.tar.gz
+          tar xzf elvish-v0.17.0.tar.gz
+          rm elvish-v0.17.0.tar.gz
           mkdir -p "$HOME/bin"
-          mv elvish-v0.16.3 "$HOME/bin/elvish"
+          mv elvish-v0.17.0 "$HOME/bin/elvish"
           echo "$HOME/bin" >>"$GITHUB_PATH"
 
       - name: Install bats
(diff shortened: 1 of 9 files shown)


Stratus3D:
This looks like a duplicate of https://github.com/asdf-vm/asdf/pull/1159? Is it not?

cherryblossom000:
Oops, it is. Sorry, I forgot to check if there was a already an existing PR!

Stratus3D:
@cherryblossom000 if you want to help review the code changes on #1159 that would be great! I'm not a elvish shell user. I am planning on reviewing that PR tomorrow.

PR #1172, opened 2022-02-17
Remove last ls command from codebase

Stratus3D:


jthegedus (approved):
LGTM. Not sure why the symlink tests are failing :shrug:

Stratus3D:
It looks like the `stat` command isn't available on Ubuntu ☹️

MichaelDimmitt:
Looks like mac has a different stat command than linux
https://unix.stackexchange.com/a/175326/188491

Maybe if it fails use a fallback for other operating systems?

Ubuntu: 
http://manpages.ubuntu.com/manpages/bionic/man1/stat.1.html

Mac: 
https://ss64.com/osx/stat.html

Stratus3D:
> Looks like mac has a different stat command than linux https://unix.stackexchange.com/a/175326/188491
> 
> Maybe if it fails use a fallback for other operating systems?
> 
> Ubuntu: http://manpages.ubuntu.com/manpages/bionic/man1/stat.1.html
> 
> Mac: https://ss64.com/osx/stat.html

I could use `ls` as the fallback for OS's that don't have `stat -f "%Y"`, but then I might as well just keep using `ls` for everything for consistency. Unless there is another way using `stat` on Ubuntu to get the link path. The general issue with `stat` is that it varies a lot. Whereas `ls`, even though it's intended for human use only, is more consistent across operating systems.

Stratus3D:
Closing since I don't think this is something we'll be able to make work across different operating systems.

PR #1185, opened 2022-03-17
feat: Support for changing the location of plugin scripts

yacchi:
# Summary

I would like to be able to install the tools I am creating with asdf.

I also want to provide the tool itself and the asdf plugin in the same git repository to simplify source code management.
However, since `bin` and `lib` are also used for other purposes, I would like to put the asdf plugin in a separate directory to avoid confusion.

I have tried to implement this by placing a configuration file `.asdf-plugin` in the root of the repository and writing the path to the asdf plugin in it.

Of course, if the `.asdf-plugin` file does not exist, the behavior will be the same as before.

For example, to place the script in `contrib/asdf`, I would configure it as follows:

.asdf-plugin
```
plugin_directory = contrib/asdf
```

```bash
$ ls contrib/asdf/bin
list-all
install
download
```

### DIFF
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -57,6 +57,7 @@ legacy_version_file = no
 use_release_candidates = no
 always_keep_download = no
 plugin_repository_last_check_duration = 60
+use_asdf_plugin_configuration = no
 ```
 
 ### `legacy_version_file`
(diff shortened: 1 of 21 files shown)


yacchi:
Since we will be introducing a new configuration file, please allow us to correct or explain any parts that are not good or difficult to understand.

yacchi:
@jthegedus 
Thanks for the review!

Am I correct in assuming that the `.asdfrc` setting can enable the ability to read `.asdf-plugin` files?
Right now there are only plugins that do not require this functionality, so it would be better to have a setting as suggested.

I will try to add the following settings:

.asdfrc:
```
load_asdf_plugin_configuration = true
```

yacchi:
@jthegedus
I have changed this feature to be enabled by adding `use_asdf_plugin_configuration = yes` to `.asdfrc`.
(I changed it from `load_asdf_plugin_configuration` because there was already a configuration item starting with `use_`)

I also followed to upstream.

Stratus3D:
Sorry for the extremely late reply here. I guess I'm inclined to not merge this as it's a good amount of added complexity just to be able to place the asdf callbacks in a different directory. While I understand this will make things more convenient for @yacchi, it's not clear this is a widely needed feature. I've never had the need for this with the several plugins I've created. @asdf-vm/core thoughts?

Stratus3D:
asdf has been rewritten in Go (https://github.com/asdf-vm/asdf/releases/tag/v0.16.0) so this PR cannot be applied. I'm going to close it now.

Please @yacchi please open a feature request if you'd still like this supported.

PR #1186, opened 2022-03-26
fix: silence plugin missing lib/commands/ error

carkmorwin:
# Summary

When runnning `asdf`, plugins that do not have lib/commands dirs result in a `find` error that is output.

Based on the plugins I have installed, this is not a requirement for functionality and/or a new convention not yet adopted globally.

```
$ asdf
...
PLUGIN nodejs
  asdf nodejs nodebuild
  asdf nodejs update nodebuild
find: /Users/mcor/.asdf/plugins/postgres/lib/commands: No such file or directory
find: /Users/mcor/.asdf/plugins/python/lib/commands: No such file or directory
find: /Users/mcor/.asdf/plugins/ruby/lib/commands: No such file or directory
find: /Users/mcor/.asdf/plugins/shellcheck/lib/commands: No such file or directory
find: /Users/mcor/.asdf/plugins/shfmt/lib/commands: No such file or directory
find: /Users/mcor/.asdf/plugins/terraform/lib/commands: No such file or directory
```

The fix here is to return in the `find_cmd()` function if the lib/comands dir does not exist. 

Not sure where a test would fit in here looking at the test suite, but can add one if deemed necessary.

This is really just to not get "warnings" and really just a cleanup (assuming nothing else breaks).

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -7,6 +7,10 @@ find_cmd() {
   local cmd_dir="$1"
   shift
 
+  if [ ! -d "$cmd_dir" ]; then
+    return
+  fi
+
   local cmd_name
   local args_offset="$#"
   cmd_name="command-$(tr ' ' '-' <<<"${@:1:${args_offset}}").bash"


yordis:
@blopker could you give us some love over here?

blopker:
I would! I'm not a maintainer though, and cannot review code.

yordis:
@blopker 😭 anyone that could help us that you would know about?

Stratus3D:
Hi @carkmorwin, maintainer here. Sorry for the late review. Somehow this PR slipped through the cracks. Thank you for fixing this, however I believe this may have already been fixed in the latest version - `0.10.0`. Can you try updating to the latest version and see if that resolves the missing `lib/commands` error? Thanks!

carkmorwin:
@Stratus3D No problem, I did see that it was fixed in `0.10.0` so closing this PR. Cheers!

PR #1190, opened 2022-03-30
fix: instead /tmp, use TMPDIR if defined

imefisto:
# Summary

This PR will allow to asdf to work in environments where the tmp dir is not /tmp and it is defined in the $TMPDIR variable. ie: [Termux](https://github.com/termux/termux-app).

### DIFF
--- a/lib/commands/reshim.bash
+++ b/lib/commands/reshim.bash
@@ -145,12 +145,15 @@ remove_obsolete_shims() {
   local formatted_shims
   local formatted_exec_names
 
+  local temp_dir
+  temp_dir=${TMPDIR:-/tmp}
+
   # comm only takes to files, so we write this data to temp files so we can
   # pass it to comm.
-  formatted_shims=$(mktemp /tmp/asdf-command-reshim-formatted-shims.XXXXXX)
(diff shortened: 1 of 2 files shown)


Stratus3D:
Hi @imefisto , thanks for the PR! And sorry for the late review here. I think these changes are a nice improvement. 

Do you think it would be safe to merge this as a `fix` instead of a `feat` so we can ship this with version `0.10.1` that will be tagged in a few weeks? `feat` means this would have to wait for `0.11.0`, which is probably many months away right now.

Also, some other PRs have been merged since yours was created. Do you want to rebase? If not, I can. Thanks!

imefisto:
Hi @Stratus3D ! Thanks for your response. I'll do as you recommended. Do you think is better if I create a new PR after rebase?

Stratus3D:
@imefisto it's whatever is easiest for you. Creating another PR won't bother me.

imefisto:
It's done!

Stratus3D:
Merged into master. Not sure why the PR didn't get closed so closing it manually. Thanks for the change @imefisto !

Issue #1180, opened 2022-03-08
bug: asdf latest --all report wrong runtimes


scalp42:
### Describe the Bug

Hi folks,

I've noticed a bug where `asdf latest --all` report the wrong runtimes version (added in https://github.com/asdf-vm/asdf/pull/1096 and fixed in https://github.com/asdf-vm/asdf/pull/1118).

Here's example:

```
$> asdf latest --all
packer	1.8.0	installed
python	stackless-3.7.5	missing
ruby	truffleruby+graalvm-21.3.0	missing
terraform	1.1.7	installed
```

If I try to run asdf latest individually, the right runtime is picked:

```
$> asdf latest python
3.10.2

$> asdf latest ruby
3.0.2
```

asdf version: `v0.9.0-7493f40`

Thanks in advance for looking into it and thank you for asdf.

### Steps to Reproduce

1. Install asdf
2. Add python and ruby plugins: `asdf plugin add python` & `asdf plugin add ruby`
3. Run `asdf latest --all`

### Expected Behaviour

It should have listed Python `3.10.2` and Ruby `3.0.2`.

### Actual Behaviour

It listed:

```
# python
stackless-3.7.5

# ruby
truffleruby+graalvm-21.3.0
```

### Environment

```shell
OS:
Darwin 42MBPM1X.local 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.8 (x86_64-apple-darwin21.0)

ASDF VERSION:
v0.9.0-7493f40

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/Users/scalp/.zinit/plugins/asdf-vm---asdf

ASDF INSTALLED PLUGINS:
bundler                      https://github.com/jonathanmorley/asdf-bundler.git master a461ac4
caddy                        https://github.com/salasrod/asdf-caddy.git master b864330
consul                       https://github.com/Banno/asdf-hashicorp.git master c53d4fd
crystal                      https://github.com/asdf-community/asdf-crystal.git master 0642d44
golang                       https://github.com/kennyp/asdf-golang.git master 85dab2f
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master 54113a9
nomad                        https://github.com/Banno/asdf-hashicorp.git master c53d4fd
[...]

PR #1191, opened 2022-03-30
fix: asdf latest --all report wrong runtimes (#1180)

roele:
# Summary

Revisites the logic for command `latest --all` which returns wrong version numbers if the plugin uses multiple distros/variants (e.g. Java, Python, Ruby, ...).

Instead of falling back to use the latest available version (`tail -1`) from all the available versions via `list all`, the code tries to list specific versions for installed distros. This is done by getting a distinct list of installed distros assuming the format <distro>-<version> and then getting the latest version for that specific distro via additional version query. If a plugin has no installed versions, the old behaviour showing the latest from all available versions is used.

Additionally if multiple distros are installed (e.g. Java; opendjdk, zulu) the output lists each installed one with its latest version and status.

```shell
  ❯ asdf list
direnv
  2.31.0
golang
  1.17
  1.18
java
  openjdk-11.0.2
  openjdk-17.0.2
  openjdk-18
  zulu-17.32.13
nodejs
  16.14.2
  lts
ruby
  No versions installed

  ❯ asdf latest --all
direnv              2.31.0                        installed
golang              1.18                          installed
java                openjdk-18                    installed
java                zulu-18.28.13                 missing
nodejs              lts                           installed
ruby                truffleruby+graalvm-22.0.0.2  missing
```

Fixes: #1180

### DIFF
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -140,7 +140,7 @@ latest_command() {
     versions=$("${plugin_path}"/bin/latest-stable "$query")
     if [ -z "${versions}" ]; then
       # this branch requires this print to mimic the error from the list-all branch
-      printf "No compatible versions available (%s %s)\\n" "$plugin_name" "$query" >&2
+      printf "No compatible versions available (%s %s)\n" "$plugin_name" "$query" >&2
       exit 1
     fi
   else
@@ -154,7 +154,7 @@ latest_command() {
     fi
   fi
 
-  printf "%s\\n" "$versions"
+  printf "%s\n" "$versions"
 }
 
 latest_all() {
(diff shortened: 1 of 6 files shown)


jvcastaneda (approved):

lib/commands/command-latest.bash
```diff
@@ -45,38 +45,75 @@ latest_all() {
...
+      for version in "${versions[@]}"; do
+        local installed_status
+        installed_status="missing"
 
-      local installed_versions
-      installed_versions=$(list_installed_versions "$plugin_name")
+        if [ -n "$installed_versions" ] && printf '%s\n' "$installed_versions" | grep -q "^$version\$"; then
+          installed_status="installed"
+        fi
+        printf "%-20s%-30s%-10s\\n" "$plugin_name" "$version" "$installed_status"
```
roele, line 114: I wonder if it's a good idea to introduce column output as this is not yet used anywhere and therefore adds inconsistency.

jthegedus: It might be useful to put this `printf` into a function to ensure consistency. We don't want to introduce a specific dependency to support columns as that was done in the past and caused issues.

Stratus3D:
Thanks for the PR and sorry for the late review here @roele. Would implementing the correct logic in a plugin's `latest-stable` callback fix this issue for plugins that have multiple distros/variants of the tool? This doesn't seem like something that should be handled by asdf core.

roele:
@Stratus3D Unfortunately I'm not familiar with this callback. As far as i have seen none of the plugins implement `latest-stable`. Isn't this callback supposed to return the latest stable version for a given query string. I don't see how this could help when someone has installations of multiple vendors (e.g. Java OpenJDK, Corretto, Microsoft). The callback would need to return the latest version for each of them and then the asdf core would still need to deal with handling those (comparing them against installed versions).

jthegedus:
@Stratus3D Your proposal would break `asdf install <tool> latest` usage. I personally expect the `<plugin>/bin/latest-stable` script to return a single version, even for a plugin like `java` which supports many versions. It is difficult to rely on `latest-stable` when plugins can support many distributions. Adding query string support to `latest-stable` may have been a mistake because we now tend to use it in places where we desire a query string filter but cannot know the appropriate filter, so have to fallback to `<tool>/bin/list-all`.

I think the solution here is the most appropriate.

`asdf latest --all` may have been a bad command to introduce as it is purely informational, difficult to implement and benefit not particularly significant or even clear.

test/fixtures/dummy_distro_plugin/bin/list-all
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+versions_list=(acme-1.0.0 acme-1.1.0 distro-2.0.0)
```
jthegedus, line 3: We should also test versions which have multiple prefix parts, EG: `zulu-jre-musl-8.42.0.23`

More examples from `asdf python`

```
pypy3.9-7.3.8
pypy3.9-7.3.9-src
pypy3.9-7.3.9
pyston-2.2
pyston-2.3
pyston-2.3.1
pyston-2.3.2
pyston-2.3.3
pyston-2.3.4
stackless-dev
stackless-2.7-dev
```

lib/functions/versions.bash
```diff
@@ -161,36 +161,72 @@ latest_all() {
...
+        # get latest version for each variant
+        for variant in $installed_variants; do
+          local version
+          version=$(list_all_command "$plugin_name" "$(echo "$variant" | tr -d '[:space:]')" |
+            grep -ivE "(^Available version:|-src|-dev|-latest|-stm|[-\\.]rc|-alpha|-beta|[-\\.]pre|-next|(a|b|c)[0-9]+|snapshot|master)" |
+            # filter for exact match or with version suffix to deal with overlaps (e.g. zulu-18.28.13, zulu-jre-18.28.13)
+            grep -iE "^($variant|$variant(-\d+.*)?)\$" |
+            sed 's/^[[:space:]]\+//' |
+            tail -1)
+
```
jthegedus, line 198: While shorter to write this way, we're calling `list_all_command` for each variant when really it only needs to be called once for the plugin. The `list_all_command` executes `<plugin>/bin/list-all` which performs a network call and file read/writes. It is usually a really slow command. Can you rework this some more?

test/fixtures/dummy_distro_plugin/bin/list-all
jthegedus:
```
pypy3.9-7.3.8
pypy3.9-7.3.9-src
```

IMO these are two different variants, though most querying added to commands are prefix-only queries, so I guess these should be resolved as a single variant.

roele: Good point, current implemenation does not yet handle suffixes. But i agree that this example should be treatet as a single variant. I would consider suffixes after the sematic version not as a variant but rather a special version like beta, GA, src or similar.

lib/functions/versions.bash
roele: Very Vvalid concern, will rework this part 👍

lib/functions/versions.bash
```diff
@@ -165,36 +165,73 @@ latest_all() {
...
+      local installed_versions
+      installed_versions=$(list_installed_versions "$plugin_name")
+
+      # Get a distinct list of distributions/variants
+      local installed_variants
+      installed_variants=$(printf "$installed_versions" |
+        sed -e "s/\(^.*\)\-[0-9].*/\1/" |   # extract variant prefix
+        sort -u |                           # unique
+        sed -e "s/^[[:space:]]*[0-9].*//" | # remove versions without variant prefix
+        sed '/^[[:space:]]*$/d')            # remove empty lines
```
Stratus3D, line 180: I'm uncomfortable with this being part of asdf. I know we have similar logic that already exists in a few places, but this doesn't feel like a future-proof solution. We're effectively using versions to store more than the version number. Versions are a combination of tool type (e.g. openjdk) and version (e.g. 1.2.3). I know this is a pretty fundamental change, but I think we might be better off either introducing a type field in addition to version, or specifying that plugins only support one type of thing (e.g. instead of a java plugin we have an openjdk plugin). Thoughts?

jthegedus: >specifying that plugins only support one type of thing
interesting proposal. I know this would also upset the Hashicorp plugin which manages the binaries for the Hashi ecosystem and not just different sources of the same runtime. I feel like Java and Hashicorp plugins are different enough and both valid in the spec we have defined 🤔

jthegedus:
As I stated earlier

> `asdf latest --all` may have been a bad command to introduce as it is purely informational, difficult to implement and benefit not particularly significant or even clear.

After further thought, I think the most beneficial thing to do is remove this `--all` flag from the `latest` command.
- #1036 was initial `latest --all` discussion
- #1096 was initial PR
- #1118 fixed a bug in initial implementation
- #1180 is another discussion of how this is not right
- #1191 (this PR) is a fix for #1180

The Issues discussing the desire for the command and the current obtuse behaviour have a total of 3 participants (outside asdf core team) and seemingly no support for this behaviour (use Issue/PR GitHub emojis as the metric).

I propose we remove it.

@Stratus3D Do you have any strong feelings either way?

@roele I'm sure hearing this is unpleasant given your input on this PR was significant. I am sorry. Are there points I am missing which might convince me otherwise?

roele:
@jthegedus Even though i would love the convenience of a single command telling me what packages are outdated,  I am fine with that decision and understand that maintainability has upmost priority.

jthegedus:
> @jthegedus Even though i would love the convenience of a single command telling me what packages are outdated, I am fine with that decision and understand that maintainability has upmost priority.

I agree on the convenience. Perhaps this is a good candidate for a one-line shell cmd to demonstrate in a cookbook/recipes section in the docs.

Until the points raised above by @Stratus3D are addressed I don't think this passes any cost/benefit analysis.

I will close this PR. I will raise a PR to back out the `latest --all` feature. 

Thanks for your contributions here and I hope you contribute again :pray:

Issue #1152, opened 2022-01-06
Highlight or mark installed versions when running `asdf list all <name>`


MitchellCash:
### Is your feature request related to a problem? Please describe

When I run `asdf list all <name>` it is useful for me to compare my installed versions against the list of available versions.

For example, I currently have installed `nodejs` versions `14.18.1` and `16.13.1`. I run `asdf list all nodejs` to see if there is a new minor or patch version bump for one of my installed major versions (i.e. `14.x.x` or `16.x.x`). 

I don't always remember exactly what versions I have installed and want an easy way to visually understand that `14.18.2` is now available for me to install and `16.13.1` is still the latest.

### Describe the proposed solution

Something like:

```sh
$ asdf list all nodejs
...
14.18.0
14.18.1 (installed)   <-- either differentiate with text or colour highlight it
14.18.2
...
16.13.0
16.13.1 (installed)   <-- either differentiate with text or colour highlight it
lts-gallium
...
```

### Describe similar `asdf` features and why they are not sufficient

Running `asdf list <name>` and then `asdf list all <name>` is cumbersome. 

### Describe other workarounds you've considered

As above, running `asdf list <name>` and then `asdf list all <name>` is cumbersome and can mean scrolling back and forth to compare the two lists.

roele:
That would come in handy indeed.

Note that the command `asdf latest --all` will show you a list of the latest versions for each plugin with its installation status. The `--all` option is as of yet only available in the `--HEAD` version and has a bug (#1180) related to variants.

jthegedus:
@MitchellCash Can I ask why `asdf install nodejs latest 14` is not sufficient here?

If it installs, you know there was a new version, you then know what that version is, and you are updated to it immediately. If the resolved version already exists on your machine it doesn't install it again, so you know there was no new version for your filter.

PR #1192, opened 2022-04-01
fix: Highlight or mark installed versions when running asdf list all <name>

roele:
# Summary

Color highlight installed versions of the `list-all` command.

> Note: It seems asdf does not use colors anywhere, hence we might decide to use another indicator (e.g. `▶ `). Simply appending `(installed)` is visually not distinctive enough IMO. If an icon like indicator is used we might have to think about consistent output formatting. 

Uses the internal function `list_installed_versions` to get all installed versions and adds a color highlight during output iteration.

Since the command `latest [--all]` uses the command `list-all` as fallback to get versions, any output format change breaks this functionality. Hence i decided to create an (internal) function `list_all_versions` which can be used by both commands. 

The change lead the test `[latest_command - dummy_legacy_plugin] No stable version should return an error` to fail which i could not figure out why. It's expecting no output even though the code actually outputs an error `"No compatible versions available ($plugin_name $query)"`. For now i came to the conclusion that maybe previous behaviour might have been faulty.

Fixes: #1152

### DIFF
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -106,8 +106,16 @@ list_all_command() {
 
   IFS=' ' read -r -a versions_list <<<"$output"
 
+  mapfile -t installed_versions < <(list_installed_versions "$plugin_name")
+
   for version in "${versions_list[@]}"; do
-    printf "%s\\n" "${version}"
+    # exact match version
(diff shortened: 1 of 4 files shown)


roele:
Removed colored output, fixed tests, shellcheck and shfmt.

Issue #952, opened 2021-05-22
Ability to specify a different default plugins repository


adongy:
**Is your feature request related to a problem? Please describe.**

Imagine a workplace where plugins should be internalized to assume full control before being usable. Users should only install these vetted plugins. Currently, users have to specify a custom plugin URL for each of these plugins.

**Describe the solution you'd like**
I'd like to be able to change the default repository URL for plugins, currently located in https://github.com/asdf-vm/asdf/blob/0a3865d0b1f8adad4f4d612869abd801e1c2a27d/lib/utils.bash#L35-L37

**Describe similar asdf features and why they are not sufficient**

The plugin list may list plugins that are actually not allowed.

**Additional context**

It could be a setting in the .asdfrc. 

Thanks,

jthegedus:
>Currently, users have to specify a custom plugin URL for each of these plugins

This is preferred behaviour as it is explicit and users are informed that plugins are merely git repos and not magic of asdf.

I think this should be a setting. `.asdfrc` is the place to have it.

I would like to note that the plugin repo format is subject to change which would require updates to external plugin repos.

adongy:
Yes I believe .asdfrc is a good place for it.

The repository URL could be printed alongside the current logs to ease debug and avoid people reporting issues when using a custom repository.

https://github.com/asdf-vm/asdf/blob/32371673384ae7ddba3532670d6c04b3f7dce249/lib/utils.bash#L393

As to the repo format, that would mean format update need to be documented so that plugin repository operators can update the format accordingly.

jthegedus:
>The repository URL could be printed alongside the current logs to ease debug and avoid people reporting issues when using a custom repository

Yes, that is a good suggestion.

>As to the repo format, that would mean format update need to be documented so that plugin repository operators can update the format accordingly

Yes, we should document it when we get to a stable release. Adding this as a configurable option would mean we are essentially expanding an API scope. We are using semver and on version `0.x.x`, which means our feature-set and API are not yet set in stone.

jthegedus:
After thinking about this more I don't think the idea of custom plugins repo is a good idea.

>Currently, users have to specify a custom plugin URL for each of these plugins

This is the preferred user flow. The plugin repo only saves users typing/copy-pasting the `<git-url>` portion of the command, so that isn't a valid reason IMO. If it's for discoverability, there are many other solutions to just sharing URLs amongst your project/team.

`asdf plugin add <name> <git-url>` is preffered because:
- it always works
[...]

PR #1197, opened 2022-04-07
docs: Add link to central plugin repository

mrzool:
# Summary

The question I had in my mind while skimming the Getting Started page was: "What can I use this for?". I had to Google and click around a bit to find the answer — adding a quick note with a link in the appropriate section of the guide fixes that.

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -245,7 +245,9 @@ Restart your shell so that `PATH` changes take effect. Opening a new terminal ta
 
 This completes the installation of the `asdf` core :tada:
 
-`asdf` is only useful once you install a **plugin**, install a **tool** and manage its **versions**. Continue the guide below to learn how to do this.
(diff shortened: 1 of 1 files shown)


Stratus3D:
HI @mrzool ! Thank you for the PR, however we cannot merge this as many of the maintainers were not on board with the idea of a central plugin repository (see https://github.com/asdf-vm/asdf/issues/952)

```
You can also head to [this page](https://github.com/asdf-vm/asdf-plugins) for an up-to-date list of all supported plugins.
```

This is exactly why we oppose the idea. There is no such thing as an up-to-date list of supported plugins. None of the plugins in the list (except 4) are supported by us, and we provide no guarantees for any of the others.

A better solution here is to simply do a Google or Github search to find the plugins you need, and only install plugins from people you trust. Google works really well `asdf <what you want to version>` e.g. `asdf kubernetes`.

Thanks for taking the time to create this PR, but hopefully you can understand why we aren't going to accept it.

mrzool:
Hi @Stratus3D, thanks for the explanation and no worries! I get your reasoning.

PR #1198, opened 2022-04-07
fix: remove comments from whole file instead of line by line for performance

### LINKED ISSUE #1195: bug: strip_tool_version_comments could be faster for a large amount of comments
### Describe the Bug

I'm not sure if this is a bug or feature, but I noticed the `strip_tool_version_comments` performance could be improved.

The current code is: 
```
strip_tool_version_comments() {
  local tool_version_path="$1"

  while IFS= read -r tool_line || [ -n "$tool_line" ]; do
    # Remove whitespace before pound sign, the pound sign, and everything after it
    new_line="$(cut -f1 -d"#" <<<"$tool_line" | sed -e 's/[[:space:]]*$//')"

    # Only print the line if it is not empty
    if [[ -n "$new_line" ]]; then
      printf "%s\\n" "$new_line"
    fi
  done <"$tool_version_path"
}
```
This iterates over every line in `.tool-versions`, then shells out to `cut` and `sed` for each line. I looked around a bit, and it seems like this loop can be replaced with one call to `sed` for the whole file (found here: https://unix.stackexchange.com/a/157619, which has a good breakdown of what the command is doing). The alternate version is:

```
strip_tool_version_comments() {
  local tool_version_path="$1"
  sed '/^[[:blank:]]*#/d;s/#.*//' "$tool_version_path"
}
```

This should handle lines that are only comments and comments at the end of tool version lines, like the original code. As for performance, I tested with a `.tool-versions` file with 50 lines of comments (excessive, but not unrealistic) and the timing worked like:

Before change: 0.22s user 0.69s system 120% cpu 0.762 total
After change: 0.07s user 0.21s system 99% cpu 0.276 total

After a few more tests, it looks like the original code slows a bit with every line added, where the 1 `sed` version stays constant. The proposed version is even around 10ms faster for a `.tool-versions` with only one line in it.

If this change seems interesting I can go ahead and make a PR. Let me know!

### Steps to Reproduce

1. Create a `.tool-versions` file with 50 lines of comments.
2. Run `time asdf current`.
3. See that it takes ~200ms.

### Expected Behaviour

It should not take any more time than a file with no comments.

### Actual Behaviour

It gets slower with every comment in the file.

### Environment

```shell
OS:
Darwin blopker-pro 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.8 (x86_64-apple-darwin21.0)

ASDF VERSION:
v0.9.0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/opt/homebrew/opt/asdf/libexec

ASDF INSTALLED PLUGINS:
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master 1bc55cb
python                       https://github.com/danhper/asdf-python.git master 57a4d72
shellcheck                   https://github.com/luizm/asdf-shellcheck.git master 31cb328
shfmt                        https://github.com/luizm/asdf-shfmt.git master f893252
```
```


_No response_

blopker:
# Summary

This change uses 1 call to `sed` for the whole `.tool-versions` file instead of for each line. This can substantially speed up file parsing if there are many comments in a file. I also had an issue with the current root `.tool-verisons` file not installing `bats` correctly because of the leading `v`.

This fixes issue #1195.

Changes:
* Remove comments from whole `.tool-versions` at once instead of line by line for performance
* Remove the leading `v` from the root `.tool-versions` file to help asdf resolve it correctly.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-bats v1.3.0
+bats 1.3.0
 shellcheck 0.7.2
 shfmt 3.3.0
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -637,16 +637,14 @@ shim_plugins() {
 
 strip_tool_version_comments() {
   local tool_version_path="$1"
-
-  while IFS= read -r tool_line || [ -n "$tool_line" ]; do
-    # Remove whitespace before pound sign, the pound sign, and everything after it
-    new_line="$(cut -f1 -d"#" <<<"$tool_line" | sed -e 's/[[:space:]]*$//')"
-
-    # Only print the line if it is not empty
-    if [[ -n "$new_line" ]]; then
-      printf "%s\\n" "$new_line"
-    fi
-  done <"$tool_version_path"
+  # Use sed to strip comments from the tool version file
+  # Breakdown of sed command:
+  # This command represents 3 steps, seperated by a semi-colon (;), that run on each line.
+  # 1. Delete line if it starts with any blankspace and a #.
+  # 2. Find a # and delete it and everything after the #.
+  # 3. Remove any whitespace from the end of the line.
+  # Finally, the command will print the lines that are not empty.
+  sed '/^[[:blank:]]*#/d;s/#.*//;s/[[:blank:]]*$//' "$tool_version_path"
 }
 
 asdf_run_hook() {


Stratus3D:
This is excellent @blopker !

Stratus3D (approved):

PR #1202, opened 2022-04-12
local plugin in then clause too

znz:
# Summary

I think `plugin` variable should be `local` not only in `else` clause but `then` clause too.


Stratus3D:
Thanks for the PR @znz ! These changes look good, however the commit is failing the conventional commit check (https://github.com/asdf-vm/asdf/runs/5987076283?check_suite_focus=true). See https://www.conventionalcommits.org/en/v1.0.0/ for proper formatting. If you are unable to amend your comment message let me know and I can amend it for you.

Stratus3D:
Closed in favor of https://github.com/asdf-vm/asdf/pull/1203

PR #1209, opened 2022-04-20
Making fish setup match other shells

sjml:
# Summary

The asdf.sh file only sets `ASDF_DIR` if it hasn't already been set. This change to the fish file has it do the same.


sjml:
I'm not entirely clear on what needs to be done to pass that lint check... please advise!

Stratus3D (approved):

Stratus3D:
Thanks for the PR @sjml ! The changes here look good to me.

To fix the lint check you'll need to follow the commit message formatting documented here - https://www.conventionalcommits.org/en/v1.0.0/

For your commit I'd recommend amending it to:

```
fix: make fish shell setup match other shells

The asdf.sh file only sets `ASDF_DIR` if it hasn't already been set. This change to the fish file has it do the same.
```

Let me know if you need any help amending it.

sjml:
> Let me know if you need any help amending it.

@Stratus3D I amended the commit message but I guess it also wants the PR title to change?

PR #1217, opened 2022-04-27
fix: more robust SHELL checks

jacobbednarz:
While debugging an unrelated issue, I found that my `$SHELL --version`
output was returning as zsh but I actually run `fish` via nix. Instead
of using `$SHELL`, which isn't accurate in some cases, I've updated this
to attempt to use the various `*_VERSION` environment variables that are
available in the shells before failing back to the previous `$SHELL`
check. Realistically, `$SHELL --version` should be replaced with
something else but this should improve the current situation with
misidentifying shells in use.

### DIFF
--- a/lib/commands/command-info.bash
+++ b/lib/commands/command-info.bash
@@ -2,9 +2,21 @@
 # shellcheck source=lib/functions/plugins.bash
 . "$(dirname "$(dirname "$0")")/lib/functions/plugins.bash"
 
+get_shell() {
+  if [[ -z $FISH_VERSION ]]; then
+    echo "fish"
+  elif [[ -z $ZSH_VERSION ]]; then
+    echo "zsh"
+  elif [[ -z $BASH_VERSION ]]; then
+    echo "bash"
+  else
+    echo $SHELL --version
+  fi
+}
+
 info_command() {
   printf "%s:\\n%s\\n\\n" "OS" "$(uname -a)"
-  printf "%s:\\n%s\\n\\n" "SHELL" "$($SHELL --version)"
+  printf "%s:\\n%s\\n\\n" "SHELL" "$(get_shell)"
   printf "%s:\\n%s\\n\\n" "ASDF VERSION" "$(asdf_version)"
   printf "%s:\\n%s\\n\\n" "ASDF ENVIRONMENT VARIABLES" "$(env | grep -E "ASDF_DIR|ASDF_DATA_DIR|ASDF_CONFIG_FILE|ASDF_DEFAULT_TOOL_VERSIONS_FILENAME")"
   printf "%s:\\n%s\\n\\n" "ASDF INSTALLED PLUGINS" "$(plugin_list_command --urls --refs)"
--- a/lib/commands/command-info.bash
+++ b/lib/commands/command-info.bash
@@ -4,11 +4,11 @@
 
 get_shell() {
   if [[ -z $FISH_VERSION ]]; then
-    echo "fish, version $FISH_VERSION"
+    printf "fish, version %s" "$FISH_VERSION"
   elif [[ -z $ZSH_VERSION ]]; then
-    echo "zsh, version $ZSH_VERSION"
+    printf "zsh, version %s" "$ZSH_VERSION"
   elif [[ -z $BASH_VERSION ]]; then
-    echo "bash, $BASH_VERSION"
+    printf "bash, version %s" "$BASH_VERSION"
   else
     $SHELL --version
   fi
--- a/lib/commands/command-info.bash
+++ b/lib/commands/command-info.bash
@@ -10,7 +10,7 @@ get_shell() {
   elif [[ -z $BASH_VERSION ]]; then
     echo "bash, $BASH_VERSION"
   else
-    echo $($SHELL --version)
+    $SHELL --version
   fi
 }
 
--- a/lib/commands/command-info.bash
+++ b/lib/commands/command-info.bash
@@ -4,13 +4,13 @@
 
 get_shell() {
   if [[ -z $FISH_VERSION ]]; then
-    echo "fish"
+    echo "fish, version $FISH_VERSION"
   elif [[ -z $ZSH_VERSION ]]; then
-    echo "zsh"
+    echo "zsh, version $ZSH_VERSION"
   elif [[ -z $BASH_VERSION ]]; then
-    echo "bash"
+    echo "bash, $BASH_VERSION"
   else
-    echo $SHELL --version
+    echo $($SHELL --version)
   fi
 }
 


lib/commands/command-info.bash
```diff
@@ -2,9 +2,21 @@
...
 
+get_shell() {
+  if [[ -z $FISH_VERSION ]]; then
+    echo "fish"
+  elif [[ -z $ZSH_VERSION ]]; then
+    echo "zsh"
+  elif [[ -z $BASH_VERSION ]]; then
+    echo "bash"
+  else
+    echo $SHELL --version
```
Stratus3D, line 13: Currently the code invokes `$SHELL --version`, but this line actually just prints it, so we'd get `zsh --version` instead of `5.8`.

lib/commands/command-info.bash
```diff
@@ -2,9 +2,21 @@
 # shellcheck source=lib/functions/plugins.bash
 . "$(dirname "$(dirname "$0")")/lib/functions/plugins.bash"
 
+get_shell() {
+  if [[ -z $FISH_VERSION ]]; then
+    echo "fish"
```
Stratus3D, line 7: The current code would print something like `fish, version 3.3.1`, whereas this only prints `fish`. I think both version and type are useful.

lib/commands/command-info.bash
```diff
@@ -2,9 +2,21 @@
...
+    echo "bash"
+  else
+    echo $SHELL --version
+  fi
+}
+
 info_command() {
   printf "%s:\\n%s\\n\\n" "OS" "$(uname -a)"
-  printf "%s:\\n%s\\n\\n" "SHELL" "$($SHELL --version)"
+  printf "%s:\\n%s\\n\\n" "SHELL" "$(get_shell)"
```
Stratus3D, line 19: I think we may want to report both `$SHELL --version` and whatever shell and version is computed by your `get_shell` function above.

lib/commands/command-info.bash
jacobbednarz: ah good catch! addressed in [91527a3](https://github.com/asdf-vm/asdf/pull/1217/commits/91527a3df04b44a69df9de8a9789920ef42ecb0b)

lib/commands/command-info.bash
jacobbednarz: sure thing, i've updated the output in [91527a3](https://github.com/asdf-vm/asdf/pull/1217/commits/91527a3df04b44a69df9de8a9789920ef42ecb0b)

lib/commands/command-info.bash
jacobbednarz: sure thing, i've updated the output in [91527a3](https://github.com/asdf-vm/asdf/pull/1217/commits/91527a3df04b44a69df9de8a9789920ef42ecb0b)

lib/commands/command-info.bash
Stratus3D: Sorry, I should have been more specific in my comment. Something like this:

```
...
printf "%s:\\n%s\\n\\n" "SHELL" "$($SHELL --version)"
printf "%s:\\n%s\\n\\n" "COMPUTED_SHELL" "$(get_shell)"
...
```

So that we print both `$SHELL --version` and whatever your `get_shell` function determines side by side. Sometimes the values will match, sometime they won't.

lib/commands/command-info.bash
```diff
@@ -2,9 +2,21 @@
...
 
+get_shell() {
+  if [[ -z $FISH_VERSION ]]; then
+    echo "fish, version $FISH_VERSION"
+  elif [[ -z $ZSH_VERSION ]]; then
+    echo "zsh, version $ZSH_VERSION"
+  elif [[ -z $BASH_VERSION ]]; then
+    echo "bash, $BASH_VERSION"
+  else
+    echo $($SHELL --version)
```
Stratus3D, line 13: I believe this line can be simplified to `$SHELL --version` since the shell will print the version anyway.

lib/commands/command-info.bash
jacobbednarz: i'm happy to add this however, i'm unsure what you'll actually get from it, except for confusion. take my example, I don't use zsh (it is on Mac by default) yet, the variable was incorrectly suggesting it was. asdf isn't installed in zsh nor configured so it is a red herring for any diagnosis related. 

is there a particular reason you'd like this to stick around?

lib/commands/command-info.bash
jacobbednarz: like I had previously?

Stratus3D: Previously this line was:

```
echo $SHELL --version
```

Now it is (and it behaves correctly):

```
echo $($SHELL --version)
```

But it can be simplified to just:

```
$SHELL --version
```

jacobbednarz: gotcha, that is clearer! thanks.

lib/commands/command-info.bash
Stratus3D: So my thinking was this if we have two imperfect ways of determining the current shell, why not just print both? If they are the same, it's confirmation that they both are correct. If they differ, we can inquire as to why.

jacobbednarz: gotcha, so the ordering of the `if/else` _should_ prevent the imperfect approach as if the variable is found, the shell is actually in use. if none of those are found, fallback to the best effort `$SHELL --version`. i've tested this on fish, zsh, bash and elvish locally and all of them returned the correct shell each time instead of falling back to the `$SHELL --version` that reports incorrectly.

Stratus3D (approved):

Stratus3D:
I was going to merge but then I realized test jobs are failing due to use of the `echo` command. We had issues with `echo` in the past and mandated `printf` everywhere as a result. I can fix this before merging your PR.

Stratus3D:
Testing locally this prints:

```
SHELL:
fish, version 
```

When I am on zsh (which is also my `$SHELL`). It appears this is due to the `-z` flag in the if statements. I think you probably wanted to use `-n` there, which checks if the variable is not empty.

Stratus3D:
Actually, printing all three version variables reveals that on my system `ZSH_VERSION` is never set inside the script despite being the shell I use and also being `$SHELL`. And `BASH_VERSION` is always set as bash is what is executing the script.

jacobbednarz:
> I was going to merge but then I realized test jobs are failing due to use of the echo command. We had issues with echo in the past and mandated printf everywhere as a result. I can fix this before merging your PR.

sure thing, let me update.

> Actually, printing all three version variables reveals that on my system `ZSH_VERSION` is never set inside the script despite being the shell I use and also being `$SHELL`. And `BASH_VERSION` is always set as bash is what is executing the script.

This is very interesting and different behaviour to what I'm seeing 😢 

```
$ zsh
jacob@C02DG4A3ML85 asdf % echo $ZSH_VERSION
5.8
jacob@C02DG4A3ML85 asdf % echo $FISH_VERSION

jacob@C02DG4A3ML85 asdf % echo $BASH_VERSION

jacob@C02DG4A3ML85 asdf % exit
```

```
$ bash
bash-4.4$ echo $FISH_VERSION

bash-4.4$ echo $ZSH_VERSION

bash-4.4$ echo $BASH_VERSION
4.4.23(1)-release
bash-4.4$ exit
```

```
~/s/asdf (better-shell-checks)$ echo $BASH_VERSION

~/s/asdf (better-shell-checks)$ echo $ZSH_VERSION

~/s/asdf (better-shell-checks)$ echo $FISH_VERSION
3.3.1
```

Stratus3D:
@jacobbednarz it looks like you are echoing those variables in your shell (Zsh, Bash, Fish). But the asdf info script script isn't executed by your current shell, it's always executed by Bash (since it is written in Bash). Can you try echo those inside the script?

jacobbednarz:
ah damn, good find. that is indeed a problem here :/ let me think about this as it may need more to determine the default shell.

Stratus3D:
No worries @jacobbednarz , take all the time you need.

I'm beginning to think we may not be able to improve much upon `$SHELL`, but if you can come up with something better I'll get it merged.

jacobbednarz:
ok, so after a little digging, i know _why_ this is happening and lo and behold, it was my doing 🤦  

on Mac, i use iterm2 which allows you to define a shell or command for startup. unknown to present me, past me used fish as the startup command, not the actual shell. so while everything worked as expected, it *wasn't actually* my shell. once i set the startup shell to fish, this works as expected.

for anyone that stumbles on a similar issue in the future, first confirm that your shell is actually what you think it is as `echo $SHELL` works providing you actually have it set as your login shell.

Issue #1194, opened 2022-04-05
Allow configuring `asdf` to use a custom plugin repository


shakefu:
### Is your feature request related to a problem? Please describe

Yes, right now `asdf` has hard coded the plugin repository which maps names to repository URLs.

https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash#L36

This means it cannot be used in high-security or airgapped environments.

### Describe the proposed solution

Allowing the plugin repository to be configured through environment variable would be trivial, or perhaps through a new setting in `~/.asdfrc`.

### Describe similar `asdf` features and why they are not sufficient

N/A

### Describe other workarounds you've considered

Create a fork of `asdf-vm/asdf` and edit it directly.

hmnd:
Though I know this doesn't entirely solve the issue, you can already provide your own Git URL to a plugin without having to go through the centralized repo. See [here](https://asdf-vm.com/manage/plugins.html#add).

shakefu:
> here

This is true, but my specific use case includes being able to specify `.tool-versions`, without knowing ahead of time which tools are going to be installed, and which repository URIs should be added for the plugins. This means it will use the default plugin name mapping instead, which is what I'd like to override. 

Of course, it would be fairly trivial to `cat .tool-versions | awk '{print $1}' | xargs -I{} asdf plugin add {} github.com/myorg/asdf-{}` but ... that's exactly what the default plugin repository is doing...

hmnd:
This might interest you too :) https://github.com/asdf-vm/asdf/issues/240#issuecomment-1048680932

shakefu:
TBH my biggest problem with `asdf` as a user (use cases above not withstanding) is actually finding plugins for various tools that aren't in the short-name "plugin repo".  Regular search engines and 

I still think there's a strong use case in high security/air-gapped environments to _only_ allow short-name plugins from a pre-approved/security vetted "plugin repo" list, but if it's not part of the intended long term road map, obviously you can close this.

Thanks for the replies. 👍

jthegedus:
>I still think there's a strong use case in high security/air-gapped environments to only allow short-name plugins from a pre-approved/security vetted "plugin repo" list, but if it's not part of the intended long term road map, obviously you can close this.

[...]

PR #1227, opened 2022-05-10
feat: Disable short-name repository with config value

jrbeverly:
# Summary

Support disabling the installation & updating of the short-name repository using the configuration value `disable_short_name_repository`. This does not remove any existing cache/copy of the short-name repository.

This change allows the short-name plugins repository to be disabled in distributed zip files (or existing installations) by setting the the configuration field `disable_short_name_repository` as `yes`. This has the effect of disabling `asdf plugin add <name>`, as no short-name repository will be configured to allow installation without a URL.

Fixes: #1194 

## Motivation

One of the issues that I've encountered with `asdf` is in the compliance/security realm. Since the [short-name repository](https://github.com/asdf-vm/asdf-plugins/blob/master/plugins/awscli) is a map of git repositories for lookup, it hides the true source of how a plugin is installed & works. This can result in confusion scenarios where someone wasn't aware they were running bash scripts from an unknown github repository, of which they were working off latest.

Documentation can always recommend running `asdf plugin add <name> <git-url>` (where `git-url` is from a known 👍 set), but it is a bit of a concern point about "what ifs" and all that.

For this change I wanted to allow the plugin repository to be disabled, requiring `asdf plugin add <name> <git-url>` in a way that was a minimally disruptive to the existing workflows that might rely on `asdf plugin add <name>`.

## Other Information

For this change I was looking for a way to disable the default short-name plugins repository, so that `asdf plugin add <name> <git-url>` was the only way to install a plugin. I think completely removing the short-name plugins repository would be a major breaking change that would negatively impact a lot of workflows. So this change focuses on allowing it to be disabled.

~An opt-out system (e.g. `disable_asdf_plugins_repository`) would accomplish this too, but a bug or misconfiguration could result in the `asdf plugin add <name>` functionality becoming enabled. By using the `asdf_repository_url`, as long as the distributed `asdf.tar.gz` has the `defaults` file patched to remove this configuration, then the short-name repository should never be configured.~

~Distributing the `asdf.tar.gz` with the git URL patched (or the `/repository` directory pre-configured) is a way of disabling the short-name repository at the moment.~

~A concern with this change is that the introduction of a configuration field like `asdf_repository_url` could/may/will give the false(?) impression that providing a custom short-name plugins repository is a supported feature (with associated overhead of defining API/etc/etc), which isn't the intention of this change (e.g. [`asdf plugin add <name> <git-url>`](https://github.com/asdf-vm/asdf/issues/952#issuecomment-853436831)). I'm assuming based on that previous conversation that it isn't your intention to support that, so I have added to the docs accordingly. If that isn't the case, then this bit can be ignored (& related `warning` in docs removed)~

Previous discussions: #952


lib/utils.bash
```diff
@@ -404,7 +400,11 @@ initialize_or_update_repository() {
   local repository_url
   local repository_path
 
-  repository_url=$(asdf_repository_url)
+  repository_url="$(get_asdf_config_value "asdf_repository_url")"
+  if [ ! -n "$repository_url" ]; then
```
jrbeverly, line 404:
```suggestion
  if [ ! -z "$repository_url" ]; then
```

lib/utils.bash
```diff
@@ -404,7 +400,11 @@ initialize_or_update_repository() {
   local repository_url
   local repository_path
 
-  repository_url=$(asdf_repository_url)
+  repository_url="$(get_asdf_config_value "asdf_repository_url")"
+  if [ ! -z "$repository_url" ]; then
```
jrbeverly, line 404: Logic of this is the wrong way, should be `-z`

```suggestion
  if [ -z "$repository_url" ]; then
```

jthegedus:
I will perform final review of this soon.

jthegedus:
@jrbeverly Thanks for reworking this PR 🙏

I made some additional changes:
- renamed the option to the more verbose `disable_plugin_short_name_repository`
- wrote a test for `asdf plugin list all` code path with this config
- simplified some of the test code to match existing test pattern
- updated the docs for "Syncing the asdf Short-name Repository"
- updated the example config with the default value "no" instead of the optional "yes" to match rest of examples

Can you give this a once over before I merge?

jthegedus (approved):

jrbeverly:
LGTM, the doc improvements talking about sync events make a lot of sense 👍

jthegedus:
Awesome, thanks for your contribution @jrbeverly

Issue #1231, opened 2022-05-19
bug: asdf 0.10.1 can't find the commands directory on linux or macos using homebrew and fish


nathanpalmer:
### Describe the Bug

I just installed asdf through homebrew on linux using the fish shell. When I run `asdf` it tells me this

```
/home/linuxbrew/.linuxbrew/Cellar/asdf/0.10.1/libexec/bin/asdf: line 82: /home/linuxbrew/.linuxbrew/opt/asdf/lib/commands/command-help.bash: No such file or directory
```

the lib directory just has two files in it, no subdirectories. `ls -l /home/linuxbrew/.linuxbrew/opt/asdf/lib/`

```
asdf.fish
asdf.sh
```

I can tell that `/home/linuxbrew/.linuxbrew/opt/asdf/libexec/lib` has a `commands` directory, but it's not looking in the libexec subfolder.

I can get this to work by editing

**libexec/bin/asdf**

adding libexec to the asdf_cmd_dir

```
find_asdf_cmd() {
  local asdf_cmd_dir
  asdf_cmd_dir="$(asdf_dir)/lib/commands"
```

```
find_asdf_cmd() {
  local asdf_cmd_dir
  asdf_cmd_dir="$(asdf_dir)/libexec/lib/commands"
```

and then editing `libexec/lib/commands/command-help.bash`

adding libexec to the help.txt path

```
asdf_help() {
  printf "version: %s\\n\\n" "$(asdf_version)"
  cat "$(asdf_dir)/help.txt"
}
```

```
asdf_help() {
  printf "version: %s\\n\\n" "$(asdf_version)"
  cat "$(asdf_dir)/libexec/help.txt"
}
```

and then editing `libexec/lib/utils.bash`

adding libexec to the version.txt path

```
asdf_version() {
  local version git_rev
  version="v$(cat "$(asdf_dir)/version.txt")"
  if [ -d "$(asdf_dir)/.git" ]; then
    git_rev="$(git --git-dir "$(asdf_dir)/.git" rev-parse --short HEAD)"
    printf "%s-%s\\n" "$version" "$git_rev"
  else
    printf "%s\\n" "$version"
  fi
}
```

```
asdf_version() {
  local version git_rev
  version="v$(cat "$(asdf_dir)/libexec/version.txt")"
  if [ -d "$(asdf_dir)/.git" ]; then
    git_rev="$(git --git-dir "$(asdf_dir)/.git" rev-parse --short HEAD)"
[...]

Issue #531, opened 2019-05-29
Reshim doesn't update asdf path


pikeas:
### Steps to reproduce

```
$ which asdf
/usr/local/Cellar/asdf/0.7.2/bin/asdf

$ python
/Users/aris/.asdf/shims/python: line 3: /usr/local/Cellar/asdf/0.7.1/bin/asdf: No such file or directory
/Users/aris/.asdf/shims/python: line 3: exec: /usr/local/Cellar/asdf/0.7.1/bin/asdf: cannot execute: No such file or directory

$ cat ~/.asdf/shims/python
#!/usr/bin/env bash
# asdf-plugin: python 3.7.3
exec /usr/local/Cellar/asdf/0.7.1/bin/asdf exec "python" "$@"

$ asdf reshim python

$ cat ~/.asdf/shims/python
#!/usr/bin/env bash
# asdf-plugin: python 3.7.3
exec /usr/local/Cellar/asdf/0.7.1/bin/asdf exec "python" "$@"
```

### Expected behavior

`asdf reshim python` should update the asdf path from 0.7.1 to 0.7.2

### Actual behavior

The shim file was unchanged.

### Environment

**OS**: OS X 10.14

**asdf version**: 0.7.2

pikeas:
This happens because my shell sourced asdf from `/usr/local/Cellar/asdf/<version>` instead of `brew --prefix asdf`. Updating the source command to the latter, and running `rm -rf ~/.asdf/shims && asdf reshim` recreates the shims with the correct exec path.

pikeas:
Re-opening this issue in case because a case can be made for `reshim` updating the exec path when it changes.

Stratus3D:
@pikeas we haven't had anyone else report this issue recently, can you check and see if this is still a problem on the latest version of asdf and asdf-python? Thanks!

pikeas:
Just checked, reshim does not update when the exec command changes.

```
$ grep exec ~/.asdf/shims/python
exec /usr/local/opt/asdf/bin/asdf exec "python" "$@"

$ asdf reshim

$ grep exec ~/.asdf/shims/python
exec /usr/local/opt/asdf/bin/asdf exec "python" "$@"

$ grep asdf ~/.config/fish/config.fish
source /usr/local/Cellar/asdf/0.8.0/asdf.fish

$ rm ~/.asdf/shims/python
$ asdf reshim

$ grep exec ~/.asdf/shims/python
[...]

PR #1236, opened 2022-05-24
fix: Remove static paths to asdf for shims

jrogov:
# Summary
Switch from generating static paths to asdf exec file to using user's ASDF_DIR
env variable to dynamically determine asdf location.

It ensures that regardless of updates for asdf installed on user system path
won't rely on single correct path to asdf. Since the path might contain version
number (e. g. `/opt/homebrew/Cellar/asdf/0.9.0/libexec/bin/asdf`), asdf
installation would break as soon as new release is installed.

Fixes: #1231

### DIFF
--- a/lib/commands/reshim.bash
+++ b/lib/commands/reshim.bash
@@ -89,7 +89,9 @@ write_shim_script() {
   shim_path="$(asdf_data_dir)/shims/$executable_name"
 
   # Temporary fix for #1231
+  # shellcheck disable=SC2016
   if ! grep '^exec \${ASDF_DIR}' "$shim_path" >/dev/null; then
+    # shellcheck disable=SC2016
     sed -i.bak -e 's/^exec .*\/bin\/asdf /exec ${ASDF_DIR}\/bin\/asdf /' "$shim_path"
     rm -rf "$shim_path".bak
   fi
--- a/lib/commands/reshim.bash
+++ b/lib/commands/reshim.bash
@@ -88,6 +88,12 @@ write_shim_script() {
   local shim_path
   shim_path="$(asdf_data_dir)/shims/$executable_name"
 
+  # Temporary fix for #1231
+  if ! grep '^exec \${ASDF_DIR}' "$shim_path" >/dev/null; then
+    sed -i.bak -e 's/^exec .*\/bin\/asdf /exec ${ASDF_DIR}\/bin\/asdf /' "$shim_path"
+    rm -rf "$shim_path".bak
+  fi
+
   if [ -f "$shim_path" ]; then
     if ! grep -x "# asdf-plugin: ${plugin_name} ${version}" "$shim_path" >/dev/null; then
       sed -i.bak -e "s/exec /# asdf-plugin: ${plugin_name} ${version}\\"$'\n''exec /' "$shim_path"
@@ -97,7 +103,7 @@ write_shim_script() {
     cat <<EOF >"$shim_path"
 #!/usr/bin/env bash
 # asdf-plugin: ${plugin_name} ${version}
-exec $(asdf_dir)/bin/asdf exec "${executable_name}" "\$@" # asdf_allow: ' asdf '
+exec \${ASDF_DIR}/bin/asdf exec "${executable_name}" "\$@" # asdf_allow: ' asdf '
 EOF
   fi
 


jrogov:
FYI: ran it on my machine, works as intended:
```
~/c/asdf (master) $ cat ~/.asdf/shims/erl
#!/usr/bin/env bash
# asdf-plugin: erlang 24.2.1
exec /Users/jrogov/code/asdf/bin/asdf exec "erl" "$@" # asdf_allow: ' asdf '

~/c/asdf (master) $ git checkout fix/static-asdf-paths-for-shims
Switched to branch 'fix/static-asdf-paths-for-shims'

~/c/asdf (fix/static-asdf-paths-for-shims) $ ASDF_DIR=$(pwd) ./bin/asdf reshim

~/c/asdf (fix/static-asdf-paths-for-shims) $ cat ~/.asdf/shims/erl
#!/usr/bin/env bash
# asdf-plugin: erlang 24.2.1
exec ${ASDF_DIR}/bin/asdf exec "erl" "$@" # asdf_allow: ' asdf '

~/c/asdf (fix/static-asdf-paths-for-shims) $ erl
Erlang/OTP 24 [erts-12.2.1] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1]

Eshell V12.2.1  (abort with ^G)
```

lib/commands/reshim.bash
```diff
@@ -88,6 +88,12 @@ write_shim_script() {
   local shim_path
   shim_path="$(asdf_data_dir)/shims/$executable_name"
 
+  # Temporary fix for #1231
+  if ! grep '^exec \${ASDF_DIR}' "$shim_path" >/dev/null; then
+    sed -i.bak -e 's/^exec .*\/bin\/asdf /exec ${ASDF_DIR}\/bin\/asdf /' "$shim_path"
+    rm -rf "$shim_path".bak
+  fi
+
```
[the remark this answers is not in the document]
jrogov: That is absolutely correct. 

Although this covers all package managers that use version in paths to package installations.

jthegedus:
Do we actually want to disable `SC2016`?

doubledup:
> Do we actually want to disable `SC2016`?

It looks like it's an [exception](https://github.com/koalaman/shellcheck/wiki/SC2016#exceptions) to the rule, because we want the text `${ASDF_DIR}` inside the shim. ie. We don't want `ASDF_DIR` expanded when reshimming as it is currently, instead we want `${ASDF_DIR}` inside the shim itself so that it's expanded whenever the shim is executed. @jrogov Please correct me if I'm wrong here.

We might be able to keep `SC2016` on this line by switching to double quotes and adding more escaping, but that's probably less readable.

doubledup:
Is there some case where we need to expand `ASDF_DIR` when reshimming because `ASDF_DIR` isn't set when running a shim? I'm guessing not, given that the scripts that add asdf & the shims to `PATH` ensure that `ASDF_DIR` is set.

jthegedus:
> Is there some case where we need to expand `ASDF_DIR` when reshimming because `ASDF_DIR` isn't set when running a shim? I'm guessing not, given that the scripts that add asdf & the shims to `PATH` ensure that `ASDF_DIR` is set.

Perhaps, can you (@doubledup) see #531 and see what this change's impact might be?

lib/commands/reshim.bash
```diff
@@ -88,6 +88,14 @@ write_shim_script() {
   local shim_path
   shim_path="$(asdf_data_dir)/shims/$executable_name"
 
+  # Temporary fix for #1231
```
jthegedus, line 91: Does anyone have an idea how long this should live within the repo?

lib/commands/reshim.bash
```diff
@@ -97,7 +105,7 @@ write_shim_script() {
     cat <<EOF >"$shim_path"
 #!/usr/bin/env bash
 # asdf-plugin: ${plugin_name} ${version}
-exec $(asdf_dir)/bin/asdf exec "${executable_name}" "\$@" # asdf_allow: ' asdf '
```
jthegedus, line 100: The function `asdf_dir` (https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash#L22) seemingly checks for the value of `ASDF_DIR`, sets it if it doesn't exist, and then returns that same value. Are we hiding a bug in that functions behaviour by replacing it? Does anyone have time to investigate?

jthegedus: Awaiting feedback

Stratus3D: > The function `asdf_dir` (https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash#L22) seemingly checks for the value of `ASDF_DIR`, sets it if it doesn't exist, and then returns that same value. Are we hiding a bug in that functions behaviour by replacing it? Does anyone have time to investigate?

Good question. I believe the intent here is to replace `$(asdf_dir)`, which expanded to a path at **shim generation time** with syntax for the `$ASDF_DIR` variable, which is expanded at **shim execution time**, and therefore dynamically changes as `$ASDF_DIR` is changed, making shims more adaptable.

I need to spend some time thinking about this. It might be the right solution, or there may be a better solution. In either case I think we need some better testing around this.

Stratus3D: I wonder if instead of

``` bash
exec $(asdf_dir)/bin/asdf exec "${executable_name}" "\$@" # asdf_allow: ' asdf '
```

if we couldn't just do:

``` bash
exec asdf exec "${executable_name}" "\$@" # asdf_allow: ' asdf '
```

Should `asdf` already be on the `$PATH`? If so, why do we need to specify the absolute path of it? I need to investigate this.

lib/commands/reshim.bash
```diff
@@ -88,6 +88,14 @@ write_shim_script() {
   local shim_path
   shim_path="$(asdf_data_dir)/shims/$executable_name"
 
+  # Temporary fix for #1231
+  # shellcheck disable=SC2016
+  if ! grep '^exec \${ASDF_DIR}' "$shim_path" >/dev/null; then
+    # shellcheck disable=SC2016
+    sed -i.bak -e 's/^exec .*\/bin\/asdf /exec ${ASDF_DIR}\/bin\/asdf /' "$shim_path"
+    rm -rf "$shim_path".bak
+  fi
```
Stratus3D, line 97: If we do end up merging this I suggest we remove this `if` block and instead make sure there is an easy way for developers to delete all their shims and re-generate them from scratch.

Stratus3D:
@jthegedus do we still want this change? If so I can investigate the issue above and update the PR based on what I find.

jthegedus:
> @jthegedus do we still want this change? If so I can investigate the issue above and update the PR based on what I find.

I don't know if this is still necessary. I am not too familiar with this area of the code, how our other recent changes might affect this etc. If this is indeed a "fix", then it would be good to capture the scenario as a test case, so we don't cause regressions.

jrogov:
Hey guys, sorry about completely falling of the radar, it's been busy 😄 

I see that the issue was finally resolved in #1236, good job **jthegedus**!

Hence, I'm closing this one :slightly_smiling_face:

PR #1239, opened 2022-06-01
fix: updating references to legacy github.io site

robbyrussell:
Hi there!

While clicking on a few links to test for including in our [Ruby on Rails community survey results](https://rails-hosting.com/2022/)and noticed that the links to the previous URLs like https://asdf-vm.github.io/asdf/plugins/create.html were resulting in a 301 Redirect to http://asdf-vm.com/plugins/create.html, which would show a "Not Secure" warning in the browser.

# Summary

I went ahead and updated all these references to just go directly to https://asdf-vm.com/... 

Fixes: No existing issues found for this

## Other Information

You might want to look at your legacy github.io site redirect logic, too, so that external links get sent to the SSL site, too.


robbyrussell:
Going to redo this as I didn't pay attention to the linter rules!

PR #1241, opened 2022-06-06
fix: update plugin-add regex to support other languages

### LINKED ISSUE #1237: bug: Unable to install awscli plugin with recent Swedish locale
### Describe the Bug

I run Kubuntu 20.04 in Swedish and I'm unable to install the awscli plugin:
```
$ asdf plugin-add awscli
awscli is invalid. Name must match regex ^[a-zA-Z0-9_-]+$
```
After some digging I found that there may be a problem with the swedish locale not including the letter "w" properly. (We swedes have always had a complicated relationship with "w", e.g. in collation it should according to some be sorted under "v".) After reading more about it, it struck me that expressions like `a-z` very much depends on the language implied. If languages have additional characters within the a-z range or if alphabets are ordered differently, expressions like a-z do not make sense.

One way to solve this is to specify the language every time such expressions are used. E.g. in `plugins.bash`, grep could be preceded with `LANG=C`.

Relevant posts for more background:
* [Stack Overflow: Regular expression [A-Za-z] seems to not include letter W and w](https://stackoverflow.com/questions/52570103/regular-expression-a-za-z-seems-to-not-include-letter-w-and-w)
* [Glibc bug report: Handle [a-z] and [A-Z] in consistent portable fashion regardless of locale](https://sourceware.org/bugzilla/show_bug.cgi?id=23393) 

### Steps to Reproduce

LANG=sv_SE.UTF-8 asdf plugin-add awscli

### Expected Behaviour

Plugin is being added or the message `Plugin named awscli already added` shows if it's already installed.

### Actual Behaviour

asdf fails with return code 1 and the message 
```
awscli is invalid. Name must match regex ^[a-zA-Z0-9_-]+$
```

### Environment

```shell
OS:
Linux lap5cg1111r2p 5.13.0-44-generic #49~20.04.1-Ubuntu SMP Wed May 18 18:44:28 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
zsh 5.8 (x86_64-ubuntu-linux-gnu)

ASDF VERSION:
v0.10.1-f8109e0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIRENV_BIN=/home/johannem/.asdf/installs/direnv/2.31.0/bin/direnv
ASDF_DIR=/home/johannem/.asdf

ASDF INSTALLED PLUGINS:
ansible-base                 https://github.com/amrox/asdf-pyapp.git master 45a4efd
awscli                       https://github.com/MetricMike/asdf-awscli.git main ec4ccdc
direnv                       https://github.com/asdf-community/asdf-direnv.git master 7dcd571
golang                       https://github.com/kennyp/asdf-golang.git master cc8bc47
golangci-lint                https://github.com/hypnoglow/asdf-golangci-lint.git master 5b537fa
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master f9957f3
yadm                         https://github.com/particledecay/asdf-yadm.git master 7504791
yq                           https://github.com/sudermanjr/asdf-yq.git master 772992f

Also:
glibc (libc-bin, locales etc): 2.31-0ubuntu9.9
grep: 3.4-1
```

_No response_

Stratus3D:
[:alpha:] and [:digit:] character classes support characters from other
languages whereas ranges like a-z and 0-9 may not.

Fixes #1237

### DIFF
--- a/lib/functions/plugins.bash
+++ b/lib/functions/plugins.bash
@@ -56,8 +56,9 @@ plugin_add_command() {
 
   local plugin_name=$1
 
-  if ! printf "%s" "$plugin_name" | grep -q -E "^[a-zA-Z0-9_-]+$"; then
-    display_error "$plugin_name is invalid. Name must match regex ^[a-zA-Z0-9_-]+$"
+  local regex="^[[:alpha:][:digit:]_-]+$"
+  if ! printf "%s" "$plugin_name" | grep -q -E "$regex"; then
+    display_error "$plugin_name is invalid. Name must match regex $regex"
     exit 1
   fi
 
--- a/test/plugin_add_command.bats
+++ b/test/plugin_add_command.bats
@@ -20,16 +20,32 @@ teardown() {
   [ "$output" = "plugin_with-all-valid-CHARS-123" ]
 }
 
+@test "plugin_add command with LANG=sv_SE.UTF-8 and plugin name matching all valid regex chars succeeds" {
+  ORIGINAL_LANG="$LANG"
+  LANG=sv_SE.UTF-8
+
+  install_mock_plugin_repo "plugin-with-w"
+
+  # https://stackoverflow.com/questions/52570103/regular-expression-a-za-z-seems-to-not-include-letter-w-and-wA
+  # https://github.com/asdf-vm/asdf/issues/1237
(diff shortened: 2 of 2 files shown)


jthegedus:
Ah, nice improvement! Very interesting to know

Issue #866, opened 2021-02-10
`asdf current` should set spacing dynamically


MetricMike:
**Is your feature request related to a problem? Please describe.**

Some projects (hi, java!) use names that cause some overflow on the `asdf current` command.

```bash
metricmike@MTOWER:~/projects/dev-environment$ asdf current
awscli          ______          No version set. Run "asdf <global|shell|local> awscli <version>"
dotnet-core     5.0.102         /home/metricmike/.tool-versions
golang          1.15.8          /home/metricmike/.tool-versions
gradle          7.0-milestone-1 /home/metricmike/.tool-versions
java            corretto-11.0.10.9.1 /home/metricmike/.tool-versions
maven           3.6.3           /home/metricmike/.tool-versions
```

**Describe the solution you'd like**

`asdf current` should set its columns dynamically based on the plugins it's going to output

```bash
metricmike@MTOWER:~/projects/dev-environment$ asdf current
awscli          ______                    No version set. Run "asdf <global|shell|local> awscli <version>"
dotnet-core     5.0.102                   /home/metricmike/.tool-versions
golang          1.15.8                    /home/metricmike/.tool-versions
gradle          7.0-milestone-1           /home/metricmike/.tool-versions
java            corretto-11.0.10.9.1      /home/metricmike/.tool-versions
maven           3.6.3                     /home/metricmike/.tool-versions
```

**Additional context**

Happy to work on this myself, this issue is mostly a placeholder for me :)

jthegedus:
Hey @MetricMike thanks for proposing this and volunteering to work on it. There's a lot of background to improving the terminal output. I will link here and encourage you to read before working on this.

`column` was previously added and removed as we cannot assume it is available on all machines:
- initial add: https://github.com/asdf-vm/asdf/pull/604
- reported error from user: https://github.com/asdf-vm/asdf/issues/661
- removal of `column`: https://github.com/asdf-vm/asdf/pull/721
- ban of command: https://github.com/asdf-vm/asdf/pull/754

Wrap terminal line outputs to 80 chars (https://github.com/asdf-vm/asdf/issues/646). This issue discusses:
- 80 line width being our target for commands that output to the terminal. As you have discovered this sometimes isn't the best result
- a `--porcelain` flag for scripts to consume (removing their reliance on the current output allowing use to develop that with a user focus)
- a dynamic method to detect the current terminal width and render based on that
[...]

PR #1253, opened 2022-06-19
feat: Add dynamic column widths to `current`

MetricMike:
# Summary

`asdf current` collects output, determines appropriate column widths given the current terminal, and truncates an individual column if necessary. Practically, this will usually truncate the path/3rd column, but
This behavior improves scannability and utilization of large terminals. It also gets close to restoring pre-0.7.6 formatting, but without relying on the now-banned `column`.

I say close, because `current/which` never attempted dynamic columns. But if this approach is acceptable, I can use a similar approach for the `asdf plugin` commands (specifically `list`).

Fixes: #866

## Other Information

From the history that @jthegedus compiled on #866:
- [x] this does not use `column` or other banned commands. It does add `stty`, but that's part of coreutils and *should* be safe
- [ ] no `--porcelain` flag for scripts to consume
- [ ] no flag for users to force full output, other than expanding their terminal size
- []


MetricMike:
Rebased to match master/v0.11.0. Could someone approve the other workflows to ensure I haven't broken any tests? "Works on my machine", but would appreciate the reassurance from github actions

jthegedus:
I have only given this a quick look, but I like the direction, compute term > gen output > truncate if needed & configured > output 👍 This is great!

It looks like we might be able to make this solution generic to handle X columns, where X < 3 or 4. I am not actually sure what our max column count is, I think it is 4 with `asdf plugin list --urls --refs`. It would be good to have a single function for output. What are your thoughts?

---

This is just a quick brain dump of related thinking, please do not include in this PR:
- I like `--tsv` moreso than `--porcelain` as it implies the format directly, we will see what Stratus says in #1291 and we will go with whatever they preference.
- I want to add global flags, help, porcelain and verbose should probably all live there. We can re-factor after the fact.

MetricMike:
Thanks for the feedback! I really appreciate how much detail y'all have been putting into other issues - while it adds a bit of time for reading, it pays off by providing a good "feel" for the asdf's direction.

oof, yeah definitely missed something with all those test failures! Will try to track those down this week.

I agree on generalizing this for other commands / 4+ columns. It feels a little hacky to be intercepting plugin_list_command output, and the broader goal of supporting porcelain/verbose flags would be much easier if asdf had an internal printer/logger that dealt with formatting and spacing concerns (e.g., `asdf current` should care only about retrieving the currently activated versions, not how to display them). The original issue was scoped to just `current`, so I'd like to get this "mvp" handled before working on generalizing this approach for other commands.

jthegedus:
>I'd like to get this "mvp" handled before working on generalizing this approach for other commands.
I concur

Stratus3D:
Already commented on #866 but I'll comment here as well. I'm focusing on the Go rewrite #1805 right now. Once that is done I'm going to close this PR.

Stratus3D:
Closing as #1805 has been merged.

PR #1272, opened 2022-06-23
chore: bump vuepress to latest beta

jthegedus:
# Summary

This updates Vuepress to the latest Beta release `2.0.0.-beta.48`  factoring in breaking changes from the Vuepress changelog that the Dependabot PRs are unaware exist.

Tested locally to ensure the local search, i18n and navbar/sidebar still work as expected.

## Other Information

- Vuepress now defaults to using Vite (previously the default was Webpack) so we no longer needed to specify some deps and compat libs. This can be seen in the changes in `package.json`. Notably `package-lock.json` is much smaller so we should have fewer Dependabot updates because we use fewer dependencies now.
- `npm run format` ran the prettier over the JS code as well, hence the large diff as it is just appending trailing commas (`,`).
- I also confirmed the build pipeline remains the same, so deployment did not have any breaking changes


jthegedus:
Merging this in to quite Dependabot PRs

PR #1275, opened 2022-06-25
fix: add missing "does not add paths to PATH more than once" test for elvish

elijahr:
# Summary

This test exists for `fish` and for `sh`, may as well add it for `elvish`.


Stratus3D (approved):

Stratus3D:
Thanks for the PR @elijahr ! This is great.

PR #1277, opened 2022-06-25
fix: gitignore macOS .DS_Store files

elijahr:
# Summary

macOS Finder creates these hidden files. It's common to add this entry to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ repository
 .vagrant
 keyrings
 /tmp
+.DS_Store


Stratus3D:
In my opinion only project-specific files should go in a project's `.gitignore`. Global gitignore should be used for files specific to your OS, tools, or workflows. 

See http://stratus3d.com/blog/2018/06/03/stop-excluding-editor-temp-files-in-gitignore/

jthegedus:
Shall we close this PR then @Stratus3D ?

PR #1293, opened 2022-07-12
ci: update 

jthegedus:
# Summary

- use bash strict mode in `scripts/` files
- fix `sed` for replacing version in `docs/guide/getting-started` page. Move to standalone script for easier dev testing
- add script for updating version in sec policy. This script will only replace if a new `major.minor` version as that is the current format for the sec pol list. Exec this script in the same CI pipeline as documentation version update so there is only a single auto-commit and avoid potential git race conditions of multiple updates from different jobs or workflows.

## Other Information

I believe this will fix the issues with auto-docs updating defined in #1000 & #1232


Stratus3D (approved):

scripts/update_version_list_in_security_policy.bash
```diff
@@ -0,0 +1,12 @@
...
+
+set -euo pipefail
+
+version_major_minor_x="$(cut -f1-2 -d "." version.txt).x"
+
+# skip if version is already in the list
+if ! grep -q "$version_major_minor_x" SECURITY.md; then
+  # prepend new version to the list
+  sed -i "s/white_check_mark:/x:               /g" SECURITY.md
+  sed -i "s/^\\(| -* | -* |\\)$/\\1\\n| $version_major_minor_x  | :white_check_mark: |/" SECURITY.md
```
jthegedus, line 11: The whitespace for the newline is not perfect, but not too messy and the markdown rendering of GitHub resolves the table correctly regardless of whitespace.

Issue #1248, opened 2022-06-14
Custom default version 


chulkilee:
### Is your feature request related to a problem? Please describe

I put all asdf config to XDG directories, and used to use `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` to set the default version.

Apparently using `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` for that purpose is incorrect (#912) - and now it's fixed (#1238) - and now I have to use `$HOME/.tool-versions` the default, which is only asdf file outside the XDG.

### Describe the proposed solution

It would be nice to have env var to specify the location of the global default file - e.g. `ASDF_GLOBAL_TOOL_VERSIONS_PATH` or something like.

### Describe similar `asdf` features and why they are not sufficient

None

### Describe other workarounds you've considered

For now I'm using symlink - but it still keep the `$HOME/.tool-versions` file - which I'd like to remove.

bigodel:
as a workaround, i've been setting `ADSF_DEFAULT_TOOL_VERSIONS_FILENAME` to `.config/asdf/tool-versions`, and it seems to be working fine albeit the ugly hack.

chulkilee:
@bigodel that was used to work but now not working at `v0.10.2-7e7a1fa` with the fix. Now it ignores the local `.tool-versions` since it uses the variable for the local tool version

jthegedus:
This thread is essentially about XDG support - https://github.com/asdf-vm/asdf/issues/687

Comment https://github.com/asdf-vm/asdf/issues/687#issuecomment-1005195311 from January 2022 requested a ticket rename to better describe the intent, sorry for not getting to that sooner.

I am going to mark this as a duplicate of #687 If you feel this is incorrect please reopen this Issue.

Issue #687, opened 2020-03-29
Common location for default packages for all plugins


jthegedus:
### Idea

Currently `asdf-nodejs`, `asdf-ruby` and `asdf-python` among other plugins allow for `$HOME/.default-<tool>-packages` config files. It is clear from the discussions in each plugin that people want to be able to configure the file location.

### Proposals

Perhaps `asdf` core can provide plugins with `ASDF_CONFIG` dir that is the directory of `.asdfrc` by default?

This way, users can configure this, and plugin authors can focus on the default packages filename appropriate for their plugin and install those packages instead of also having to support a diverging set of dirs. 

Or perhaps this is too much and simply doing:

```shell
default_gem_path() {
  local config_path=${ASDF_CONFIG_FILE:-"$HOME/.asdfrc"}
  local config_dir=$(dirname "$config_path") 
```

as is proposed in `asdf-ruby` https://github.com/asdf-vm/asdf-ruby/pull/114/files is enough.

### Related

- https://github.com/asdf-vm/asdf-ruby/issues/111
- https://github.com/asdf-vm/asdf-nodejs/issues/152
- https://github.com/asdf-vm/asdf-ruby/pull/114

andrewthauer:
The main reason I would want to see this is to be more [XDG](https://wiki.archlinux.org/index.php/XDG_Base_Directory) compliant. For instance, I currently do the following in order to keep my `$HOME` directory cleaner.

```
export ASDF_CONFIG_FILE="${XDG_CONFIG_HOME}/asdf/asdfrc"
```

Ideally the `${XDG_CONFIG_HOME}` would be the default location for all `asdf` related config files vs `$HOME`. Currently it's possible to override this for some parts of `asdf` via environment variable, but not for `deafult-*` files.

Stratus3D:
I know some existing version managers (like rbenv for Ruby) currently look for `.default-gems` in the users home directory. I don't know if rbenv allows the user to customize the location of the file, but I would guess most rbenv users don't, so ideally asdf-ruby would look for `.default-gems` in the users home directory by default too. I'm open to a new feature that allows users to better organize things with asdf, but I also think we should imitate the behavior of other widely used legacy version managers whenever possible.

andrewthauer:
To be clear, I brought up `XDG` as a very valid example of why people may want to override the location of these files. Currently the `default-*` files are the only standard parts I'm unable to do this with.

[...]

PR #1295, opened 2022-07-12
feat: Allow for custom global tool-versions location

brianvanburken:
# Summary

Since a recent bug fix to the behaviour around the env variable `ADSF_DEFAULT_TOOL_VERSIONS_FILENAME`, which I also abused to follow the XDG spec, I was unable to move my global `.tool-versions`. As was also reported by other people in #1248.

This PR allows the user to give a custom location via `ASDF_CONFIG_DIR` if set (for those who want it to move the file and don't follow the XDG spec), `XDG_CONFIG_HOME`, or falls back to `HOME`.

I'm open to suggestions if another approach is better.

Related issues: #687

### DIFF
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -20,7 +20,7 @@ version_command() {
   file_name="$(version_file_name)"
 
   if [ "$cmd" = "global" ]; then
-    file="$HOME/$file_name"
+    file="$(asdf_config_dir)/$file_name"
   elif [ "$cmd" = "local-tree" ]; then
     file=$(find_tool_versions)
   else # cmd = local
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -50,6 +50,18 @@ asdf_data_dir() {
   printf "%s\\n" "$data_dir"
 }
 
+asdf_config_dir() {
+  local config_dir="${HOME}"
+
+  if [ -n "${ASDF_CONFIG_DIR}" ]; then
+    config_dir="${ASDF_CONFIG_DIR}"
+  elif [ -n "${XDG_CONFIG_HOME}" ]; then
+    config_dir="${XDG_CONFIG_HOME}/asdf"
+  fi
+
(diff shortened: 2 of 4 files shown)


jthegedus:
I only took a quick look, but I think there are some aspects missing from this. Importantly is the resolution of `.tool-versions` because I believe it is just a walk up the directory tree, where the location of `global` is not special and simply likely to be part of the users current working directory.

I also think we could utilise Bats test conventions to unset the XDG value in a single place instead of in each individual test.

jthegedus:
`ASDF_GLOBAL_TOOL_VERSIONS_PATH` would still require more work from this PR to actually work though, as current lookup of Global is at `$HOME/ADSF_DEFAULT_TOOL_VERSIONS_FILENAME`. We would need to change the lookup method.

brianvanburken:
@jthegedus Thank you for taking the time to look at my PR! I'll improve the testing using your feedback.

I'm not so familiar with the project so bear with me. I looked at the code as how it resolves global and local `.tool-versions`. As far as I can see, there are two branches; one branch checks directly the location of the global files. The other branch looks locally in the folder and walks upward as you described. My PR targets the global file and allows users to move this using different options. I think walking up the directory tree for the file is irrelevant, if I'm not mistaken. Could you tell me more specifically what aspect I'm missing here?

As for the `ASDF_GLOBAL_TOOL_VERSIONS_PATH`, it is a small change to make. But, I'm looking at a bit bigger picture. I saw that there is a `ASDF_DATA_DIR` variable where everything data related is stored for asdf. So in the same spirit, using `ASDF_CONFIG_DIR`, we could also implement a future change to store `.asdfrc` in that directory.

I'm currently finishing up some things at work before my vacation, so I'm a bit swamped. When I return, I'll finish the implementation in one week.

brianvanburken:
I've switched from `asdf` and won't know about the project and bash, along with time, to finish this. If anyone wants to pick up where I left off, feel free to take my code

Issue #1288, opened 2022-07-05
bug: ci tests fail when repository has no tags


jsoref:
### Describe the Bug

ci tests fail when repository has no tags

```
git tag
```
yields no output if a repository has no tags. This is something that should be expected and tolerated/handled.

### Steps to Reproduce

1. Fork this repository
2. Delete all tags
3. Push a commit to the repository that adds a blank line to the README
4. Create a PR `chore: test`

### Expected Behaviour

the tests should all run to completion and pass

### Actual Behaviour

```
not ok 206 asdf update should checkout the latest non-RC tag
# (in test file test/update_command.bats, line 41)
#   `git tag | grep $tag' failed with status 2
...
not ok 207 asdf update should checkout the latest tag when configured with use_release_candidates = yes
# (in test file test/update_command.bats, line 52)
#   `git tag | grep $tag' failed with status 2
```

### Environment

```shell
(I don't have asdf, the CI failed)
https://github.com/check-spelling/asdf/commit/624aede7f3480cd6a8817d5a73b72227e87fa668
```

PR #1296, opened 2022-07-13
ci: tolerate the lack of tags in the repository

jsoref:
# Summary

Wrap uses of `git tag` output with a check to see if there's a value...

Fixes: #1288 

## Other Information

Tested here: https://github.com/check-spelling/asdf/actions/runs/2660784666


jthegedus (approved):
This seems like an adequate solution given the minor cause, thanks!

Issue #1306, opened 2022-07-19
bug: "milestone" releases are not excluded when searching for "latest" version


log2:
### Describe the Bug

`asdf latest xyz` should not list, and thus select, `*-milestone-*` releases of `xyz` plugin.

However, `asdf latest gradle` shows `7.5-milestone-1` as latest, since it is the last line in `asdf list-all gradle`.

### Steps to Reproduce

1. Enter `asdf plugin add gradle`
2. Enter `asdf latest gradle`

### Expected Behaviour

`asdf latest` should not show a non-release version, such as `*-milestone-*`

### Actual Behaviour

Version `7.5-milestone-1` is shown, rather than actual latest, stable version `7.5`

### Environment

```shell
OS:
Darwin Mac-mini-di-Lorenzo-3.local 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:22 PDT 2022; root:xnu-8020.121.3~4/RELEASE_X86_64 x86_64

SHELL:
zsh 5.8.1 (x86_64-apple-darwin21.0)

ASDF VERSION:
v0.10.2

ASDF ENVIRONMENT VARIABLES:
ASDF_DIRENV_BIN=/Users/log2/.asdf/installs/direnv/2.32.1/bin/direnv
ASDF_DATA_DIR=/Users/log2/.asdf
ASDF_DIR=/Users/log2/.asdf-code

ASDF INSTALLED PLUGINS:
alias                        https://github.com/andrewthauer/asdf-alias.git master 94da3a7
awscli                       https://github.com/MetricMike/asdf-awscli.git main c645552
cosign                       https://gitlab.com/wt0f/asdf-cosign.git main ee306c4
direnv                       https://github.com/asdf-community/asdf-direnv.git master a5aefdf
dive                         https://github.com/looztra/asdf-dive.git master 1321207
eksctl                       https://github.com/elementalvoid/asdf-eksctl.git master 5309e4b
flux2                        https://github.com/tablexi/asdf-flux2.git master 771755e
gcloud                       https://github.com/jthegedus/asdf-gcloud.git main f59e332
gradle                       https://github.com/rfrancis/asdf-gradle.git master 34e1e36
grype                        https://github.com/poikilotherm/asdf-grype.git main d6e0ece
helm                         https://github.com/Antiarchitect/asdf-helm.git master a39e17b
java                         https://github.com/halcyon/asdf-java.git master 9490cdc
[...]

PR #1307, opened 2022-07-19
fix: excludes "milestone" releases in "latest" command

log2:
# Summary

Fixes: https://github.com/asdf-vm/asdf/issues/1306

### DIFF
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -146,7 +146,7 @@ latest_command() {
   else
     # pattern from xxenv-latest (https://github.com/momo-lab/xxenv-latest)
     versions=$(list_all_command "$plugin_name" "$query" |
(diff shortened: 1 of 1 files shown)


jthegedus:
This is a minor change that I don't think it will hurt anyone, however we want to encourage plugin authors to implement their own `bin/latest-stable` script which you can see is called 9 lines above this change. Docs: https://asdf-vm.com/plugins/create.html#bin-latest-stable

log2:
I've added [PR](https://github.com/rfrancis/asdf-gradle/pull/11) to asdf-gradle project, as promised, @jthegedus

Stratus3D (approved):
I'm fine with the changes here but the existence of this code bothers me the in the same way as the PR I was reviewing earlier - https://github.com/asdf-vm/asdf/pull/1191#discussion_r929074452

Stratus3D:
Thanks for the PR @log2 ! Sorry for the late review.

Issue #1286, opened 2022-07-04
bug: asdf reshim cann't work with local path


alexezio:
### Describe the Bug

version: v0.10.2

when you define a local path in .tool-versions and reshim this. you will get a sed issue because of sed delimiter character(/) is in conflict with the Unix path(/root/xxx/xxx):
sed: 1: "s/exec /# asdf-plugin:  ...": bad flag in substitute command: 'U'

### Steps to Reproduce

1. define the local path in .tool-versions file
2. asdf reshim <name> <path>
3. get a error

This is my .tool-versions file:
terraform path:/Users/benchilian/.asdf/installs/golang/1.16.4/packages

### Expected Behaviour

reshim should work as expected, and ~/.asdf/shims/terraform should contains a local path plugin line

### Actual Behaviour
here is my bash test output
+ sed -i.bak -e 's/exec /# asdf-plugin: terraform path:/Users/benchilian/.asdf/installs/golang/1.16.4/packages\
exec /' /Users/benchilian/.asdf/shims/terraform
sed: 1: "s/exec /# asdf-plugin:  ...": bad flag in substitute command: 'U'

### Environment

```shell
OS:
Darwin Benchis-MBP 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.8.1 (x86_64-apple-darwin21.0)

ASDF VERSION:
v0.10.2

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/opt/homebrew/opt/asdf/libexec
ASDF_DIRENV_BIN=/Users/benchilian/.asdf/installs/direnv/2.32.1/bin/direnv

ASDF INSTALLED PLUGINS:
ansible                      https://github.com/amrox/asdf-pyapp.git master f616c6f
direnv                       https://github.com/asdf-community/asdf-direnv.git master a5aefdf
golang                       https://github.com/kennyp/asdf-golang.git master cc8bc47
terraform                    https://github.com/asdf-community/asdf-hashicorp.git master bc4dec6
```

### asdf plugins affected (if relevant)

terraform and all plugins

Issue #1115, opened 2021-12-06
bug: reshim does not update asdf path


axelson:
### Describe the Bug

```
jason@Jasons-Work-MacBook-Air ~/f (jax/webpack-quiet)> echo $ASDF_DIR
/opt/homebrew/Cellar/asdf/0.9.0/libexec
jason@Jasons-Work-MacBook-Air ~/f (jax/webpack-quiet)> cat ~/.asdf/shims/dotenv
#!/usr/bin/env bash
# asdf-plugin: ruby 3.0.2
exec /opt/homebrew/Cellar/asdf/0.8.1_1/libexec/bin/asdf exec "dotenv" "$@"
jason@Jasons-Work-MacBook-Air ~/f (jax/webpack-quiet)> asdf reshim ruby
jason@Jasons-Work-MacBook-Air ~/f (jax/webpack-quiet)> cat ~/.asdf/shims/dotenv
#!/usr/bin/env bash
# asdf-plugin: ruby 3.0.2
exec /opt/homebrew/Cellar/asdf/0.8.1_1/libexec/bin/asdf exec "dotenv" "$@"
```

Might be related to #1103

### Steps to Reproduce

Before upgrading to 0.9.0 via homebrew:
1. Install ruby via asdf
2. gem install dotenv
3. Update asdf to 0.9.0 via homebrew
4. Run `asdf reshim ruby`
5. Path in `~/.asdf/shims/dotenv` is still incorrect

Note: Untested

### Expected Behaviour

In step 5 the dotenv shim should have the 0.9.0 path

### Actual Behaviour

The path in the dotenv shim has the old 0.8.1 path

### Environment

```shell
OS:
Darwin Jasons-Work-MacBook-Air.local 20.6.0 Darwin Kernel Version 20.6.0: Wed Jun 23 00:26:27 PDT 2021; root:xnu-7195.141.2~5/RELEASE_ARM64_T8101 arm64

SHELL:
fish, version 3.3.1

ASDF VERSION:
v0.9.0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/opt/homebrew/Cellar/asdf/0.9.0/libexec

ASDF INSTALLED PLUGINS:
elixir                       https://github.com/asdf-vm/asdf-elixir.git master 63e9cd8
erlang                       https://github.com/asdf-vm/asdf-erlang.git master 9ca2dea
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master 95a2091
[...]

PR #1311, opened 2022-07-25
fix: `reshim` did not rewrite executable path

dylan-chong:
# Summary

Fixes reshimming not working by *always* recreating the file from scratch.
We grab the existing plugin versions from the existing shim file and squish those into a brand new shim file.
This solves the problem of updating asdf from homebrew causing the executable path to be wrong, and not fixable by `asdf reshim`

Regenerating the shim file from scratch is much simpler, no additional tests required like https://github.com/asdf-vm/asdf/pull/1287.

Fixes: https://github.com/asdf-vm/asdf/issues/1115
Fixes: #1231 (dupe)
Fixes: #1286 
Makes change in #1287 not needed, although the test is still useful

## TLDR for comment section

Read from this comment down https://github.com/asdf-vm/asdf/pull/1311#issuecomment-1200301398


anildigital (approved):

alexezio:
hello, @dylan-chong :
  What if the execution have more than one shim path?  
   if we re-generate the reshim file,  the shim file will only contains one shim path other than all of their.  This will cause issue for origin shim path.  could you please run the bats for verify the results?

dylan-chong:
Hey

@alexezio you're right it is breaking some tests.
Why was it implemented as using regex to accumulate data in a file? That gives me the heebie jeebies. Why not take a list of 
Also why does the shim have to list the versions of the plugin?

If we could simplify the API such that we remove all shim files and regenerate them all (so `asdf reshim` takes no arguments) then that would simplify things massively. There would be only one code path:

1. delete all existing shims
2. generate shims for all versions of each executable for each plugin

There's no need to keep files around and do complicated regexes on them if we can regenerate them from scratch every time (which is what you have to do when upgrading asdf with homebrew anyway). Unless, there's something I'm missing?

Happy to have a go at implementing this idea if you'd like ^.
Think it'd be a good improvement to make

alexezio:
hey dylan:
  In my opinion, asdf is stateless,  so the shim file have also saved the program version state.  If we regenrate them from scrach,  We have no where to fetch version state except the file itself.
  although regex is annoying,  It can modify the version state without break them(by increment).
  I am a newbee for asdf, Maybe there is a better approach to reshim.  but for now, I vote for regex...
  thanks~

dylan-chong:
@alexezio but you can delete the shim files manually and regenerate the files from scratch? Does this not put the version comments in the file?
Is it not possible to just list all versions of the plugin?

if you would prefer to keep the regex  I could have a go at replacing everything except the existing version list. This would fix the issue in the description.

dylan-chong:
Cc @Stratus3D @jthegedus which option would you prefer:

1. simplify the reshim API  and have the command re-generate all files from scratch [for any given plug-in]. Would need a way to be able to list all versions of the plugin supported by the executable
2. Replace sed with grep so in the existing function we grab out the existing versions and copy them over into a freshly generated file, which is written over the top

IMO 1 is better long term , but requires more work now.

dylan-chong:
The easiest option was to implement no 2 (see above), which I have committed to this branch.

@Stratus3D @jthegedus This is ready to merge

@alexezio This means that the change to the sed pattern in your PR #1287 is not longer required. But it would still be great to merge the test that you wrote in your PR. I have checked and that test passes in this branch, but I don't want to steal your contribution so I'll let you commit that :)

jthegedus:
I am not sure why the tests didn't run here, but merging `master` so they can.

jthegedus:
@Stratus3D looking at our open PRs we have 3 which attempt to resolve this issue. Are you able to check this as it seems to me to be the best option. I am concerned about performance, but here correctness is preferred, we can come back and fix perf.

lib/commands/reshim.bash
```diff
@@ -88,22 +88,31 @@ write_shim_script() {
   local shim_path
   shim_path="$(asdf_data_dir)/shims/$executable_name"
 
+  local temp_dir
+  temp_dir=${TMPDIR:-/tmp}
+
+  local temp_versions_path
+  temp_versions_path=$(mktemp "$temp_dir/asdf-command-reshim-write-shims.XXXXXX")
```
Stratus3D, line 95: I'm not sure line 91-92 are necessary as `mktemp` should fallback to `TMPDIR` and `/tmp` automatically.

lib/commands/reshim.bash
```diff
@@ -88,22 +88,31 @@ write_shim_script() {
...
 EOF
-  fi
+
+  rm "$temp_versions_path"
 
   chmod +x "$shim_path"
 }
 
+# TODO this function doesn't seem to be used
 generate_shim_for_executable() {
```
Stratus3D, line 116: You are correct. Please delete this function if you want.

Stratus3D (commented):
Other than my two comments above this PR looks good to me and I think we should get it merged.

I agree with @dylan-chong that approach #1 is better but this PR should work for now.

Stratus3D:
> @Stratus3D looking at our open PRs we have 3 which attempt to resolve this issue. Are you able to check this as it seems to me to be the best option. I am concerned about performance, but here correctness is preferred, we can come back and fix perf.

I completely agree. Let's try to get this shipped. I can re-assess performance later if necessary. Thanks @jthegedus !

lib/commands/reshim.bash
dylan-chong: @jthegedus, do you have any thoughts since you added this?

Stratus3D (approved):
Approving now as I don't want to hold this up. But I think we should address the `TMPDIR` issue if we can.

lib/commands/reshim.bash
jthegedus: This seemed fine to me, but after looking at existing use of `mktemp` this was the outlier, so I changed it to align with the other usages. We can back this out later once we clean up and decide to put this behind a function or something.

jthegedus:
Let's see how it goes 😀

PR #1317, opened 2022-08-15
refactor: use `fish_add_path` during fish setup if available

ypconstante:
# Summary

Instead of modifying the PATH variable directly, use `fish_add_path` utility function provided by fish.

## Other Information

`fish_add_path` documentation: https://fishshell.com/docs/current/cmds/fish_add_path.html

Use `--global` option to modify only current session.
Use `--move` option to move the paths to front of other items in the path instead of keeping them on the previous position if already in the path.

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -2,12 +2,38 @@ if not set -q ASDF_DIR
   set -x ASDF_DIR (dirname (status -f))
 end
 
-fish_add_path --global --move $ASDF_DIR/bin
+# Add asdf to PATH
+# fish_add_path was added in fish 3.2, so we need a fallback for older version
+if type -q fish_add_path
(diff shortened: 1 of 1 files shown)


danhper:
Hi @ypconstante, thanks for the PR.
I'm in general in favor of this but the issue is that we do want to keep things compatible with older fish versions.
`fish_add_path` has been added in fish 3.2, which is quite recent. For example, Ubuntu 20.04 has fish 3.1 by default.
Maybe we could check the `$version` variable and use `fish_add_path` only for fish versions where it is available?

ypconstante:
I've added a check to see if `fish_add_path` is available, and if it's not run the previous path manipulation code. I've also modified the new code to only make one call to `fish_add_path`.

PR #1319, opened 2022-08-21
chore: Add `--all` completion to `asdf latest`

### LINKED ISSUE #1206: chore: `--all` flag missing from completions for `asdf latest` command
### Describe the Bug

The completions for the `--all` flag to the `latest` command are missing.

### Steps to Reproduce

Type `asdf latest ` and press <TAB>.

### Expected Behaviour

Show `--all` flag besides installed plugins.

### Actual Behaviour

Only shows plugins.

### Environment

```shell
SHELL:
zsh 5.8 (x86_64-apple-darwin21.0)

ASDF VERSION:
v0.10.0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/Users/acidghost/.dotfiles/asdf-vm

ASDF INSTALLED PLUGINS:
golang                       https://github.com/kennyp/asdf-golang.git master cc8bc47
julia                        https://github.com/rkyleg/asdf-julia.git master 1788f0f
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master f9957f3
perl                         https://github.com/ouest/asdf-perl.git master 1d6d39c
python                       https://github.com/danhper/asdf-python.git master 57a4d72
ruby                         https://github.com/asdf-vm/asdf-ruby.git master 3179071
```

_No response_

hyperupcall:
Closes #1206

### DIFF
--- a/completions/_asdf
+++ b/completions/_asdf
@@ -170,7 +170,9 @@ case "$subcmd" in
   ;;
 (latest)
   if (( CURRENT == 3 + IntermediateCount )); then
-    _asdf__installed_plugins
+    _alternative  \
+      'all:all:(--all)' \
+      'asdf-available-plugins:Installed ASDF Plugins:_asdf__installed_plugins'
   elif (( CURRENT == 4 + IntermediateCount )); then
     local pkg="${words[3+IntermediateCount]}"
     local query=${words[4+IntermediateCount]}
--- a/completions/asdf.bash
+++ b/completions/asdf.bash
@@ -67,6 +67,10 @@ _asdf() {
       COMPREPLY=($(compgen -W "$plugins" -- "$cur"))
     fi
     ;;
+  latest)
+    # shellcheck disable=SC2207
+    COMPREPLY=($(compgen -W "--all" -- "$cur"))
+    ;;
   *)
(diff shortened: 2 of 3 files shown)


hyperupcall:
Sorry about the force push mess, it's ready now

completions/_asdf
```diff
@@ -170,7 +170,9 @@ case "$subcmd" in
   ;;
 (latest)
   if (( CURRENT == 3 + IntermediateCount )); then
-    _asdf__installed_plugins
+    _alternative  \
```
Stratus3D, line 173: For `plugin-update` above I see:

```
(( CURRENT == 3 + IntermediateCount )) && _alternative \
    'all:all:(--all)' \
    ...
```

Is there a significance between the two different syntaxes?

Stratus3D (approved):
Looks good to me but I am not an expert on these completion files.

completions/_asdf
hyperupcall: Not sure if I completely understand your question, but I did the `if (( )); then ...` thing instead of `(( )) && ...` because there was already an if statement like `install` and it wasn't like `plugin-update`. In the context of completion, I _think_ there is no difference, since `set -e` is not set - I wish I remembered why I avoid it in regular scripts

PR #1328, opened 2022-09-02
test: use resulting PATH on duplicated source tests

ypconstante:
# Summary

On the tests verifying if the path items are not added multiple times to PATH when asdf is sourced twice, instead of checking the resulting PATH, the tests were checking the current terminal PATH


Stratus3D (approved):
Good catch!

PR #1331, opened 2022-09-15
fix: Use ${BASH_SOURCE[0]} instead of $0 to link scripts

quilicicf:
# Summary

While trying to run `asdf` in a Jenkins kubernetes agent, we noticed the following error:

```
/home/jenkins/.asdf/bin/asdf: line 5: ./lib/utils.bash: No such file or directory
/home/jenkins/.asdf/bin/asdf: line 28: asdf_dir: command not found
/home/jenkins/.asdf/bin/asdf: line 81: asdf_dir: command not found
Unknown command: `asdf current`
```

We noticed that this stems from using `dirname "$0"` to find the path to the current script.

Depending on the context in which the script is run, `$0` seems to return different values.

Changing all occurrences by `${BASH_SOURCE[0]}` or directly using `${ASDF_DIR}` seems to fix this.

You can reproduce this by:

* creating the Dockerfile below
    ```Dockerfile
    FROM ubuntu

    RUN apt update \
      && apt install git -y \
      && git clone https://github.com/asdf-vm/asdf.git ~/.asdf

    ENV BASH_ENV=/root/.asdf/asdf.sh
    ```
* building the Docker image like with:
    ```bash
    docker buildx build --tag asdf .
    ```
* run the following command:
    ```bash
    docker run -it --rm "$IMAGE" bash asdf current
    ```

## Test the fix

I tested the fix by following the same instructions with the Docker image created from my code by putting the following Dockerfile in the repository:

```Dockerfile
FROM ubuntu

COPY . /root/.asdf

RUN apt update && apt install git -y

ENV BASH_ENV=/root/.asdf/asdf.sh
```

This was not a full test though.

Linked to #1334

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # shellcheck source=lib/utils.bash
-. "$(dirname "$(dirname "$0")")/lib/utils.bash"
+. "$(dirname "$(dirname "${BASH_SOURCE[0]}")")/lib/utils.bash"
 
 find_cmd() {
   local cmd_dir="$1"
--- a/bin/private/asdf-exec
+++ b/bin/private/asdf-exec
@@ -3,7 +3,7 @@
 # remove this asdf-exec file when releasing >=0.6.5
 printf "asdf is self upgrading shims to new asdf exec ...\\n"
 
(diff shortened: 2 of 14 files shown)


quilicicf:
The issue was in the Docker image, not in `asdf`. Fixed it.

PR #1349, opened 2022-10-17
refactor: Improve Bash Syntax

### LINKED ISSUE #1313: Improve consistency with bash syntax

While reading the code, I've noticed a bit of inconsistency with how bash syntax is used. The one I've most noticed is the usage of the conditional braces (``[[`` vs `[`), often within the same file, e.g. [this](https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash).

It's clear that bash syntax is intended, and that POSIX compatibility isn't desired, which is totally fine. However, if bash is being embraced, it should be embraced fully.

I would like to see consistency in the syntax usage. This is not just for readability. There are real noticeable affects. For example, the use of `[` is actually a completely separate program. Try `which [` in a bash shell. If you use `[` instead of `[[`, it's actually creating a new process to do the comparison. If `[[` is used, the comparison happens completely in the bash process. Aside from that, the quoting rules are a bit more forgiving with `[[`.

I'm happy to take on this project, but I would like to know how this PR will be received, first.

N/A

N/A

hyperupcall:
This fixes various subtle Bash issues / inconsistencies. Generally, these improve semantics and make things a bit clearer - extra details in the commit messages where applicable.

As a side note, how does replacing `basename` and `dirname` sound (in a separate PR)? Doing that will improve performance greatly (along with the $(pwd) i already did!) - asking because it's a bit more involved since the builtin facilities (like `${var%/*}`) don't automatically handle extra trailing slashes and whatnot, so I wanted to ask to be sure

Also, does the project specify a minimum Bash version - couldn't find any info on the website or source code? I saw some opportunities to cut out of some `cat` invocations, etc., but I held back for that reason. For my personal Bash projects, the minimum is `v4.3`, but I assume this projects supports the latest of the `v3` series due to compatibility with macOS?

Edit by @jthegedus:

Fixes #1313

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -16,9 +16,9 @@ find_cmd() {
   done
 
   if [ -f "$cmd_dir/$cmd_name" ]; then
-    printf "%s %s\\n" "$cmd_dir/$cmd_name" "$((args_offset + 1))"
+    printf "%s %s\n" "$cmd_dir/$cmd_name" "$((args_offset + 1))"
   elif [ -f "$cmd_dir/command.bash" ]; then
-    printf "%s %s\\n" "$cmd_dir/command.bash" 1
+    printf "%s %s\n" "$cmd_dir/command.bash" 1
   fi
 }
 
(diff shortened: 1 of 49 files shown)


jthegedus:
@hyperupcall Sorry for the delay in reviewing this, it looks like a lot of effort.

With changes like these, they can make all the other open PRs have conflicts, so we would prefer to have each change in its own PR so we can merge and test each of these updates independently.

Did you make these changes in an automated way with sed or something so we can just do one at a time?

Some of these changes to consistency have been on my list to fix, so I would like to see these get merged which would be easier to do individually.

jthegedus:
Regarding your questions on Bash version, we are trying to support POSIX shells as best we can. I believe currently there might be some Bash-specific syntax, but we might remove those in future depending on how we develop our support approach in future.

hyperupcall:
No worries - yeah these changes were semi-automated, I used VSCode's find and replace all since sed doesn't support PCRE's but I guess I can rewrite a quick Perl one liner for each of these changes and have those in separate PRs one at a time

Stratus3D:
@hyperupcall while the PR is too much to merge all at once, it would be great if we had some automated way of enforcing these things via github actions. We currently require shfmt and shellcheck to pass for PRs, is there another tool we can use to enforce better Bash conventions? I know I would typically NOT catch all these things when reviewing a PR manually.

hyperupcall:
@Stratus3D Not that I know about. The only other shell tool/linter utility that I know of is shellharden and bash-language-server, but they don't really have a solution for that. I've been meaning to build my own since I write a lot of Bash scripts and libraries but I haven't gotten around to it

hyperupcall:
I'm closing this issue since the changes it adds be contained in future, separate PRs

PR #1352, opened 2022-10-18
Fix wrong path

yus-ham:


Stratus3D:
Many tests fail with this change. I am closing the PR.

PR #1375, opened 2022-12-12
ci: monthly dependabot npm schedule

jthegedus:
# Summary

`monthly` is better here because we are likely to see many updates in the `npm` ecosystem, especially with the release candidates we're using for our docs site, and since we offer no APIs etc there is little risk delaying bugfixes in our static website

## Other Information

Trying to avoid Dependabot spam since it can't update more than one dependency per PR so we get ~5 at once


Stratus3D:
Excellent! Thank you @jthegedus !

PR #1384, opened 2022-12-18
fix: Remove unecessary backslashes

hyperupcall:
# Summary

This fixes things so in some places extraneous backslashes are removed. This makes things less confusing for people not accustomed to Bash's escaping behavior.

This doesn't fix all issues - only ones that could not be caught in an [automated](https://github.com/asdf-vm/asdf/pull/1349#issuecomment-1347347875) way.

This partially addresses https://github.com/asdf-vm/asdf/issues/1313, and is related to https://github.com/asdf-vm/asdf/pull/1349


Stratus3D:
Thanks for the PR @hyperupcall !

Can you refresh my knowledge of Bash's backslash escaping? I seem to remember I added double backslashes at some.

hyperupcall:
@Stratus3D Sure, here is a desc from https://github.com/asdf-vm/asdf/pull/1349/commits/3a80652a3231edaf993ba52b13d2cb89bae2a353, taken from the bash manual

> The backslash retains its special meaning only when followed by one of the following characters: $, `, ", \, or <newline>

This means that the literal value of the backslash is preserved since
it's followed by either an 'n', 't', '(', etc.

hyperupcall:
#1385 has a rule that checks and fixes this for `printf` stuff. Although now that I think about it, it should have been slightly modified to work with every double quoted string. I guess I can work out that change at the end

Stratus3D (approved):

PR #1385, opened 2022-12-18
fix: lint errors from `scripts/checkstyle.py`

hyperupcall:
# Summary

Sorry for the onslaught of PR's, this is going to be the last one until all the other ones are either closed or merged. This adds preliminary support for an [automated way](https://github.com/asdf-vm/asdf/pull/1349#issuecomment-1348549238) of enforcing particular styles that shellcheck doesn't catch.

Right now, it's a simple Python file (instead the originally proposed Perl for obvious readability reasons), which just checks a single "rule" - "unecessary printf backslashes".

The second commit uses this rule to format a single file like so:

```sh
$ ./scripts/checkstyle.py ./lib/functions/plugins.bash --fix
```

- If `--fix` is not supplied, then it only prints out the errors
- If no file paths are supplied, then it reads all of the files matching `*.{bash,sh,bats}`.
- The comand sometimes needs to be run multiple times if there is more than one style violation on a single line.

I have the regular expressions and will write auto-fixers for each of the fixed things from #1349, but I'm only adding them one-at-a-time as stated.

Soon I guess it would be good to add GitHub actions and stuff, but I think should that be in a separate PR? I'm also looking for feedback on the checkstyle script, hopefully it's readable enough?


jthegedus:
My personal opinions of Python aside, this looks fine to me as a starting point. I would prefer to use existing tooling if it exists with options to configure regex checks and fixes, but I couldn't find anything with a quick look.

Would be good to add Python to the `.tool-versions` file.

And adding the GitHub Action step now would be fine, no need to wait for another PR, just another step in the `.github/workflows/lint.yml` file.

hyperupcall:
Sounds good! - In that case I'll run the script on every file so there aren't a million different lint commits

hyperupcall:
I made those changed :+1: 

The regex was also updated to match within any case of double-quotes (not just after `printf` statements) as I mentioned [here](https://github.com/asdf-vm/asdf/pull/1384#issuecomment-1360160012). This caught one extra violation at `./lib/commands/command-current.bash:47`

.editorconfig
```diff
@@ -6,3 +6,6 @@ indent_size = 2
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+[*.py]
+indent_size = 4
```
jthegedus, line 11: 4 :+1:

scripts/checkstyle.py
```diff
@@ -0,0 +1,101 @@
...
+    BLUE = '\033[94m'
+    MAGENTA = '\033[95m'
+    CYAN = '\033[96m'
+    RESET = '\033[0m'
+    BOLD = '\033[1m'
+    UNDERLINE = '\033[4m'
+
+# Before: printf '%s\\n' '^w^'
+# After: printf '%s\n' '^w^'
+def noDoubleBackslashFixer(line: str, rule: Dict[str, str], m: re.Match[str]) -> str:
```
jthegedus, line 22: Not saying we should block this PR because of this, more curiosity than anything, is there a way for us to write some python tests and have them inline in this same file?

scripts/checkstyle.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+import re
+import sys
+import os
+import argparse
+from pathlib import Path
+from typing import List, Dict, Any # compat
+
```
jthegedus, line 8: Can we get a comment section here about Usage, capturing the same stuff you have in the PR title for future users who find this file?

scripts/checkstyle.py
jthegedus: For instance, I came across a tool last week which would be a good candidate to generate the regex with: https://github.com/pemistahl/grex
>automatically generating a single regular expression from user-provided test cases

jthegedus:
@Stratus3D 

Building on your comment - https://github.com/asdf-vm/asdf/pull/1349#issuecomment-1348549238 

>@hyperupcall while the PR is too much to merge all at once, it would be great if we had some automated way of enforcing these things via github actions. We currently require shfmt and shellcheck to pass for PRs, is there another tool we can use to enforce better Bash conventions? I know I would typically NOT catch all these things when reviewing a PR manually.

Is there a reason we can't capture this like we do `test/banned_commands.bats`?

While the automated fix is nice, we only need the automated fix once, then we can just ban with regex checks right?

scripts/checkstyle.py
hyperupcall: If I am understanding your first question correctly, each fixer function isn't inline because Python's lambas are expression-only and they aren't able to span multiple lines - a tradegy. I worry grex won't be general enough, but if it works well, then it might be useful.

jthegedus: My first question was just about where you can put tests for Python code. I am not familiar with the patterns. Ideally, we wouldn't populate the `scripts/` directory with all the file structure of a Python library. So can we write tests in the `scripts/checkstyle.py` file? If we can, can we capture each of the cases found in this particular PR as test cases. You said you already had to update the Regex to capture another case in `command-current.bash`, so there is already an example of us finding an edge case. Since we want to guard against regressions, test cases would be valuable.

hyperupcall: Yeah good idea - test case would be nice. I can just add a flag like `--internal-test-regex` and have an extra array for positive and negative cases within the `rules` array. Is it OK if I hold off on doing that until I add another rule (probably one that matches [this](https://github.com/asdf-vm/asdf/pull/1349/commits/16836ef766813968b17a784fdee789a1c0af86c8) changeset)

hyperupcall: About doing the right patterns for the tests, I chose not to do any convention from any particular ecosystem because putting everything in a single file is important for simplicity.

scripts/checkstyle.py
jthegedus: Thanks

scripts/checkstyle.py
jthegedus: Sounds good to me. Thanks

jthegedus:
@Stratus3D I am good with the changes here, but will leave final approval & merge to you.

jthegedus (approved):

jthegedus:
Cheers @hyperupcall

Issue #1366, opened 2022-11-07
bug: Ambiguous documentation


GeorgeSchafer:
### Describe the Bug

I was trying to install asdf and I ran across these [instructions](https://asdf-vm.com/guide/getting-started.html#_1-install-dependencies:~:text=macOS%3A,%23). The way this is written, it is not clear that I should install homebrew OR spack, not homebrew AND spack.

### Steps to Reproduce

1. Go [here](https://asdf-vm.com/guide/getting-started.html#_1-install-dependencies:~:text=macOS%3A,%23)
2. Install Homebrew
3. Waste two hours trying to install spack needlessly. 🤣 

Note: I am more amused than annoyed, just wanted to give a heads up.

### Expected Behaviour

Instructions explaining that I should install homebrew OR spack

### Actual Behaviour

Instructions implying I need homebrew AND spack

### Environment

```shell
OS:
Darwin [USER] 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64 x86_64

SHELL:
zsh 5.8.1 (x86_64-apple-darwin21.0)

ASDF VERSION:
v0.10.2

ASDF ENVIRONMENT VARIABLES:


No plugins installed
ASDF INSTALLED PLUGINS:
```

Stratus3D:
The instructions don't say either one should be installed. The assumption was you'd run the appropriate command for your package manager (the odd one being Brew which doesn't need any dependencies installed in advance).

Can you provide a link to docs for a similar project that don't have this issue? Most projects end up displaying install/setup instructions in a matrix like we do because there are different OS/package manager combinations.

jthegedus:
I have a suggested improvement in a comment in #1388 to address this, though spelling out this specific scenario is a bit much IMO.

jthegedus:
Closed in #1388

Issue #1387, opened 2022-12-20
bug: installing via sudo-less homebrew may not work (outside of /usr/local/bin)


devinrhode2:
### Describe the Bug

```
dotfiles ➤ brew install asdf

Running `brew update --auto-update`...
==> Auto-updated Homebrew!
Updated 2 taps (homebrew/core and homebrew/cask).
==> New Casks
devsquadron

You have 58 outdated formulae and 2 outdated casks installed.
You can upgrade them with brew upgrade
or list them with brew outdated.

==> Fetching dependencies for asdf: automake, gmp, coreutils, openssl@3 and unixodbc
==> Fetching automake
==> Downloading https://ghcr.io/v2/homebrew/core/automake/manifests/1.16.5
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/automake/blobs/sha256:f68481d06be7fa3f0a0881edb825a336e7f6548191c762d68bd817183b238f5a
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sha256:f68481d06be7fa3f0a0881edb825a336e7f6548191c762d68bd817183b2
######################################################################## 100.0%
==> Fetching gmp
==> Downloading https://ghcr.io/v2/homebrew/core/gmp/manifests/6.2.1_1
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/gmp/blobs/sha256:2436cd120e5678d67c24020a50cbbf7c0220e7ecaac63981335872b9d666bcad
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sha256:2436cd120e5678d67c24020a50cbbf7c0220e7ecaac63981335872b9d66
######################################################################## 100.0%
==> Fetching coreutils
==> Downloading https://ghcr.io/v2/homebrew/core/coreutils/manifests/9.1
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/coreutils/blobs/sha256:ae9dc313da2a3971c8e633d9f3262fd1bd431c303207b3346924aed60bb0965d
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sha256:ae9dc313da2a3971c8e633d9f3262fd1bd431c303207b3346924aed60bb
######################################################################## 100.0%
==> Fetching openssl@3
[...]

PR #1388, opened 2022-12-20
docs: improve promotion of git install method

devinrhode2:
This can generally avoid complications due to installing under a package manager (homebrew, etc).

Related: https://github.com/asdf-vm/asdf/issues/1387

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -35,11 +35,17 @@
 
 ## 2. Download asdf
 
-We recommend using Git, though there are other platform specific methods:
+We recommend using Git:
+```sh
+git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.11.0
+```
+See latest release version here: https://github.com/asdf-vm/asdf/releases
+
+
+Platform specific alternatives:
 
 | Method   | Command                                                                                                                                                             |
 | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


jthegedus:
Hi @devinrhode2

I was intending on making a change like this. I was actually going to propose we go a step further and have an Official and All methods section, this way we can put the `git` method twice. I don't want people to skip to the table instead of reading and just grab `homebrew` as the top item in the table, so was going to propose `git` in the table as well as featured prominently above.

Eg:

jthegedus:
Also, I need to fix the automation of the version change in the docs, so would prefer to leave it as the older one here so I can validate that.

jthegedus:
I updated this PR with my preferred suggested example 1 whereby we separate Official & Community download/install methods.

jthegedus:
@devinrhode2 Do you think this change will cover what you were initially concerned with?

@Stratus3D does this look okay to you? I think this might be enough to resolve #1366

devinrhode2:
Sorry to not have a lot of detail here, but it looks great

There is the bug related to sudo-less homebrew, probably more important than doc tweaking

jthegedus (approved):
@Stratus3D I am merging as I have tested thoroughly locally and we can always iterate on the docs. If there are issues after it deploys I will hotfix.

Edit: everything worked fine, no need to hotfix or anything.

Stratus3D:
Looks great @devinrhode2 and @jthegedus ! Sorry for the late review here.

PR #1391, opened 2022-12-21
ci: explicitly set shell in lint & format scripts

jthegedus:
# Summary

In #1313 I ran some analysis of our Shellcheck errors with the Shell set explicitly to POSIX. Having done this I thought it might be worth updating our `scripts/*` to explicitly check specific files for specific Shell adherence (something discussed in #1313).

What I discovered was that we were not testing some files or dirs, and also setting `bats` in Shellcheck introduced a whole set of errors which have clearly been missed with auto-detection.

~What I want with regards to review discussion is whether we actually capture `asdf.sh` as a POSIX file or whether we should capture it in the `.bash` linting? If we consider it Bash, should we rename it `asdf.sh -> asdf.bash`? Same applies to `/lib/asdf.sh`~

## TODO:

- [x] use long name flags in `scripts/*`
- [x] separate `shfmt` and `shellcheck` for explicitly scanning `.sh`, `.bats` and `.bash` files with the correct Shell settings.
- [x] add `--indent` flag on `shfmt`. `shfmt` has integration to read from `.editorconfig`, but it seems if you add `--language-dialect` flag it no longer reads formatting settings from `.editorconfig`, so we must match it manually with flags (as was done in the past).
	- we're still using 2-space indent 🤮 
- [x] add `scripts/format.bash` to run `shfmt` with correct settings across the codebase en masse.

Related: #1313, #1349

### DIFF
--- a/scripts/format.bash
+++ b/scripts/format.bash
@@ -4,6 +4,7 @@ set -euo pipefail
 
 # check .sh files
 shfmt --language-dialect posix --indent 2 --write \
+  asdf.sh \
   lib/*.sh
 
 # check .bash files
--- a/scripts/format.bash
+++ b/scripts/format.bash
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# check .sh files
+shfmt --language-dialect posix --indent 2 --write \
+  lib/*.sh
+
+# check .bash files
+shfmt --language-dialect bash --indent 2 --write \
+  completions/*.bash \
+  bin/asdf \
+  bin/private/asdf-exec \
+  lib/utils.bash \
+  lib/commands/*.bash \
+  lib/functions/*.bash \
+  scripts/*.bash \
+  test/test_helpers.bash \
+  test/fixtures/dummy_broken_plugin/bin/* \
+  test/fixtures/dummy_legacy_plugin/bin/* \
+  test/fixtures/dummy_plugin/bin/*
+
+# check .bats files
+shfmt --language-dialect bats --indent 2 --write \
+  test/*.bats
--- a/scripts/format.bash
+++ b/scripts/format.bash
@@ -3,9 +3,10 @@
 set -euo pipefail
 
 # check .sh files
-shfmt --language-dialect posix --indent 2 --write \
-  asdf.sh \
-  lib/*.sh
+# TODO(jthegedus): unlock this check later
+# TODO  shfmt --language-dialect posix --indent 2 --write \
+# TODO  asdf.sh \
+# TODO  lib/*.sh
 
 # check .bash files
 shfmt --language-dialect bash --indent 2 --write \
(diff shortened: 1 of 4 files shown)


hyperupcall:
I saw that the `tests/*` directory wasn't being ran through shellcheck whatsoever, so thanks for getting to this before I was able to.

It looks like `/lib/asdf.sh` is sourced by `/asdf.sh` so I'd say the file extension shouldn't be Bash.

jthegedus:
I have certainly missed things, still correcting and updating. This is a rabbit hole uncovering a lot of issues 😱

hyperupcall:
> Both asdf.sh & lib/asdf.sh throw errors when running Shellcheck with sh, because they're using Bash features. So should we change their names, or remove the Bash from the code, or do nothing?

I'd say we keep their names and keep the declaration at the top. Since we want CI to pass for everyone, we can just add `# shellcheck disable=...` lines before each Bashism for now, until we decide what to do with them later.

jthegedus:
>until we decide what to do with them later

Right, I was wondering if we should just solve that now. Happy to ignore specifics for these files now.

hyperupcall:
Oh wow yeah, there are 5 real violations in those files, yeah. In my opinion most of the file should be reworked anyways. The variables `ASDF_BIN` `ASDF_USER_SHIMS` shouldn't be capitalized and not all the variables are prefixed with an underscore. And I'm not sure why we're munging PATH - if really we don't want to append to the path twice, we can just set an environment variable with values (but a naive approach will break in several circumstances). So I guess that's to say I'm for doing everything separately, especially since if there is one mistake in this crutial code path, everything likely could break

jthegedus:
Yes, there are a lot of errors here.

@Stratus3D My suggested path forward is this:

* merge #1385 
* rebase this PR
* disable the check of `asdf.sh` & `lib/asdf.sh`
* go through and fix all reported issues (I can do this)
* merge this PR
* rebase all other PRs
* unlock and tackle `asdf.sh` & `lib/asdf.sh` in a separate PR

What do you think?

jthegedus:
I have commented out the updates to `scripts/*` that cause new errors, this way we can introduce them piecemeal over the next month or so. This way the branch won't have sweeping changes and become stale and reduce disruption to other PRs.

scripts/format.bash
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
Stratus3D, line 1: Why do we now have a `format.sh` and `shfmt.sh` with the same contents except `--write` flag?

jthegedus: `shfmt.bash` is what we use in CI and just prints incorrect formatting

I introduced `format.bash` as a script to be used locally to apply format changes. Previously, we had to manually fix the shfmt errors, or rely on people utilising editorconfig or shfmt IDE plugins to perform formatting. This makes it easier for us to blanket-apply changes.

Stratus3D: Excellent! Thank you @jthegedus !

PR #1399, opened 2022-12-27
fix: Remove usage of `$(pwd)` in favor of `$PWD`

hyperupcall:
As mentioned in the `checkstyle.py`, using the former increases execution time as a subshell has to be constructed. `pwd` being a builtin helps, but we can do better.

As suggested [here](https://github.com/asdf-vm/asdf/pull/1385#discussion_r1054018600), I added a way to validate the regular expressions


hyperupcall:
Rebased to make the same fix as #1409

jthegedus:
As per the conversation from #1409, this would qualify as a `fix:` as it makes code changes to the core of asdf.

jthegedus (approved):
@Stratus3D I know we discussed in the other PRs to rewrite this python script in Bash, and I am still all for that, but in the meantime feel we should merge this as it immediately improves the codebase.

Stratus3D (approved):

Stratus3D (changes requested):
It appears the checkstyle.py check is now failing due to the permission being changed on the file. If you can fix @hyperupcall I'll get this merged.

Stratus3D:
Looks like I was the one who broke permissions on the file. I updated and merge the PR. Thanks @hyperupcall !

PR #1400, opened 2022-12-27
fix: force lwrcase plugin name fix capitalization mismatch errs

### LINKED ISSUE #816: asdf Plugin Capitalization & Silent Failing
**Describe the bug**

If you add a plugin with different capitalization `asdf` will not function properly.
There are also no warnings for missing/mismatched plugins listed in the `.tool-versions`.

**To Reproduce**

Tell us what actions you performed before the issue occurred

For example if you added the following it would expect `Ruby` instead of `ruby`.

```
asdf plugin-add Ruby https://github.com/asdf-vm/asdf-ruby.git
```

If the `.tool-verions` has `ruby` instead of `Ruby` it will not install.
There's no warning for not having plugins added that are listed in a `.tool-versions`.
Also I think it should be case insensitive.

If you try to add the same plugin with different capitalization it will say it's already been added.

**Expected behavior**

I think it should be case insensitive when you run `asdf install`.

A warning would be nice if there's something listed in `.tool-versions` that you haven't yet added. It would be even nicer if it tried to push you in the right direction or asked you if you wanted to install the associated plugins.

**Actual behavior**

No error, it doesn't say anything installs if there's a capitalization mismatch.

### Environment

```shell
OS:
Darwin Coders-MBP 19.6.0 Darwin Kernel Version 19.6.0: Sun Jul  5 00:43:10 PDT 2020; root:xnu-6153.141.1~9/RELEASE_X86_64 x86_64

SHELL:
zsh 5.7.1 (x86_64-apple-darwin19.0)

ASDF VERSION:
v0.8.0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/usr/local/opt/asdf

ASDF INSTALLED PLUGINS:
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git
php                          https://github.com/asdf-community/asdf-php.git
ruby                         https://github.com/asdf-vm/asdf-ruby.git
```

**asdf plugins affected (if relevant)**:

**Screenshots**

**Additional context**

hyperupcall:
I know this was marked as Good First Issue, but there shouldn't be any silent failures whatsoever. It's also been unresolved for two years

Closes #816

### DIFF
--- a/lib/functions/plugins.bash
+++ b/lib/functions/plugins.bash
@@ -58,7 +58,7 @@ plugin_add_command() {
 
   local regex="^[[:lower:][:digit:]_-]+$"
   if ! printf "%s" "$plugin_name" | grep -q -E "$regex"; then
-    display_error "$plugin_name is invalid. Name must include only lowercase letters, numbers, '_', and '-'"
+    display_error "$plugin_name is invalid. Name may only contain lowercase letters, numbers, '_', and '-'"
     exit 1
   fi
 
--- a/lib/functions/plugins.bash
+++ b/lib/functions/plugins.bash
@@ -56,9 +56,9 @@ plugin_add_command() {
 
   local plugin_name=$1
 
-  local regex="^[[:alpha:][:digit:]_-]+$"
+  local regex="^[[:lower:][:digit:]_-]+$"
   if ! printf "%s" "$plugin_name" | grep -q -E "$regex"; then
-    display_error "$plugin_name is invalid. Name must match regex $regex"
+    display_error "$plugin_name is invalid. Name must include only lowercase letters, numbers, '_', and '-'"
     exit 1
   fi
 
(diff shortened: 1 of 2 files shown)


lib/functions/plugins.bash
```diff
@@ -56,9 +56,9 @@ plugin_add_command() {
 
   local plugin_name=$1
 
-  local regex="^[[:alpha:][:digit:]_-]+$"
+  local regex="^[[:lower:][:digit:]_-]+$"
   if ! printf "%s" "$plugin_name" | grep -q -E "$regex"; then
-    display_error "$plugin_name is invalid. Name must match regex $regex"
+    display_error "$plugin_name is invalid. Name must include only lowercase letters, numbers, '_', and '-'"
```
Stratus3D, line 61: What about instead of `Name must include only...` we phrased this `Name may only contain...`?

Stratus3D (approved):

hyperupcall:
That sounds better :+1: made those changes

jthegedus:
Looks good to me, thanks @hyperupcall

jthegedus:
This should probably have classed as `feat!:` as I believe it would constitute a breaking change.

PR #1402, opened 2022-12-27
feat: Provide arch and os to plugins

### LINKED ISSUE #1281: Provide arch & platform as vars to plugins

Different project use different variations of these values. I've had to write code to handle it in each plugin so it would be nice to have some premade vars to save all that work.

Include some globals for arch and platform with the ability to switch some on a per plugin basis

e.g
```
ASDF_ARCH_AMD=1

then if ARCH=x86_64 then ARCH=amd64

ASDF_PLATFORM_LCASE=1 then use lowercase PLATFORM var

```

nothing i'm aware of

various utils.bash files have custom code to do this.

hyperupcall:
Closes #1281

### DIFF
--- a/lib/functions/installs.bash
+++ b/lib/functions/installs.bash
@@ -182,19 +182,42 @@ install_tool_version() {
   if [ -d "$install_path" ]; then
     printf "%s %s is already installed\n" "$plugin_name" "$full_version"
   else
+    local asdf_os=
+    case $OSTYPE in
+    solaris*) asdf_os='solaris' ;;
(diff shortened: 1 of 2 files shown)


Stratus3D (commented):
Thanks @hyperupcall ! Review of this is pending discussion here - https://github.com/asdf-vm/asdf/issues/1281#issuecomment-1367314558

If we do end up merging this we'll need docs and tests as well.

Stratus3D:
Closing this. Please see https://github.com/asdf-vm/asdf/issues/1281#issuecomment-1380503299

Issue #941, opened 2021-05-17
`path:` versions can't contain `~`, lack documentation


Stratus3D:
**Describe the bug**

Any time a path version is set with a path that starts with a `~` asdf is unable to resolve the path to the right directory. It's not immediately clear why this is the case, but we may need to let the shell do some processing of the literal value before we lookup the directory.

**To Reproduce**

Tell us what actions you performed before the issue occurred

1. Install a version of elixir (say `1.10.2`) via asdf.
2. Locate the path then run `asdf local elixir path:~/.asdf/installs/elixir/1.10.2/ `

**Expected behavior**

asdf should have properly resolved this path as `$HOME/.asdf/installs/elixir/1.10.2/`.

**Actual behavior**

asdf outputs `version path:~/.asdf/installs/elixir/1.10.2/ is not installed for elixir`

**Docs**

`path` versions are only mentioned in the documentation on the page for `.tool-versions`. Documentation on the different versions types available should be more prominent and listed somewhere in the "Manage Versions" section of the documentation site. Storing under the `.tool-versions` section no longer makes sense.

PR #1403, opened 2022-12-27
fix: Allow `path:` versions to use `~`

hyperupcall:
# Summary

```sh
$ asdf plugin-add elixir https://github.com/asdf-vm/asdf-elixir.git
$ asdf install elixir 1.10.2
$ asdf local elixir path:~/.asdf/installs/elixir/1.10.2/
$ cat .tool-versions
elixir path:~/.asdf/installs/elixir/1.10.2/
$ command -v elixir
/home/edwin/.asdf/shims/elixir
```

Closes: #941

## Other Information

The [lib/functions/versions.bash](https://github.com/asdf-vm/asdf/compare/master...hyperupcall:allow-tilda-in-path?expand=1#diff-beaa406545def97144dc1b90615a7d48c27f241839b4ddff0df328ea4f6b9e5e) change is due to a bug that apparently didn't show up before (from [this](https://github.com/asdf-vm/asdf/commit/eb7dac3a2b15ad458f55a897d49a377508ea92fe)):

```sh
asdf local elixir path:~/.asdf/installs/elixir/1.10.2/
tail: cannot open '/tmp/tmp.OU7v1JCUaV/.tool-versions' for reading: No such file or directory
```


lib/utils.bash
```diff
@@ -263,7 +263,11 @@ find_install_path() {
     # And then use the binaries there
     local install_type="path"
     local version="path"
-    printf "%s\n" "${version_info[1]}"
+
+    util_resolve_user_path "${version_info[1]}"
+    local full_version="${REPLY}"
+
+    printf "%s\n" "$full_version"
```
Stratus3D, line 270: Why doesn't `util_resolve_user_path` print the resolved path directly? If it did we could simplify this to:

```bash
printf "%s\n" "$(util_resolve_user_path "${version_info[1]}")"
```

lib/utils.bash
```diff
@@ -319,8 +323,15 @@ parse_asdf_version_file() {
   if [ -f "$file_path" ]; then
     local version
     version=$(strip_tool_version_comments "$file_path" | grep "^${plugin_name} " | sed -e "s/^${plugin_name} //")
+
     if [ -n "$version" ]; then
-      printf "%s\n" "$version"
+      local full_version="$version"
+      if [[ "$version" == path:* ]]; then
+        util_resolve_user_path "${version#path:}"
+        full_version="path:$REPLY"
```
Stratus3D, line 331: Same comment as above. I don't see a need for this `REPLY` variable as the resolved path could be printed.

Stratus3D (commented):
Not sure about the `REPLY` variable. If we do want to use a variable I think it should have a more unique name, for example, `util_resolve_user_path_reply`.

hyperupcall:
@Stratus3D I remember seing your post on ASDF performance measurements, and my takeaway was that performance in ASDF could be better for `reshim`, but also just in general. And I found the `REPLY` pattern to be extremely useful to preserve the performance in all of my Bash scripts, and I thought it would be useful here too.

It obviates the spawning of extraneous subshells, which in my experience solves the death by a thousand paper cuts that Bash performance usually boils down to (if you have different experience, I'd like to hear it!). The only downsides that I've come across is that you have to write the `unset -v REPLY; REPLY=` boilerplate, and that for extremely large functions, you may have to name `REPLY` variable to something else, or move the `REPLY=` stuff to the end of the function (depending on how you write it). If you wish, I've written more about `REPLY` [here](https://github.com/hyperupcall/awesome-basalt/blob/main/BEST_PRACTICES.md#functions) (it's a little out of date, but its correct enough).

But I do use it over named variables like the one you suggested because it is obvious what the return variable is (no need to check), since as you know Bash doesn't support returning actual values.

But for whatever reason, you don't like this pattern or it's not applicable as much here, then I can rewrite it to use the more traditional approach - what are your thoughts?

lib/utils.bash
Stratus3D: @hyperupcall based on your latest comment, could this be changed to:

```bash
util_resolve_user_path "${version_info[1]}"
printf "%s\n" "${REPLY}"
```

I'd also prefer the REPLY variable be named something more specific, so as to better indicate it's origin. In this case, what about `$util_resolve_user_path_reply`?

```bash
util_resolve_user_path "${version_info[1]}"
printf "%s\n" "${util_resolve_user_path_reply}"
```

Stratus3D:
@hyperupcall that all makes sense, I'm just wondering if `REPLY` is the best choice of name. It's a stylistic thing I guess.

Stratus3D:
Looks like we encountered a schedule brownout for runners on Ubuntu 18.04. We'll need to address that at some point.

@hyperupcall did you see my PR comments above?

lib/utils.bash
hyperupcall: :+1:

lib/functions/versions.bash
```diff
@@ -64,7 +64,7 @@ version_command() {
     sed -i.bak -e "s|^$plugin_name .*$|$plugin_name ${resolved_versions[*]}|" "$file"
     rm -f "$file".bak
   else
-    # Add a trailing newline at the end of the file if missing and file present
```
Stratus3D, line 67: This change could probably be reverted.

Stratus3D (approved):

lib/functions/versions.bash
hyperupcall: Oops I don't think I intentionally removed that

Issue #1347, opened 2022-10-14
Add documentation for Linux Homebrew installs need asdf.sh added to .bashrc not .bash_profile


richtong:
### Is your feature request related to a problem? Please describe

In the current documentation, there is no entry for how to set up Homebrew on Linux (aka linuxbrew). I used the instructions for the Mac but in looking at it, there seem to be two instructions that are relevant for Git and Linux and then for Homebrew and Linux.

The net is that with Ubuntu (not sure about other versions of linux), The windowing manager Gnome will source .profile and then for each Gnome Terminal with run .bashrc, it will never actually run .bash_profile unless you are logging with ssh or a terminal shell (with CTRL-ALT-F3) for instance.

So the documentation should probably change. Since there is no /bin/sh installer for asdf, (which is too bad because it means on Linux, you always have to run asdf.sh from .bashrc just to set the PATH which is pretty inefficient.

### Describe the proposed solution

The quick solution is in the documentation site, add an entry for Linux and Homebrew which meshes together the Git and Linux and Homebrew and Mac instructions that is you should source

add to your .bashrc under Linux:

source "$(brew --prefix asdf)/libexec/asdf.sh"

### Describe similar `asdf` features and why they are not sufficient

The problem with the current documentation is that it does not include the linuxbrew case.

### Describe other workarounds you've considered

You really need the shim path to be set properly.

jthegedus:
Hi @richtong This is a setup none of the core team use, we're happy to accept a PR with your suggested changes.

PR #1404, opened 2022-12-27
docs: Add documentation for Linux Homebrew Bash installs

hyperupcall:
# Summary

When using Homebrew on Linux with Bash, as the linked issue states, instructions should specify to put `asdf` initialization in `.bashrc` instead of `.bash_profile. This implements that recommendation.
 
Fixes: #1347

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -84,6 +84,22 @@ Completions must be configured manually with the following entry in your `.bash_
 
 :::
 
+::: details Bash & Homebrew
+
+Add `asdf.sh` to your `~/.bashrc` with:
+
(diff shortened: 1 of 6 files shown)


Stratus3D (approved):
I only checked the english translation but this looks good to me. Thanks @hyperupcall !

hyperupcall:
All the documentation was copied and pasted from the `details Bash & Homebrew (macOS)` section for each language, with only the filenames changed - I don't understand Brazilian Portuguese or Mandarin :sweat_smile:

jthegedus:
This looks good to me. Merging. Thanks @hyperupcall

PR #1406, opened 2022-12-28
revert: "fix: lint errors from `scripts/checkstyle.py` (#1385)" due to breaking changes.

slewsys:
# Summary

This reverts commit 3492043241e466337c5965a6fe2e089147bc4152 due to breaking changes in some contexts from the addition of `python` to _~/.asdf/.tool-versions_. To reproduce the error, install `asdf` versions of `node` and `python`, then use `npm` to install `markov-pwgen`, whose installer runs a python script. Here is the output:

```shell
npm i -g markov-pwgen
```
> npm ERR! code 126
> npm ERR! path /home/alm/.asdf/installs/nodejs/18.12.1/lib/node_modules/markov-pwgen
> npm ERR! command failed
> npm ERR! command sh -c test -f dictionary.js || ./scripts/filter-wordlist.py
> npm ERR! No preset version installed for command python3
> npm ERR! Please install a version by running one of the following:
> npm ERR!
> npm ERR! asdf install python 3.10.9
> npm ERR!
> npm ERR! or add one of the following versions in your config file at /home/alm/.asdf/.tool-versions
> npm ERR! python 3.11.1
>
> npm ERR! A complete log of this run can be found in:
> npm ERR!     /home/alm/.npm/_logs/2022-12-28T16_47_59_838Z-debug-0.log

## Other Information

This pull request simply reverts commit 3492043241e466337c5965a6fe2e089147bc4152.  Perhaps the underlying issue is that _~/.tool-versions_ should override _~/.asdf/.tool-versions_?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ See [CONTRIBUTING.md in the repo](https://github.com/asdf-vm/asdf/blob/master/CO
 
 ## Community & Questions
 
-- [FAQ](https://asdf-vm.com/learn-more/faq.html)
+- [FAQ](https://asdf-vm.com/more/faq.html)
 - [![GitHub Issues](https://icongr.am/simple/github.svg?color=808080&size=16) GitHub Issues](https://github.com/asdf-vm/asdf/issues): report a bug or raise a feature request to the `asdf` core team
 - [![StackOverflow Tag](https://icongr.am/fontawesome/stack-overflow.svg?size=16&color=808080) StackOverflow Tag](https://stackoverflow.com/questions/tagged/asdf-vm): see existing Q&A for `asdf`. Some of the core team watch this tag in addition to our helpful community
 


slewsys:
This PR is failing with the error message that I can't decipher:

> Pull request has only one commit and it's not semantic; this may lead to a non-semantic commit in the base branch (see https://github.community/t/how-to-change-the-default-squash-merge-commit-message/1155). Amend the commit message to match the pull request title, or add another commit.

If someone could enlighten me, I'd be grateful...

slewsys:
>  Is there something I can help jou with?

Hopefully the PR is self-explanatory?  The semantics of the commit message are clear to me.  Perhaps semantic-pr script is confused by two methods of referencing the reverted commit (i.e., #1385 and commit 3492043241e466337c5965a6fe2e089147bc4152)?

slewsys:
I incorrectly assumed that  `semantic-pr` was complaining about the PR message, but evidently it was complaining about the _original_ commit message, whose title began "Revert ..." instead of "revert: ...".   After correcting that, all seems well.  Thank you, Bo!

hyperupcall:
Hey, I'm the original commit author

I was recommended to add the `python` to the `[.tool-versions](https://github.com/asdf-vm/asdf/blob/master/.tool-versions)`, and I just (apparently incorrectly) assumed that it was used only during development.

Maybe someone else can shed more light onto this - in any case, it might be a good idea just to make a commit that removes the offending line, since the commit kind of changes a lot of stuff, not sure

slewsys:
> it might be a good idea just to make a commit that removes the offending line for now,

I'm playing with _~/.asdf/lib/utils.bash_ to try to identify the override.  Just to be clear, for the interim, you're okay with just removing  `python` from _~/.asdf/.tool-versions_?  Does that break CI?

hyperupcall:
> > it might be a good idea just to make a commit that removes the offending line for now,
> 
> I'm playing with _~/.asdf/lib/utils.bash_ to try to identify the override. Just to be clear, for the interim, you're okay with just removing `python` from _~/.asdf/.tool-versions_? Does that break CI?

It shouldn't break CI because the `checkstyle-py` runs on [`ubuntu-latest`](https://github.com/asdf-vm/asdf/blob/aa323086e842007ac37252d899ab063eb45975a8/.github/workflows/lint.yml#L38), which is [Ubuntu 22.04](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/), which uses [Python 3.10](https://pkgs.org/download/python3). It should work on older versions too, I tried to make things as compatable as possible, such as importing stuff from `Typing` when appropriate instead of using globals.

I'm definitely okay with removing `python` from `.tool-versions`, and it seems like that should fix the issue, but I don't think that's my call to make

jthegedus:
Because `asdf` resolves `.tool-versions` from the current directory, walking up towards the root `/`, the commands executed in `~/.asdf/installs/*` are resolving to the versions set in `~/.asdf/.tool-versions`.

As we do with `docs/.tool-versions` for managing tools for documentation development, I think the solution here is to move `python` from the repo root `.tool-versions` to `scripts/.tool-versions`.

@Stratus3D we should consider in future how we determine where to store `installs/` etc. Similar tools to `asdf` utilise `~/.cache/<tool name>` as the default location to store downloaded or installed packages.

@slewsys Please update this PR to not revert these changes, but to move `python` from `.tool-versions` in this repos root directory to a `/scripts/.tool-versions` file. Thanks

Stratus3D:
I'd much prefer to have `scripts/checkstyle.py` written in Bash to keep things as consistent as possible here. I avoid Python at all costs as I don't like hassling with virtual envs for a single script or package. Is there any reason I can't re-write this in Bash? It looks straightforward to me and would solve the reported issue.

hyperupcall:
The main reason is that dealing with regular expressions in Bash is more cumbersome and difficult because it supports less regular expression features. Bash doesn't support stuff like named capture groups, positive/negative lookaheads/lookbehinds, and in general I didn't want to find out halfway through writing it that I needed a feature that Bash doesn't support. And as you know Bash is much slower and I really didn't want to find out by how much.

I just tested it on Python `3.6.15` (Debian 10 supports 3.7), and it works flawlessly (with a single changed line, I used a type that wasn't exported on older versions, I'll make a PR for that soon).

I agree, I wouldn't want to manage with virtualenvs or worry about asdf version switching either, so I made the script as compatable as possible. If you really wish, I can rewrite to Perl which I guess might be better, but I have no intentions to write this in Bash, _especially_ to be compatable with macOS default 3.X.

slewsys:
macOS bash 3.2 actually does have regular expression capture groups.  What it doesn't have are associative arrays, references (i.e., local -n ref=$1) and the`mapfile` builtin, among many others.

```shell
foo='1bc2e3g' IFS="|"; [[ ."$foo" =~ \.(.+c)(.+e)(.+g) ]] && echo "${BASH_REMATCH[*]:1}"
```
> 1bc|2e|3g

Looking at checkstyle.py, I don't actually see a problem implementing it in macOS bash, though if speed is an issue, then perl probably has the advantage.

slewsys:
Oops - _named_ capture groups are not supported in bash, yeah.

slewsys:
Upon  reinspection, writing checkstyle.py  without bash associative arrays wouldn't be pretty.  Sorry for the noise.

Regarding `checkstyle.py`, Python type checker,  `pyright`,  is flagging these issues:
- In noDoubleackslashFixer, the argument **rule** is not used (and per [Python docs](https://docs.python.org/3/library/typing.html#typing.Dict),  **Mapping** is preferred to **Dict** for function arguments).
-  The call to `rule['fixer_fn'](line, rule, m)` is flagged because a string "is not callable".
- Similarly, the call `rule['found'] += 1` is flagged because a string can't be incremented.

slewsys:
```diff
diff --git a/scripts/checkstyle.py b/scripts/checkstyle.py
index 18e3cfa..4d0ea35 100755
--- a/scripts/checkstyle.py
+++ b/scripts/checkstyle.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 import re
-import sys
 import os
 import argparse
 from pathlib import Path
@@ -32,7 +31,7 @@ class c:
 
 # Before: printf '%s\\n' '^w^'
 # After: printf '%s\n' '^w^'
-def noDoubleBackslashFixer(line: str, rule: Dict[str, str], m: re.Match[str]) -> str:
+def noDoubleBackslashFixer(line: str, m]) -> str:
     prestr = line[0:m.start('match')]
     midstr = line[m.start('match'):m.end('match')]
     poststr = line[m.end('match'):]
@@ -40,7 +39,7 @@ def noDoubleBackslashFixer(line: str, rule: Dict[str, str], m: re.Match[str]) ->
     fixed_line = f'{prestr}{midstr[1:]}{poststr}'
     return fixed_line
 
-def lintfile(filepath: Path, rules: List[Dict[str, str]], options: Dict[str, Any]):
+def lintfile(filepath: Path, rules: List[Dict[str, Any]], options: Dict[str, Any]):
     content_arr = filepath.read_text().split('\n')
 
     for line_i, line in enumerate(content_arr):
@@ -58,7 +57,7 @@ def lintfile(filepath: Path, rules: List[Dict[str, str]], options: Dict[str, Any
                 print()
 
                 if options['fix']:
-                    content_arr[line_i] = rule['fixer_fn'](line, rule, m)
+                    content_arr[line_i] = rule['fixer_fn'](line, m)
 
                 rule['found'] += 1
```

hyperupcall:
I guess the `import sys` can be removed and the Dict` can be renamed to `Mapping`, but the rest seem like false positives

jthegedus:
> @jthegedus @hyperupcall I'd much prefer to have `scripts/checkstyle.py` written in Bash to keep things as consistent as possible here. I avoid Python at all costs as I don't like hassling with virtual envs for a single script or package. Is there any reason I can't re-write this in Bash? It looks straightforward to me and would solve the reported issue.

My intent with the merge of the initial script was to get it into the codebase and lint CI workflow. We can certainly re-write it in whatever.

As a hotfix for `master`, we can simply move the `python` from `.tool-version` to `scripts/.tool-version` now. Then, if a rewrite in bash happens later, remove it.

Whilst having linted python code would be good, I don't want to introduce any infrastructure to this repo for a single python script we use for linting.

PR #1408, opened 2022-12-29
fix: mv dev dep from repo root to subdir to avoid clash

slewsys:
# Summary

Move python from `~/.asdf/.tool-versions` to `~/.asdf/scripts/.tool-versions`. Prior to this change, Python scripts under `~/.asdf` attempted to ~open~ use the python version defined in `~/.asdf/.tool-versions`, which generally won't be installed.

## Other Information

See also #1406 for discussion about this issue.

Edited by: @jthegedus


Stratus3D:
@jthegedus @hyperupcall I'd much prefer to have `scripts/checkstyle.py` written in Bash to keep things as consistent as possible here. I avoid Python at all costs as I don't like hassling with virtual envs for a single script or package. Is there any reason I can't re-write this in Bash? It looks straightforward to me and would solve the reported issue.

jthegedus:
> @jthegedus @hyperupcall I'd much prefer to have `scripts/checkstyle.py` written in Bash to keep things as consistent as possible here. I avoid Python at all costs as I don't like hassling with virtual envs for a single script or package. Is there any reason I can't re-write this in Bash? It looks straightforward to me and would solve the reported issue.

My intent with the merge of the initial script was to get it into the codebase and lint CI workflow. We can certainly re-write it in whatever.

As a hotfix for `master`, we can simply move the `python` from `.tool-version` to `scripts/.tool-version` now. Then, if a rewrite in bash happens later, remove it.

Whilst having linted python code would be good, I don't want to introduce any infrastructure to this repo for a single python script we use for linting.

[copied from my reply here](https://github.com/asdf-vm/asdf/pull/1406#issuecomment-1367801404)

jthegedus (approved):

PR #1398, opened 2022-12-23
chore: bump dev deps

jthegedus:
# Summary

Bump the tools in `.tool-versions` to the latest.

## Other Information

I am going on leave over the holiday period so may not be able to respond of complete this PR. Please feel free to resolve and merge in my absence.

PR #1409, opened 2022-12-29
fix: Do not use type not exported on older Python versions

hyperupcall:
# Summary

This fixes an issue I mentioned in #1406, [here](https://github.com/asdf-vm/asdf/pull/1406#issuecomment-1367642291). Now, the `checkstyle.py` script runs with Python `3.6.15`

### DIFF
--- a/scripts/checkstyle.py
+++ b/scripts/checkstyle.py
@@ -32,7 +32,7 @@ class c:
 
 # Before: printf '%s\\n' '^w^'
 # After: printf '%s\n' '^w^'
-def noDoubleBackslashFixer(line: str, rule: Dict[str, str], m: re.Match[str]) -> str:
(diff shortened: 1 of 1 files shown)


jthegedus (approved):

jthegedus:
I should have validated this more thoroughly, as I don't think this change qualifies as a `fix:` in the sense that [release-please](https://github.com/googleapis/release-please#how-should-i-write-my-commits) thinks.

A change to a CI or dev-dep shouldn't qualify as a version change to the codebase unless, as in #1398 , they introduce code changes to the tool itself.

hyperupcall:
Ah, you're right - I guess `style` or `types` may have been appropriate

jthegedus:
In this case, `ci` or `chore` would have been preferable.

It was my mistake. Thanks for understanding.

Issue #1382, opened 2022-12-15
nushell support


brianmay:
### Describe the Bug

```
> source ~/.asdf/asdf.nu
Error: nu::shell::unsupported_input (link)

  × Unsupported input
    ╭─[Host Environment Variables:62:1]
 62 │ "VIRTUAL_ENV_DISABLE_PROMPT"="1"
 63 │ "ASDF_DIR"="/home/brian/.asdf"
    ·            ─────────┬─────────
    ·                     ╰── first does not support this input type
 64 │ "LESS"="-R"
    ╰────
 ```
 
 If I work around this, then I get a brew error (I do not have bew installed):
 
 ```
 Error: nu::shell::unsupported_input (link)

  × Unsupported input
   ╭─[asdf.nu:6:1]
 6 │
 7 │ let is_brew_installed = (which 'brew' | get --ignore-errors path | default '' | str trim | str length | into bool | first )
   ·                                                                            ─┬
   ·                                                                             ╰── first does not support this input type
 8 │
   ╰────
```

### Steps to Reproduce

```
> $env | get --ignore-errors ASDF_DIR | default ''
/home/brian/.asdf

> $env | get --ignore-errors ASDF_DIR | default '' | first
Error: nu::shell::unsupported_input (link)

  × Unsupported input
    ╭─[Host Environment Variables:62:1]
 62 │ "VIRTUAL_ENV_DISABLE_PROMPT"="1"
 63 │ "ASDF_DIR"="/home/brian/.asdf"
    ·            ─────────┬─────────
    ·                     ╰── first does not support this input type
 64 │ "LESS"="-R"
    ╰────
```

### Expected Behaviour

No errors

### Actual Behaviour

Errors as per above

### Environment

```shell
OS:
Linux canidae 5.10.0-19-amd64 #1 SMP Debian 5.10.149-2 (2022-10-21) x86_64 GNU/Linux

SHELL:
zsh 5.8 (x86_64-debian-linux-gnu)

ASDF VERSION:
v0.11.0-6a4f51a

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/home/brian/.asdf

ASDF INSTALLED PLUGINS:
elixir                       https://github.com/asdf-vm/asdf-elixir.git master d08f506
erlang                       https://github.com/asdf-vm/asdf-erlang.git master 98139db
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master b2d06a7
[...]

PR #1415, opened 2023-01-03
fix: Add test for nushell integration and fix some bugs

fwfurtado:
# Summary

Fix bug of using `first`  when hasn't a value for an expression
Fix support for `ASDF_DIR` env var

Add tests for nushell integration

Closes: #1382

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   ELVISH_VERSION: v0.18.0
+  NUSHELL_VERSION: 0.73.0
 
 jobs:
   nix:
@@ -26,16 +27,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0  
-      - name: Setup nushell
-        uses: hustcer/setup-nu@main 
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
-          
+          fetch-depth: 0
 
       - name: Install test dependencies
         if: runner.os == 'macos'
-        run: brew install coreutils fish elvish
+        run: brew install coreutils fish elvish nushell
 
       - name: Install test dependencies
         if: runner.os == 'linux'
@@ -44,12 +40,22 @@ jobs:
           sudo apt-get update
           sudo apt-get -y install fish curl
 
+          # Create $HOME/bin
+          mkdir -p "$HOME/bin"
+
           # Download elvish binary and add to path
           curl https://dl.elv.sh/linux-amd64/elvish-${{ env.ELVISH_VERSION }}.tar.gz -o elvish-${{ env.ELVISH_VERSION }}.tar.gz
           tar xzf elvish-${{ env.ELVISH_VERSION }}.tar.gz
           rm elvish-${{ env.ELVISH_VERSION }}.tar.gz
-          mkdir -p "$HOME/bin"
           mv elvish-${{ env.ELVISH_VERSION }} "$HOME/bin/elvish"
+
+          # Download nushell binary and add to path
+          curl -L https://github.com/nushell/nushell/releases/download/${{ env.NUSHELL_VERSION }}/nu-${{ env.NUSHELL_VERSION }}-x86_64-unknown-linux-gnu.tar.gz -o nu-${{ env.NUSHELL_VERSION }}-x86_64-unknown-linux-gnu.tar.gz
+          tar xzf nu-${{ env.NUSHELL_VERSION }}-x86_64-unknown-linux-gnu.tar.gz
+          rm nu-${{ env.NUSHELL_VERSION }}-x86_64-unknown-linux-gnu.tar.gz
+          mv nu-${{ env.NUSHELL_VERSION }}-x86_64-unknown-linux-gnu/* "$HOME/bin"
+
+          # Add $HOME/bin to path
           echo "$HOME/bin" >>"$GITHUB_PATH"
 
       - name: Install bats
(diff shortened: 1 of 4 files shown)


Stratus3D:
Thanks for the PR! It looks like tests are failing on MacOS 10.15. Can you take a look?

Also, is rust required for nushell? Is there a way to install it without rust? I would like to keep the number of CI dependencies to a minimum.

fwfurtado:
> Thanks for the PR! It looks like tests are failing on MacOS 10.15. Can you take a look?
> 
> Also, is rust required for nushell? Is there a way to install it without rust? I would like to keep the number of CI dependencies to a minimum.

No problem! I've fixed the failing test on MacOS 10.15 a while ago.

Unfortunately the ubuntu hasn't a package of nushell, and according installation guide we need to download the nushell via **cargo**

fwfurtado:
> Thanks for the PR! It looks like tests are failing on MacOS 10.15. Can you take a look?
> 
> Also, is rust required for nushell? Is there a way to install it without rust? I would like to keep the number of CI dependencies to a minimum.

No problem! I've fixed the failing test on MacOS 10.15 a while ago.

Unfortunately the ubuntu hasn't a package of nushell, and according installation guide we need to download the nushell via **cargo**

test/asdf_nu.bats
```diff
@@ -0,0 +1,106 @@
...
+    let-env ASDF_NU_DIR = '$PWD'
+
+    source asdf.nu 
+
+    echo \$env.ASDF_DIR
+  ")
+
+  [ "$?" -eq 0 ]
+  output=$(echo "$result" | grep "asdf")
+  [ "$output" != "" ]
```
Stratus3D, line 27: Would it be possible to assert that the path is correct here?

test/asdf_nu.bats
```diff
@@ -0,0 +1,106 @@
...
+    let-env PATH = ( '$(cleaned_path)' | split row ':' )
+    let-env ASDF_NU_DIR = '$PWD'
+
+    source asdf.nu 
+
+    echo \$env.PATH
+ ")
+  [ "$?" -eq 0 ]
+  output=$(echo "$result" | grep "asdf")
+  [ "$output" != "" ]
```
Stratus3D, line 43: Same as above.

Stratus3D:
I'm looking at the runtimes of the tests on Ubuntu I'm seeing them increase from around 2 minutes on latest master to about 8 minutes on this PR branch. I'm not sure I want to increase build times that much just so we can have tests for nushell. Perhaps we could just limit them to running on MacOS where nushell install is fast?

asdf.nu
```diff
@@ -1,26 +1,18 @@
-let possible_dirs = [ 
-                        ($env | get --ignore-errors ASDF_DIR | default '' | first) , 
-                        '/opt/asdf-vm/asdf.nu',
```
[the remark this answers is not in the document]
fwfurtado: This is not needed anymore once this was an attempt to predict the installation path. However when I realized how fish and elvish do this thing I decided to change this to use the `ASDF_DIR` env var instead

fwfurtado:
> I'm looking at the runtimes of the tests on Ubuntu I'm seeing them increase from around 2 minutes on latest master to about 8 minutes on this PR branch. I'm not sure I want to increase build times that much just so we can have tests for nushell. Perhaps we could just limit them to running on MacOS where nushell install is fast?

@Stratus3D I'd change the ubuntu tests to use a binary version of nushell. 
Can you check if this build/test time is acceptable or if we need to disable nushell's tests for Ubuntu?

Although, I've needed to skip tests in Ubuntu 18.04 becase this don't support minimal  version of GCLib required for nushell latest.

test/asdf_nu.bats
```diff
@@ -0,0 +1,127 @@
...
+setup() {
+  cd $(dirname "$BATS_TEST_DIRNAME")
+}
+
+cleaned_path() {
+  echo $PATH | tr ':' '\n' | grep -v "asdf" | tr '\n' ':'
+}
+
+@test "exports ASDF_DIR" {
+  if [ "${SKIP_NUSHELL_TESTS}" = 'YES' ]; then
```
Stratus3D, line 14: Is it not possible to skip the entire file with a single check at the top?

e.g. something like this:

```
if [ "${SKIP_NUSHELL_TESTS}" = 'YES' ]; then
  echo "skipping nushell tests"
  exit 0
fi
```

.github/workflows/tests.yml
```diff
@@ -56,3 +67,4 @@ jobs:
         run: bats test
         env:
           GITHUB_API_TOKEN: ${{ github.token }}
+          SKIP_NUSHELL_TESTS:  "${{ matrix.os == 'ubuntu-18.04' && 'YES' || 'NO' }}"
```
Stratus3D, line 70: The `|| 'NO'` is probably unnecessary since you explicitly check for `YES` in the test code.

Stratus3D:
Left two more comments, otherwise looks great!

Stratus3D:
I guess my comment about exiting early from the nushell tests was wrong. Now it's saying the tests have failed when they are skipped. I did find this on SO - https://stackoverflow.com/questions/49991275/is-there-a-way-to-bail-from-a-bats-test

test/asdf_nu.bats
fwfurtado: According to [FAQ](https://bats-core.readthedocs.io/en/stable/faq.html#how-can-i-exclude-all-tests-of-a-file-from-a-test-run) of bats it is not possible.

When I tried this the test was broken.

fwfurtado:
> I guess my comment about exiting early from the nushell tests was wrong. Now it's saying the tests have failed when they are skipped. I did find this on SO - https://stackoverflow.com/questions/49991275/is-there-a-way-to-bail-from-a-bats-test

I moved the skip statement to `setup` function

Stratus3D:
Thanks for addressing all the comments @fwfurtado ! 🚀

Stratus3D (approved):

PR #1420, opened 2023-01-06
ci: Stop using Ubuntu 18.04

hyperupcall:
# Summary

As mentioned [here](https://github.com/asdf-vm/asdf/pull/1403#issuecomment-1372163612), we encountered a [brownout for Ubuntu 18.04](https://github.com/actions/runner-images/issues/6002). This fixes that

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,6 @@ jobs:
           - macos-10.15
           - macos-11
           - macos-12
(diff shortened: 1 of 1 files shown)


jthegedus:
Wow, the announcement to stop supporting 18.04 is surprising. I would have expected an org like GitHub to follow the Extended Maintenance timelines of major OSs.

PR #1421, opened 2023-01-06
fix: Remove `==` inside `[`

hyperupcall:
# Summary

This is the last PR that implements what #1349 originally proposed that isn't covered by shellcheck. It ensures that all calls to `[` use single equal sign, even if Bash enables double equals sign. This makes the code more consistent, and code snippets more copy-pastable to POSIX.


jthegedus:
As this changes core asdf code, not just tests, this will require a `fix:` PR title to cause a patch bump.

hyperupcall:
:+1:

Stratus3D (approved):

PR #1422, opened 2023-01-06
test: Improve Elvish tests

hyperupcall:
# Summary

Old versions of Elvish cause spectacular failures

image

My patches fix that:

image

The module handling was also improved. Rather than copying the `adsf.elv` file and munging with user config files (which was already broken in cases where `$XDG_CONFIG_HOME` was not set to `$HOME/.config`), it uses the files in place by using relative imports. Additionally, `XDG_CONFIG_HOME` and `XDG_DATA_HOME`, etc. were set to empty so user config files are never parsed (it seems the `-norc` flag of Elvish is limited), improving test isolation.

### DIFF
--- a/test/asdf_elvish.bats
+++ b/test/asdf_elvish.bats
@@ -3,9 +3,7 @@
 load test_helpers
 
 setup() {
-  cd $(dirname "$BATS_TEST_DIRNAME")
-  mkdir -p $HOME/.config/elvish/lib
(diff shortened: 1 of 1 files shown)


jthegedus (approved):
This looks great! Using XDG is a good change, though if/when we implement our own usage of XDG this may need to change.

hyperupcall:
Good point - likely will

Related to #1351

PR #1423, opened 2023-01-06
test: Do not run Nu test if it does not exist

hyperupcall:
# Summary

This is more automatic compared against `SKIP_NUSHELL_TESTS`.


test/asdf_nu.bats
```diff
@@ -3,10 +3,11 @@
...
 
 setup() {
-  if [ "${SKIP_NUSHELL_TESTS}" = 'YES' ]; then
-    skip "skipping nushell tests"
-  fi
   cd $(dirname "$BATS_TEST_DIRNAME")
+
+  if ! command -v nu; then
+    skip "Nu is not installed"
+  fi
```
Stratus3D, line 10: Just curious: why did you move this `if` from the top of the `setup` function to the bottom?

Stratus3D (approved):
Changes look good to me but I'm going to wait until CI passes to merge.

test/asdf_nu.bats
hyperupcall: I believe I wanted to make it clear the the current directory was an invariant with respect to the test file, so moving it to the very top would accomplish that. I think it matches other test files as well

PR #1427, opened 2023-01-13
ci: fix regex in `update_version_in_docs.bash` script

elebow:
The previous one was out of sync with the current doc formatting.


jthegedus (approved):
Thanks, I had been meaning to do this for ages!

I reverted the docs update so I can test running the actual CI job using the fix of the script. I tested locally and it works great, cheers!

PR #1429, opened 2023-01-13
ci: use release-please to bump versions in docs

jthegedus:
# Summary

I am not sure why this was failing to execute. It may be because of the single quotes around the asterisk, or not, I have no idea. We should probably just use the actual `release` trigger for this workflow anyway.

Related #1427


Stratus3D (approved):
Seems fine to me. I'm not sure why the current code isn't working either. From what I've read it should.

jthegedus:
I will test this in a different repo to confirm before merging. I have a feeling it is because of the single quotes around the asterisk. 

```diff
-   - '*'
+   - *
```

jthegedus:
Okay, so I figured out why it wasn't working. It is because of something I missed in the `release-please` docs:

>When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workolow runs.
>
>https://github.com/google-github-actions/release-please-action#github-credentials

Release-please uses the `GITHUB_TOKEN` by default to function. You can get around this by using a Personal Access Token. But managing ownership of those keys is hard in an organisation without creating a bot account to own it (this is definitely not my preferred solution).

While reading the docs I came across an alternative solution. The `release-please-action` has the capability of being able to replace version numbers in arbitrary files already. It just requires a comment in the file to define the boundaries which it will attempt to perform replacements. This significantly simplifies the whole process.

The only caveat is that it I needed to change the `SECURITY.md` file to replace the table of versions with a single one. The table solution wasn't a useful mechanism long-term.

---

I know this isn't a `script/` which we can run locally, but the fact the implementation is maintained by someone else and works without extension investigation is a positive in my book.

jthegedus:
I will merge this now to work on other aspects, we can undo it later if you have objections or alternatives 👍

jthegedus:
And immediately after merging I realised we're not bumping versions in the translated files either.

PR #1432, opened 2023-01-16
docs: Properly quote variables in instructions

hyperupcall:
# Summary

ASDF is extremely popular and used by many people. I think it is a good idea to show best practices whenever shell code is written in. Mainly, the proper quoting of variables.

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -45,7 +45,6 @@ We highly recommend using the official `git` method.
 | Homebrew | `brew install asdf`                                                                                                                                                 |
 | Pacman   | `git clone https://aur.archlinux.org/asdf-vm.git && cd asdf-vm && makepkg -si` or use your preferred [AUR helper](https://wiki.archlinux.org/index.php/AUR_helpers) |
 
-
 ## 3. Install asdf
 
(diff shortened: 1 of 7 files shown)


docs/guide/getting-started.md
```diff
@@ -263,7 +262,7 @@ Completions are placed in a ZSH friendly location, but [ZSH must be configured t
 Add `asdf.nu` to your `~/.config/nushell/config.nu` with:
 
 ```shell
-"\nlet-env ASDF_NU_DIR = ($env.HOME | path join '.asdf')\n source " + ($env.HOME | path join '.asdf/asdf.nu') | save --append $nu.config-path
```
jthegedus, line 266: I agree we should use Shell best-practices in the documentation code snippets, however, I am not so sure if this is considered best practice for Nushell. Also iffy about Fish.

hyperupcall: Yeah the non-POSIX shells could have better snippets. Since I don't deal much with those shells I didn't touch their code snippets (besides the trailing whitespace)

jthegedus: Oh I see! 😅 silly me, I didn't read the diff properly. Thanks!

jthegedus (approved):
Looks good to me! Thanks!

Issue #1396, opened 2022-12-23
bug: enable shellcheck on bats tests


jthegedus:
### Describe the Bug

#1391 updated the `scripts/*.bash` to add a specific Shellcheck command to validate our `test/*.bats` code which was previously being missed by "automatic" detection rules.

The updated checking was introduced but disabled. We should:

- enable the check
- fix the results

### Steps to Reproduce

Uncomment the code in https://github.com/asdf-vm/asdf/blob/master/scripts/shellcheck.bash

```diff
# check .bats files
- # TODO(jthegedus): unlock this check later
- # TODO  shellcheck --shell bats --external-sources \
- # TODO  test/*.bats
+shellcheck --shell bats --external-sources \
+  test/*.bats
```

### Expected Behaviour

CI will run this script and reveal a bunch of issues to be fixed.

### Actual Behaviour

These are currently skipped.

### Environment

```shell
╭─jthegedus@art at 20:58:41
├ ~/dev/asdf
├  master
├ Last command took 596ms 
╰─λ asdf info
OS:
Linux art-win11 5.15.79.1-microsoft-standard-WSL2 #1 SMP Wed Nov 23 01:01:46 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
fish, version 3.3.1

ASDF VERSION:
v0.11.0-5af7625

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/home/jthegedus/.asdf

ASDF INSTALLED PLUGINS:
bats                         https://github.com/timgluz/asdf-bats.git main 299551f
deno                         https://github.com/asdf-community/asdf-deno.git master 87442fe
firebase                     https://github.com/jthegedus/asdf-firebase.git main 4a0ac90
gcloud                       https://github.com/jthegedus/asdf-gcloud main a71ab95
java                         https://github.com/halcyon/asdf-java.git master 0720a4b
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master c9e5df4
pnpm                         https://github.com/jonathanmorley/asdf-pnpm.git master 95752a9
python                       https://github.com/danhper/asdf-python.git master 8505457
shellcheck                   https://github.com/luizm/asdf-shellcheck.git master 0f79f0b
shfmt                        https://github.com/luizm/asdf-shfmt.git master a42c5ff
v                            https://github.com/jthegedus/asdf-v main 99136a7
```

### asdf plugins affected (if relevant)

NA

---

Related:
- #1397

PR #1433, opened 2023-01-16
chore: fix ShellCheck errors in tests

hyperupcall:
# Summary

This PR is the **first** of a [multi-part series](https://github.com/asdf-vm/asdf/compare/master...hyperupcall:asdf:shellcheck-tests) that fixes various issues with the tests. I broke this change up as to be easier to review, and it seemed similar to #1349, in which different PRs were preferred over a single larger one.

Here is a description of each commit/part:

- ~~Parts 1-5 are quite trivial and fix up obvious mistakes like quoting.~~
- Parts 1 does a few miscellaneous fixes
- Parts 2-5 are quite trivial and fix up obvious mistakes like quoting.
- Part 6 has the first non-obvious changes that remove use of `$?` by using the `run` function provided by Bats 
- Part 7 fixes all errors and warnings shown by Bats. The significant code change is that some regex comparison were changed to glob comparisons.
- Part 8 enables ShellCheck for warnings and errors
- Part 9 (the last commit) Fixes all infos, and enables ShellCheck to lint infos as well

At the end of the commit series, some lints were disabled:
- Codes [SC2031](https://www.shellcheck.net/wiki/SC2030) and [SC2030](https://www.shellcheck.net/wiki/SC2031), and [SC2164](https://www.shellcheck.net/wiki/SC2164) were disabled as being false positives, due to the Bats environment taking care of that for us.
- [SC2012](https://www.shellcheck.net/wiki/SC2012) was also disabled at the top of `test/reshim_command.bats` since I wanted to be sure not to break any tests and the conversion to `find` isn't as trivial as the other fixes - it can be fixed later.

Partially addresses #1396

### DIFF
--- a/test/asdf_elvish.bats
+++ b/test/asdf_elvish.bats
@@ -3,16 +3,17 @@
 load test_helpers
 
 setup() {
-  export XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_DATA_DIRS=
+  export XDG_CONFIG_HOME=
+  export XDG_DATA_HOME=
+  export XDG_DATA_DIRS=
 
-  local version=
-  version=$(elvish -version)
+  local ver_major=
+  local ver_minor=
+  local ver_patch=
+  IFS='.' read -r ver_major ver_minor ver_patch <<<"$(elvish -version)"
 
-  local ver_major= ver_minor= ver_patch=
-  IFS='.' read -r ver_major ver_minor ver_patch <<<"$version"
-
-  if ((ver_major == 0 && ver_minor <= 17)); then
-    skip "Elvish version is not at least 0.17"
+  if ((ver_major == 0 && ver_minor < 18)); then
+    skip "Elvish version is not at least 0.18. Found ${ver_major}.${ver_minor}.${ver_patch}"
   fi
 }
 
--- a/test/current_command.bats
+++ b/test/current_command.bats
@@ -134,8 +134,8 @@ foobar          1.0.0           $PROJECT_DIR/.tool-versions"
 }
 
 @test "current should handle comments" {
-  cd $PROJECT_DIR
-  echo "dummy 1.2.0  # this is a comment" >>$PROJECT_DIR/.tool-versions
+  cd "$PROJECT_DIR"
+  echo "dummy 1.2.0  # this is a comment" >>"$PROJECT_DIR/.tool-versions"
   expected="dummy           1.2.0           $PROJECT_DIR/.tool-versions"
 
   run asdf current "dummy"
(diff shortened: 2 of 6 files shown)


test/asdf_elvish.bats
```diff
@@ -3,12 +3,15 @@
 load test_helpers
 
 setup() {
+  # shellcheck disable=SC1007
   export XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_DATA_DIRS=
```
jthegedus, line 7: Though this make sense if you read the Shellcheck report for SC1007, it is easier for devs who don't know SC1007 and for all future contributors that we just change this code to:

```
  export XDG_CONFIG_HOME=
  export XDG_DATA_HOME=
  export XDG_DATA_DIRS=
```

Otherwise, we must mentally keep track of why a Shellcheck rule has been turned off. If I come back to this code 1yr from now, I don't want to have to look up the rule to be reminded of it, re-read the code to ensure it is being used correctly and decide that disabling the rule is the correct thing to do. All that for a minor convenience in syntax.

It also makes refactoring easier as we don't need to make this evaluation of whether the rule should still be disabled or not.

Generally, the preference would be to just change the code than disable a rule. It is more future proof.

Sorry for the nit.

test/asdf_elvish.bats
```diff
@@ -3,12 +3,15 @@
...
 
   local version=
   version=$(elvish -version)
 
+  # shellcheck disable=SC1007
```
jthegedus, line 12: Change to

```
  local ver_major=
  local ver_minor=
  local ver_patch=
```

for same reasons as above.

jthegedus (approved):
Pressed the wrong button 😫

jthegedus (changes requested):
See comments for requested changes.

I will approve & merge afterwards, cheers!

I know this is a slow process, sorry @hyperupcall your help is appreciated 🙏

test/asdf_elvish.bats
hyperupcall: :+1:

test/asdf_elvish.bats
```diff
@@ -3,12 +3,15 @@
...
   local ver_major= ver_minor= ver_patch=
+  # shellcheck disable=SC2034
```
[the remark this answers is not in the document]
hyperupcall: I changed everything except the `required_elvish_version_major` because I thought the meaning was already quite clear with `ver_major` and `ver_minor` comparison, next to the `skip` message.

Your changes also fix a bug in which the test's wouldn't properly show "skipped" if `elvish` wasn't installed. (my bad for that, I was the author that originally proposed these minimum version requirements).

There was also a bug in the "skip" message - the earliest version supported is 0.18 - the conditional was right, but the message was fixed to match.

hyperupcall:
Rebased over `master` :+1:

jthegedus (approved):
Looks great, thanks for the changes!

jthegedus:
On to the next!

hyperupcall:
woot!

Issue #1428, opened 2023-01-13
aesthetic feat: `asdf list` commands to sort output by version, not lexical


Kache:
### Is your feature request related to a problem? Please describe

`asdf list`, `asdf list python`, etc output version numbers that are lexicographically sorted, e.g.:

```sh
❯ asdf list python
  2.7.18
  3.10.0
  3.10.3
  3.10.6
 *3.10.9
  3.11.1
  3.6.15
  3.7.14
  3.7.6
  3.8.13
  3.9.11
```

But it'd be nicer to see them version-sorted (example below)

### Describe the proposed solution

Would like to see this:

```sh
❯ asdf list python
  2.7.18
  3.6.15
  3.7.6
  3.7.14
  3.8.13
  3.9.11
  3.10.0
  3.10.3
  3.10.6
 *3.10.9
  3.11.1
```

### Describe similar `asdf` features and why they are not sufficient

n/a

### Describe other workarounds you've considered

piping to `sort -V` kinda works, but the `*` messes it up

stephanGarland:
I would alternatively suggest moving away from `sort_versions()` in its current incarnation, and simplifying it as much as possible by utilizing tool's built-in availability. See my [PR here](https://github.com/asdf-vm/asdf-plugin-template/pull/58/files#diff-937b04fc6932f1d7026f08a65a8da13f4eb55a0be104478e717523a8b3eaf0e8R39) for asdf-plugin-template, and [issue comment here](https://github.com/asdf-vm/asdf/issues/290#issuecomment-1383056355) for a discussion on why I think this is advantageous.

In `sort_versions()` specifically, there are three piped commands, and that function is being piped into from another command. The following were done on a base-model Macbook Air M1, with zsh-5.8.1.

```
# First downloaded tagged refs to avoid network bias
❯ git ls-remote --tags --refs "https://github.com/python/cpython" > python_versions.txt

# And again, but using git ls-remote's built-in sort
❯ git -c 'versionsort.suffix=a' -c 'versionsort.suffix=b' -c 'versionsort.suffix=r' -c 'versionsort.suffix=p' -c 'versionsort.suffix=-' -c 'versionsort.suffix=_' ls-remote --exit-code --tags --ref --sort="version:refname" "https://github.com/python/cpython" > python_sorted_versions.txt

# Gather times for both, dumping stdout to /dev/null, and running 100 times for each
[...]

PR #1437, opened 2023-01-18
fix!: remove `use_release_candidate` option

### LINKED ISSUE #1515: bug: `asdf update --head` odd behavior
### Describe the Bug

Currently `asdf update --head` has a few incorrect underlying assumptions (`command-update.bash`)

- Assumes `master` is default branch (these days more people are configuring Git to use `main` by default)
- Assumes "default" remote name is `origin` (command breaks on rename)
- It also seems that the ` "asdf update --head should checkout the master branch"` test consistenly breaks when it is ran from a branch other than `master`

### Steps to Reproduce

`bats test`, `asdf update --head`

### Expected Behaviour

A `main` branch with configured upstream remote `me` properly is updated to latest `asdf` HEAD

### Actual Behaviour

- Branch `master` gets created

### Environment

```shell
OS:
Linux Arf 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.

ASDF VERSION:
v0.11.3-bbcbddc

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/home/edwin/.asdf

ASDF INSTALLED PLUGINS:
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master 644ada3
```

N/A

jthegedus:
# Summary

Now we are using [release-please](https://github.com/google-github-actions/release-please-action) & GitHub Releases to manage our changelog generation and release pipeline we're no longer using release candidates (as they were originally intended) with asdf core.

People who wish to test the latest code should use `HEAD` via `asdf update --head`.

If `HEAD` is too unstable for people and they want an RC equivalent, then we have two options:

1. we can implement `asdf update <sha>`
2. we can recommend users `cd` to asdf installation and perform a manual `git checkout <sha>`

Given no users have raised missing the RC I feel we are safe to go with option 2 here. We can add `asdf update <sha>` in future if it becomes necessary to do so.

## Other Information

This also removes `sort_versions` from the repository 🎉 

Related:
- https://github.com/asdf-vm/asdf/issues/1428 - we were discussing removing `sort_versions` which this PR does

Fixes #1515

### DIFF
--- a/lib/commands/command-update.bash
+++ b/lib/commands/command-update.bash
@@ -26,26 +26,19 @@ do_update() {
     printf "Updated asdf to latest on the master branch\n"
   else
     # Update to latest release
+    local sha_of_tag
+    local tag
+
+    # fetch tags from remote
     git fetch origin --tags || exit 1
 
-    if [ "$(get_asdf_config_value "use_release_candidates")" = "yes" ]; then
-      # Use the latest tag whether or not it is an RC
-      tag=$(git tag | sort_versions | sed '$!d') || exit 1
-    else
-      # Exclude RC tags when selecting latest tag
-      tag=$(git tag | sort_versions | grep -vi "rc" | sed '$!d') || exit 1
-    fi
+    sha_of_tag=$(git rev-list --tags --max-count=1) || exit 1
+    tag=$(git describe --tags "$sha_of_tag") || exit 1
 
     # Update
     git checkout "$tag" || exit 1
     printf "Updated asdf to release %s\n" "$tag"
   fi
-}
-
-# stolen from https://github.com/rbenv/ruby-build/pull/631/files#diff-fdcfb8a18714b33b07529b7d02b54f1dR942
-sort_versions() {
-  sed 'h; s/[+-]/./g; s/.p\([[:digit:]]\)/.z\1/; s/$/.z/; G; s/\n/ /' |
-    LC_ALL=C sort -t. -k 1,1 -k 2,2n -k 3,3n -k 4,4n -k 5,5n | awk '{print $2}'
 }
 
 update_command "$@"
--- a/defaults
+++ b/defaults
@@ -1,7 +1,6 @@
 # See the docs for explanations: https://asdf-vm.com/manage/configuration.html
 
 legacy_version_file = no
-use_release_candidates = no
 always_keep_download = no
 plugin_repository_last_check_duration = 60
 disable_plugin_short_name_repository = no
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -69,15 +69,6 @@ Plugins **with support** can read the versions files used by other version manag
 | `no` <Badge type="tip" text="default" vertical="middle" /> | Use `.tool-versions` to read versions                                      |
 | `yes`                                                      | Use plugin fallback to legacy version files (`.ruby-version`) if available |
 
-### `use_release_candidates`
-
-Configure the `asdf update` command to upgrade to the latest Release Candidate instead of the latest Semantic Version.
-
-| Options                                                    | Description               |
-| :--------------------------------------------------------- | :------------------------ |
-| `no` <Badge type="tip" text="default" vertical="middle" /> | Semantic Version is used  |
-| `yes`                                                      | Release Candidate is used |
-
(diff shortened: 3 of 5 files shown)


jthegedus:
Big question before we merge, is this considered a "breaking change"? I would think **_yes_**, as we're changing an external API, though it is only for users to update, but it could be used in a CI pipeline? 🤔 

Thoughts @Stratus3D

lib/commands/command-update.bash
```diff
@@ -26,26 +26,19 @@ do_update() {
...
     git fetch origin --tags || exit 1
 
-    if [ "$(get_asdf_config_value "use_release_candidates")" = "yes" ]; then
-      # Use the latest tag whether or not it is an RC
-      tag=$(git tag | sort_versions | sed '$!d') || exit 1
-    else
-      # Exclude RC tags when selecting latest tag
-      tag=$(git tag | sort_versions | grep -vi "rc" | sed '$!d') || exit 1
-    fi
+    sha_of_tag=$(git rev-list --tags --max-count=1) || exit 1
```
Stratus3D, line 35: Is this `git rev-list` command returning the most recently created tag, or the tag with the highest semantic version? 

If this is returning the most recently created tag, we may want to stick with the `sort_versions` approach here in case we ever tag a new patch version for old minor version of asdf. I know we've not done that thus far, so maybe it's not necessary.

Stratus3D (commented):
I feel like the old approach with `sort_versions` was more foolproof, as it always returned the right tag. It seems like the new code could return any tag, regardless of what name the tag was, as long as it was the newest.

jthegedus:
When I created this change I was not considering the scenario where we would manually create `rc` or `beta` tags.

I will update this PR with another suggestion.

Stratus3D:
> Big question before we merge, is this considered a "breaking change"?

I guess I never answered your question yesterday. Yes, I'd consider this a breaking change just to be safe.

lib/commands/command-update.bash
Stratus3D: Yeah, looking at this again I'd suggest keeping `sort_versions` for now. All other changes look good to me.

jthegedus: I have a different solution to propose here which would still remove `sort_versions`, I have just been busy with work so haven't had a chance to write it up. I would like to discuss that alternative before keeping `sort_versions`

jthegedus:
My last push was to just rebase before my new changes. I won't get to proposing my new changes until next week (2nd week April 2023).

The gist of my upcoming changes are:
- prepare
	- detect default branch
	- detect remote
	- git fetch from remote
- `asdf update`: use GitHub Release to detect "latest" version, then checkout the tag (as we do now)
	- manual GH Releases do NOT default to "latest", that is a checkbox to manually opt-in
	- automated release-please GH Releases will assign "latest"
- `asdf update --head`: checkout default branch and reset to HEAD (as it does now)
- `asdf update <sha|tag>`: checkout the specific SHA or Tag

Stratus3D:
Hi @jthegedus I've been working on a rewrite of asdf in Golang. I've got everything working in the new Go version. PR is here - https://github.com/asdf-vm/asdf/pull/1805. And dev binaries are available here - https://github.com/asdf-vm/asdf-core-go. 

With the work to rewrite asdf in Go, I felt it did not make sense to continue to support the `asdf update` command at all. Since the Go version ultimately runs as a binary, and most users will be using their OS package manager to download or build that binary. asdf won't be a Git repository on disk anymore, and it doesn't seem to me that an auto update feature that's independent from the OS package manager would be worth the trouble.

As for release candidate and other dev builds - we can have a GitHub workflow that uploads dev builds of the master branch to GitHub for users who want to try out the latest builds.

I already merged https://github.com/asdf-vm/asdf/pull/1806 today, but I can revert it if you think we should keep the `asdf update` feature. Thoughts?

Stratus3D:
@jthegedus I'm going to go ahead and close this. Please let me know what your thoughts are on this. Happy to re-open.

PR #1438, opened 2023-01-18
ci: only test on latest GitHub Runners

jthegedus:
# Summary

GitHub have deprecated the macOS 10.15 runner.

https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/

I am not sure of the value we're getting here. I originally wanted to execute our tests on each version of a Runner just to be certain about what we support, but it is clear, given GitHubs eager deprecation of macOS 15.10 that it is largely pointless and just wasteful of general resources and our time (on small PRs).

Related:
- https://github.com/asdf-vm/asdf/pull/1420

Performance of macOS Runner is still very poor:
- https://github.com/actions/runner-images/issues/1336
- https://github.com/actions/runner-images/issues/2247
- https://github.com/actions/runner-images/issues/2707
- https://github.com/actions/runner-images/issues/3885

Our `macos` jobs consistently take 100%-300% more time to complete than `ubuntu` 

## Other Information

GitHub should provide an option for Runners which executes on all current supported runners so as they deprecate our pipelines change, `ubuntu-all` & `macos-all` 🤔 Immutability of that is probably bad.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,14 +11,8 @@ env:
   NUSHELL_VERSION: 0.73.0
 
 jobs:
-  nix:
-    strategy:
-      fail-fast: false
-      matrix:
-        os:
-          - macos-latest
-          - ubuntu-latest
-    runs-on: ${{ matrix.os }}
+  ubuntu:
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks @jthegedus , I think this is the right choice for now, but I agree it was nice seeing the results for each runner OS.

Issue #290, opened 2018-02-21
asdf shims adding too much time to use them in a shell prompt


stevenocchipinti:
Hello,

Not sure if this is the right place to report this issue but I've noticed using `asdf` is quite slow for executing commands.

This is `ruby` using `chruby`:
```
$> time ~/.rubies/ruby-2.5.0/bin/ruby -v
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin16]

real	0m0.019s
user	0m0.009s
sys	0m0.006s
```

and `node` using `nvm`:
```
$> time ~/.nvm/versions/node/v9.1.0/bin/node -v
v9.1.0

real	0m0.015s
user	0m0.007s
sys	0m0.005s
```

### Steps to reproduce

This is `ruby` using `asdf`:
```
$> time /usr/local/opt/asdf/shims/ruby -v
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin16]

real	0m0.305s
user	0m0.130s
sys	0m0.126s
```

and `node` using `asdf`:
```
$> time /usr/local/opt/asdf/shims/node -v
v8.9.4

real	0m0.272s
user	0m0.119s
sys	0m0.109s
```

When having a shell prompt that prints the version number of `ruby` and `node`, this adds half a second to each command!

I love the idea of `asdf` but I don't want to give up version numbers in my prompt so I won't be using `asdf` in its current state unless I can find a workaround.

Any ideas or advice would be welcome - thanks

### Environment

**OS**: `OSX`

**asdf version**: `v0.4.2`

Stratus3D:
You are right, we need to investigate this. `asdf current` take almost a second to run on my machine:

```
asdf current  0.39s user 0.50s system 100% cpu 0.893 total
```

Now I have over half a dozen plugins installed, but even running it for just one version number still causes it to take a quarter of a second.

Stratus3D:
Reminder for myself: http://tldp.org/LDP/abs/html/optimizations.html

ppdeassis:
Same here. Showing `ruby`, `node` and `python` versions is adding almost a second (~.75) with `asdf` version `v0.5.0`.

```bash
→ time ~/.asdf/shims/ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin16]

real	0m0.333s
user	0m0.200s
sys	0m0.081s
```
```bash
→ time ~/.asdf/shims/node -v
[...]

Issue #1397, opened 2022-12-23
bug: enable shellcheck & shfmt for asdf entrypoints


jthegedus:
### Describe the Bug

#1391 updated the `scripts/*.bash` to add a specific Shellcheck & shfmt commands to validate our `asdf.sh` and `lib/*.sh` files. This code was previously part of a single check, but was probably misclassifying the files as Bash, when they're seemingly intended to be POSIX.

The updated checking was introduced but disabled. We should:

- determine if we want `asdf.sh` or `lib/*.sh` to be POSIX or Bash
- enable these checks
- fix the results

### Steps to Reproduce

Uncomment the code in https://github.com/asdf-vm/asdf/blob/master/scripts/shellcheck.bash

```diff
# check .sh files
-# TODO(jthegedus): unlock this check later
-# TODO  shellcheck --shell sh --external-sources \
-# TODO  asdf.sh \
-# TODO  lib/*.sh
+shellcheck --shell sh --external-sources \
+  asdf.sh \
+  lib/*.sh
```

Uncomment the code in https://github.com/asdf-vm/asdf/blob/master/scripts/shfmt.bash

```diff
# check .sh files
-# TODO(jthegedus): unlock this check later
-# TODO  shfmt --language-dialect posix --indent 2 --diff \
-# TODO  asdf.sh \
-# TODO  lib/*.sh
+shfmt --language-dialect posix --indent 2 --diff \
+  asdf.sh \
+  lib/*.sh
```

Uncomment the code in https://github.com/asdf-vm/asdf/blob/master/scripts/format.bash

```diff
# check .sh files
-# TODO(jthegedus): unlock this check later
-# TODO  shfmt --language-dialect posix --indent 2 --write\
-# TODO  asdf.sh \
-# TODO  lib/*.sh
+shfmt --language-dialect posix --indent 2 --write\
+  asdf.sh \
+  lib/*.sh
```

### Expected Behaviour

CI will run this script and reveal a bunch of issues to be fixed.

### Actual Behaviour

These are currently skipped.

### Environment

```shell
╭─jthegedus@art at 20:58:41
├ ~/dev/asdf
├  master
├ Last command took 596ms 
╰─λ asdf info
OS:
Linux art-win11 5.15.79.1-microsoft-standard-WSL2 #1 SMP Wed Nov 23 01:01:46 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
fish, version 3.3.1

ASDF VERSION:
[...]

PR #1441, opened 2023-01-20
perf: speed improvements

stephanGarland:
# Summary

This is a draft for now. It pulls in the patch I [linked here](https://github.com/asdf-vm/asdf/issues/290#issuecomment-1383056355), as well as some other work. I mostly wanted to get people's opinion on it early before I continue with the entire project.

In general, it speeds up performance by reducing fork calls, either by combining various combinations of `sed | grep | cut` etc. into a single `awk` call, or where possible, using shell built-ins like parameter substitution and modification.

## Fixes

It _addresses_ but does not _fix_ [issue 290](https://github.com/asdf-vm/asdf/issues/290)

## Other Information

I haven't ran the full test suite yet, but I know at least one (`where`) fails; I believe this to be a false positive, and will address it.


lib/utils.bash
```diff
@@ -7,6 +7,45 @@ GREP_COLORS=
...
+        outputStr="${inputStr@U}"
+        ;;
+      "[:upper:]")
+        outputStr="${inputStr@L}"
+        ;;
+      "*")
+        printf "ERROR: fast_tr() expects [:lower:] and [:upper:] as arguments"
+    esac
+    # and then use parameter substitution to replace characters in the string
+    # if the 4th or 5th elements are not set, the substitution has no effect
```
stephanGarland, line 41: tbf this is only true because of the check on L24

lib/utils.bash
```diff
@@ -190,7 +229,8 @@ find_versions() {
   version=$(get_version_from_env "$plugin_name")
   if [ -n "$version" ]; then
     local upcase_name
-    upcase_name=$(printf "%s\n" "$plugin_name" | tr '[:lower:]-' '[:upper:]_')
+    tr_args=("$plugin_name" "[:lower:]" "[:upper:]" "-" "_")
+    upcase_name=$(fast_tr ${tr_args[@]})
```
stephanGarland, line 233: Due to the subshell this is at best equally as fast as before. Probably a way to do this without a subshell.

lib/utils.bash
```diff
@@ -647,11 +691,11 @@ shim_plugin_versions() {
 
 shim_plugins() {
   local executable_name
-  executable_name=$(basename "$1")
+  executable_name=$(fast_basename "$1")
   local shim_path
   shim_path="$(asdf_data_dir)/shims/${executable_name}"
   if [ -x "$shim_path" ]; then
-    grep "# asdf-plugin: " "$shim_path" 2>/dev/null | sed -e "s/# asdf-plugin: //" | cut -d' ' -f 1 | uniq
+    awk '/# asdf-plugin: / { l=$(NF-1) } END { print l} ' "$shim_path" 2>/dev/null
```
stephanGarland, line 698: > { print l}
ugh, need to fix the spacing

jthegedus:
@stephanGarland A few thoughts after a quick glance.

Before we introduce changes like these I would like to standardize and have in our CI pipelines performance testing and reporting to PRs. That way we identify the actual performance uplift of a specific change.

Once we have the above, we can then go through and add each of these changes one at a time. That way we should have clear data about the improvement for each case.

Further, #1433 & #1138 are partially addressing #1396 & #1397 which will lead us down the path towards using more native shell features. I am not sure if changes for those will clash with what you have here, but be prepared for some rework here.

This draft PR is a great place for us to discuss your performance improvements and ideas, so please keep this open, keep updating it, and I will hopefully have time to review the rest of it soon.

Thanks! This is exciting work!

hyperupcall:
Awesome! Performance increases are a much needed fix.

I know #1433 was mentioned, and as its author, I know that PR is pretty big, but I only fixed formatting (mostly), saving perf for later. So any merge conflicts [of those commits](https://github.com/asdf-vm/asdf/compare/master...hyperupcall:asdf:shellcheck-tests) should be minimal and I'm sure I'll have no problem rebasing over the changes here if this is merged first. And on second thought, you have nothing to worry about since those PRs will touch only files in `test/*`, which doesn't appear to be happening here.

I've identified some Bash incompatibilities, and some possible improvements. There might be some disagreement of some sort about the first suggestion, so maybe that one can be hashed out once the mentioned additional perf test is integrated in the CI pipeline

lib/utils.bash
```diff
@@ -7,6 +7,45 @@ GREP_COLORS=
 
 ASDF_DIR=${ASDF_DIR:-''}
 ASDF_DATA_DIR=${ASDF_DATA_DIR:-''}
+BASH_MAJOR_VERSION=${BASH_VERSION%%.*}
+
+fast_basename() {
```
hyperupcall, line 12: This can be changed to:?

```sh
local path="$1"
path=${path%/}
path=${path##*/}
printf '%s\n' "$REPLY"
```

The extra tests seem redundant as Bash should perform them internally.

I think it's _very_ worth pursuing having a `REPLY`-like pattern as shown [here](https://github.com/asdf-vm/asdf/blob/d81b81f9de2dc5961624464df04cef7cafae588c/lib/utils.bash#L852) to completely eliminate the subshell (I strongly prefer the variable being called `REPLY`, though). In my experience this gives gigantic perf wins, something similar as what it seems you have alluded to in your other comment. I'm not sure how on-board the core-team is with this, though.

lib/utils.bash
```diff
@@ -7,6 +7,45 @@ GREP_COLORS=
...
+  # if the path ends in a slash, remove it, then use parameter substitution
+  # if it doesn't, use parameter substitution on the path as-is
+  dirpath="$1"
+  [[ "$dirpath" == */ ]] && dirpath="${dirpath%/}" && dirpath="${dirpath##*/}" || dirpath="${dirpath##*/}"
+  printf "$dirpath"
```
hyperupcall, line 17: This, and other invocations to `printf` could have `%s` as a first parameter to prevent unintentional formatting

lib/utils.bash
```diff
@@ -7,6 +7,45 @@ GREP_COLORS=
...
+    return 1
+  fi
+  # if bash >= v4, use parameter modification to upper/lower the string
+  if [ ${BASH_MAJOR_VERSION} -gt 3 ]; then
+    case ${inputArr[1]} in
+      "[:lower:]")
+        outputStr="${inputStr@U}"
+        ;;
+      "[:upper:]")
+        outputStr="${inputStr@L}"
```
hyperupcall, line 35: This breaks since `@U` and `@L` were added in bash 5.0 while `,,` and `^^` were added in bash 4.0. But the check is still good because we support 3.2 at a minimum.

hyperupcall (changes requested):

lib/utils.bash
stephanGarland: Agreed that the additional subshells can be removed, ty for this recommendation.

Re: `$REPLY`, I thought that was specifically for [read](https://tldp.org/LDP/abs/html/internalvariables.html)? Using your code sample as-is resulted in an empty output for me.

lib/utils.bash
stephanGarland: Ah, good catch. I originally had the `^^` syntax but then found out about `@U` and used it since it seemed more intuitive. I didn't check the version requirements for it, though.

lib/utils.bash
hyperupcall: Yeah it is for `read` and also `select`, I'm borrowing the convention (more info [here](https://github.com/hyperupcall/awesome-basalt/blob/main/BEST_PRACTICES.md#functions)). It will work if the function looks something like this:

```sh
fast_basename() {
  unset -v REPLY; REPLY=

  local path="$1"
  path=${path%/}
  path=${path##*/}
  REPLY=$path
}
```
And is called like this:

```sh
fast_basename "$executable_path"
local cmd="$REPLY"
```

The code sample I provided was for existing subshell code, not the `REPLY`-thing.

lib/utils.bash
```diff
@@ -622,9 +626,7 @@ with_plugin_env() {
   local path exec_paths
   exec_paths="$(list_plugin_exec_paths "$plugin_name" "$full_version")"
 
-  # exec_paths contains a trailing newline which is converted to a colon, so no
-  # colon is needed between the subshell and the PATH variable in this string
-  path="$(tr '\n' ':' <<<"$exec_paths")$PATH"
+  path="${exec_paths}:$PATH"
```
stephanGarland, line 630: I didn't find the comment to be true in testing, and this works fine as-is. Open to proof to the contrary.

lib/utils.bash
```diff
@@ -892,7 +888,9 @@ remove_path_from_path() {
   # Output is a new string suitable for assignment to PATH
   local PATH=$1
   local path=$2
-  substitute "$PATH" "$path" "" | sed -e "s|::|:|g"
+  PATH=${PATH//"$path"}
```
stephanGarland, line 891: There is an argument to be made for using the substitute call everywhere to make it more easily understandable, but at the very least, here the sed pipe is unnecessary and this can simply be modified twice.

lib/utils.bash
```diff
@@ -365,8 +365,9 @@ parse_asdf_version_file() {
 
   if [ -f "$file_path" ]; then
     local version
-    # this is actually only about 5% faster than the original, despite losing a function call and two pipes
+    # this is causing multiple failures in tests/utils.bash, despite seeming to work identically
```
stephanGarland, line 368: Need to figure out why this works fine for actual tests (including paths with spaces in them), but fails in bats.

lib/utils.bash
hyperupcall: For perf test, you have to include the capture of the variable as a part of the test. These were my results:


<details>

<summary>script.sh</summary>

```sh
#!/usr/bin/env bash

fast_basename_work() {
	local path="$1"
	path=${path%/}
	printf '%s\n' "{path##*/}"
}

slow_basename() {
	unset -v REPLY
	REPLY=$(basename "$1")
}

fast_basename() {
	unset -v REPLY
	REPLY=$(fast_basename_work "$1")
}

faster_basename() {
	unset -v REPLY; REPLY=
	local path="$1"
	path=${path%/}
	path=${path##*/}
	REPLY=$path
}

main() {
	"$1" "$PWD"
	local _result="$REPLY"
}

main "$@"
```



</details>



```sh
> $ for fn in slow_basename fast_basename faster_basename; do hyperfine -N --warmup 1000 --runs 1000 "./script.sh $fn"; done
Benchmark 1: ./script.sh slow_basename
  Time (mean ± σ):       3.6 ms ±   0.2 ms    [User: 2.0 ms, System: 1.4 ms]
  Range (min … max):     3.4 ms …   8.8 ms    1000 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 1: ./script.sh fast_basename
  Time (mean ± σ):       3.2 ms ±   0.4 ms    [User: 1.6 ms, System: 1.4 ms]
  Range (min … max):     2.9 ms …  15.1 ms    1000 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 1: ./script.sh faster_basename
  Time (mean ± σ):       2.7 ms ±   0.3 ms    [User: 1.3 ms, System: 1.2 ms]
  Range (min … max):     2.5 ms …   8.9 ms    1000 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```


For your questions about my self-defined "best practices"

> If you unset a var (or function), it's gone. In C you may need to then initialize the empty var to a known value, but this isn't C.

The `REPLY=` is to be absolutely certain that `REPLY` is what you think it is. For a single function, this doesn't do anything, but if you have hundreds of functions that set `REPLY`, this ensures that artifacts of running those other functions don't accidentally interfere with the current function.

The `unset -v` is necessary because we want to be absolutely sure that the variable is of the correct type - it's fundamentally more correct. See below:

<details>

<summary>script.sh</summary>

```sh
# shellcheck shell=bash

fn1() {
	unset -v REPLY
	REPLY=

	printf 'fn1: %s\n' "$(declare -p REPLY)"
	REPLY=(a b)
}

fn2() {
	# unset -v REPLY
	REPLY=

	printf 'fn2: %s\n' "$(declare -p REPLY)"
	REPLY=c
}

main() {
	fn1
	local -a result1="${REPLY[@]}"

	fn2
	local result2="$REPLY"
}

main "$@"
```



</details>

And I can't just do `unset -v REPLY` because for some reason, people like to enable `errunset`, and since I'm usually writing Bash libraries, I need to be compatible for the options that they wish to set.

> If it's function-scoped, why bother unsetting it? This is always going to print a null for the first echo, and 10 for the second:

Yeah, it's function scoped, and I still don't want that variable to leak into the callers of functions that are called later in that function like so:

```sh
otherfn() {
  local i=
  for i in {0..1}; do :; done

  foo
}
```

So by doing `unset -v i`, the goal is to have a behavior closer to scoped variables (i suppose like `let` vs `var` in JavaScript)

stephanGarland: > So I'm not sure what you're trying to measure when you use printf without a subshell.

In the light of day, I also have no idea what I was doing there. Good catch.

> And I can't just do unset -v REPLY because for some reason, people like to enable errunset, and since I'm usually writing Bash libraries, I need to be compatible for the options that they wish to set.

An excellent point I hadn't thought of. People love to set `-euo pipefail` because someone said it was better.

> Yeah, it's function scoped, and I still don't want that variable to leak into the callers of functions that are called later in that function like so

Again, an excellent point I hadn't thought of.

I think I'm just going to defer to your recommendations. I thought I knew shell reasonably well, but clearly I've a lot more to learn. Thank you for taking the time to demonstrate your answers with scripts.

Stratus3D:
> Before we introduce changes like these I would like to standardize and have in our CI pipelines performance testing and reporting to PRs.

For the record I agree with @jthegedus on this. I'd love to improve performance but I also don't want to be reviewing changes that only improve performance by a small percentage, or changes that only improve things for a small subset of users (not saying that is the case here, I'm just now looking at these changes).

Stratus3D (commented):
The new functions appear ok, but whether or not these changes will be accepted depends on how these changes improve the performance of asdf commands. The performance of these particular functions on their own is less important than their affect on the performance of the whole codebase.

jthegedus:
Once I finish my other PRs I will work on the performance automation for this repo.

stephanGarland:
I found and fixed a bug I had introduced with `parse_asdf_version_file` that was causing a ton of failures. There are still test failures (34/352, but tbf some of those are from fish and elvish simply because I don't have them installed yet), but the number is greatly reduced.

lib/utils.bash
stephanGarland: Done. A combination of only printing `$NF` with awk, and then having leading whitespace.

ssbarnea:
Happy 1 year anniversary for this PR! Sic.

stephanGarland:
> Happy 1 year anniversary for this PR! Sic.

I got frustrated trying to fix bugs only present in esoteric shells like elvish, and around the same time, shifted to mise (née rtx). Sorry. I might pick it up again once my professional life has calmed down some.

Stratus3D:
Hi @stephanGarland , sorry for the late reply here. I've spent the last several months rewriting asdf in Golang (#1805). The rewrite should solve most of our performance issues. It's 2x-15x faster than the Bash implementation on all operations I tested. I will be doing more thorough benchmarking of asdf soon (similar to what I've done in the past (http://stratus3d.com/blog/2022/08/11/asdf-performance/). Grab binaries from the latest dev build if you want to try the Go implementation (https://github.com/asdf-vm/asdf/releases/tag/54d15e3)

I'm going to close this PR now as it's not needed. Thank you for taking the initiative to work on this, I wish I had had the time to review it back when you created it.

ssbarnea:
@Stratus3D Interesting to read about this, I hope it progresses well. About a month ago I replaced asdf with https://mise.jdx.dev/ but there was one thing that I did not like about mise, that us has just one person behind, and that is a high risk long term. At least asdf is under and org and apparently has more contributions from others,... so it does mitigate this risk to some extent.

Stratus3D:
Yeah nothing wrong with mise! Definitely very interesting project with more features than asdf. I still think asdf's shims are a better design for my purposes. It's been a struggle trying to implement them correctly in Bash, but with Go it was much easier to implement. The rewrite revealed several bugs which I fixed in the Go code. Shims incur a cost when invoking a command (which can occur multiple times) where Mise's path method incurs a cost at shell startup (which in theory occurs less frequently).

Mise is a fine choice, and hopefully with recent improvements asdf will be just as good.

hyperupcall:
For the record, a (very) while back I was testing out performance enhances on a [personal branch](https://github.com/fox-forks/asdf/tree/hyperupcall-patch-perf), which improved performance for "quick commands" like `asdf which` and `asdf where` by ~30% (negligible difference for "long commands" like `asdf list` etc.

perf

It mostly consisted of improving various small inefficiencies using regular expressions. I'm only posting about this now because I only just got access to the disk drive. Of course, this is all moot with the Go rewrite. But thought I'd share anyways

PR #1443, opened 2023-01-22

hyperupcall:
# Summary

This PR is the **second** of a multi-part series that fixes various issues with the tests. For more detailed information, see the [first](https://github.com/asdf-vm/asdf/pull/1433) PR.

Here is a description of each commit/part:

- ~~Parts 1 does a few miscellaneous fixes~~ (#1433)
- Parts 2-5 are quite trivial and fix up obvious mistakes like quoting.
- Part 6 has the first non-obvious changes that remove use of `$?` by using the `run` function provided by Bats 
- Part 7 fixes all errors and warnings shown by Bats. The significant code change is that some regex comparison were changed to glob comparisons.
- Part 8 enables ShellCheck for warnings and errors
- Part 9 (the last commit) Fixes all infos, and enables ShellCheck to lint infos as well

Partially addresses #1396

### DIFF
--- a/test/get_asdf_config_value.bats
+++ b/test/get_asdf_config_value.bats
@@ -31,12 +31,12 @@ teardown() {
 }
 
 @test "get_config returns default value when the key does not exist" {
-  [ $(get_asdf_config_value "key2") = "value2" ]
+  [ "$(get_asdf_config_value "key2")" = "value2" ]
 }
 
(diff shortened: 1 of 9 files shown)


jthegedus (approved):
LGTM

Stratus3D:
LGTM as well, thanks for these @hyperupcall !

PR #1446, opened 2023-01-22
ci: lint github actions with actionlint

jthegedus:
# Summary

Use https://github.com/rhysd/actionlint to lint our GitHub Actions workflow files


jthegedus:
It is possible for us to add this to `.tool-versions` and manage with asdf itself, but we had clashes with adding `python` here, so although I didn't test it, am hesitant to. This is a relatively minor change we can iterate on if we so wish later.

Stratus3D:
Looks great! Thanks for adding @jthegedus !

PR #1447, opened 2023-01-22
ci: pin python version for checkstyle script

jthegedus:
# Summary

We had some issues with using our repo root `.tool-versions` to pin our python version used for `scripts/checkstyle.py`. Currently the asdf action in the `lint.yml` workflow is not installing the python version specified in `scripts/.tool-versions` because it is being run from the repo root, not `scripts/`. You cannot use the `working-directory` workflow option with `uses` steps, only `run` steps, so we cannot resolve this easily (it should be something we fix in the actual asdf action itself (I will do that soon[TM])).

Here we are pinning a minimum python version as defined in #1409 to install and run our `scripts/checkstyle.py`.

## Other Information

The aforementioned issues being that the plugins install below this directory, so when a Node.js package, executed Python, the user wasn't expecting Python to have been resolved to the asdf repo version, but rather their own global.


jthegedus:
Had to change the python version from `3.6.15` (min our checkstyle script supports) to `3.7.13` as it is the minimum the `ubuntu-22.04` (latest) runner supports 🙄

Stratus3D:
Thanks for taking care of this!

PR #1449, opened 2023-01-22

hyperupcall:
# Summary

This PR is the **fourth** of a multi-part series that fixes various issues with the tests. For more detailed information, see the [first](https://github.com/asdf-vm/asdf/pull/1433) PR.

Partially addresses #1396

### DIFF
--- a/test/asdf_elvish.bats
+++ b/test/asdf_elvish.bats
@@ -18,7 +18,7 @@ setup() {
 }
 
 cleaned_path() {
-  echo $PATH | tr ':' '\n' | grep -v "asdf" | tr '\n' ' '
+  echo "$PATH" | tr ':' '\n' | grep -v "asdf" | tr '\n' ' '
 }
 
 @test "exports ASDF_DIR" {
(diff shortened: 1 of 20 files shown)


test/test_helpers.bash
```diff
@@ -22,7 +22,7 @@ setup_asdf_dir() {
   ASDF_BIN="$(dirname "$BATS_TEST_DIRNAME")/bin"
 
   # shellcheck disable=SC2031
-  PATH="$ASDF_BIN:$ASDF_DIR/shims:$PATH"
```
jthegedus, line 25: This doesn't seem right to me. What is wrong with the single set of quotes around the outside here. Is this perhaps an issue because we're scanning this file as a `.bats` file when it is a `.bash` file?

hyperupcall: omg good catch. I think this one was just an accidental application of a find-and-replace. And Bats doesn't show anything since word splitting doesn't occur here

jthegedus: thanks

jthegedus (approved):

jthegedus:
Thanks!

PR #1452, opened 2023-01-25

hyperupcall:
# Summary

This PR is the **seventh** of a multi-part series that fixes various issues with the tests. For more detailed information, see the [first](https://github.com/asdf-vm/asdf/pull/1433) PR.

Partially addresses #1396

### DIFF
--- a/scripts/shellcheck.bash
+++ b/scripts/shellcheck.bash
@@ -22,7 +22,5 @@ shellcheck --shell bash --external-sources \
   test/fixtures/dummy_legacy_plugin/bin/* \
   test/fixtures/dummy_plugin/bin/*
 
-# check .bats files
-# TODO(jthegedus): unlock this check later
-# TODO  shellcheck --shell bats --external-sources \
-# TODO  test/*.bats
+shellcheck --shell bats --external-source \
+  --severity warning test/*.bats
--- a/test/shim_env_command.bats
+++ b/test/shim_env_command.bats
@@ -1,4 +1,5 @@
 #!/usr/bin/env bats
+# shellcheck disable=SC2164
 
 load test_helpers
 
@@ -57,7 +58,7 @@ teardown() {
   run asdf env dummy
   [ "$status" -eq 0 ]
 
-  run grep 'FOO=bar' <(echo "$output")
+  run grep 'FOO=bar' <<<"$output"
   [ "$output" = "" ]
   [ "$status" -eq 1 ]
 
@@ -78,6 +79,6 @@ teardown() {
   [ "$path_line" != "" ]
 
   # Should not contain duplicate colon
-  run grep '::' <(echo "$path_line")
-  [ "$duplicate_colon" = "" ]
+  run grep -q '::' <<<"$path_line"
+  [ "$status" -ne 0 ]
 }
--- a/test/asdf_fish.bats
+++ b/test/asdf_fish.bats
@@ -1,4 +1,5 @@
 #!/usr/bin/env bats
+# shellcheck disable=SC2164
 
 load test_helpers
 
--- a/test/asdf_nu.bats
+++ b/test/asdf_nu.bats
@@ -1,4 +1,5 @@
 #!/usr/bin/env bats
+# shellcheck disable=SC2164
 
 load test_helpers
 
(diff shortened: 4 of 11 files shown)


scripts/shellcheck.bash
```diff
@@ -22,7 +22,7 @@ shellcheck --shell bash --external-sources \
   test/fixtures/dummy_legacy_plugin/bin/* \
   test/fixtures/dummy_plugin/bin/*
 
-# check .bats files
-# TODO(jthegedus): unlock this check later
-# TODO  shellcheck --shell bats --external-sources \
-# TODO  test/*.bats
```
jthegedus, line 28: The point of #1396 & #1397 was that the absence of the `--shell` flag gave a false sense of security. People thought files were being tested that were not.

I would like to keep `--shell bats`. The `--shell bash` files in the `test/` dir are covered in the above Shellcheck execution with that shell flag set.

Why exclude `SC2164`?

Why change severity?

test/shim_env_command.bats
```diff
@@ -79,5 +79,4 @@ teardown() {
 
   # Should not contain duplicate colon
   run grep '::' <(echo "$path_line")
-  [ "$duplicate_colon" = "" ]
```
jthegedus, line 82: Why is the assertion removed? explicit assertion is preferred over implicit.

hyperupcall: The variable `duplicate_colon` isn't set anywhere, and the original [commit](https://github.com/asdf-vm/asdf/commit/549d86e7339110ac4095671c71b9ad3f14b8dd5a) that created this test didn't set it, so I thought it was a mistake

scripts/shellcheck.bash
hyperupcall: I'm not sure I see your understand your first point. I read on the shellcheck man page that:

> The default [if --shell is not specified] is to deduce the shell from the file's shell directive, shebang, or .bash/.bats/.dash/.ksh extension, in that order.  sh refers to POSIX sh (not the system's), and will warn of portability issues.

Since the files extensions of each test in `test/*` match how it should be linted, I'm not sure what the "false sense of security is" - the way I understood it, the only issue are files that end with `.sh` - since that is potentially ambiguous semantically (without a shebang or `#shellcheck shell=bash` entry at the top). In my opinion, `shellcheck.bash` doesn't make all that much sense to me - if we used the command `shellcheck --external-sources ./**/*.{sh,bash,bats}` from the start, then we wouldn't be in the predicament of the `test/` directory not being linted in the first place. With how we currently have it, the checking against the type of shell can potentially differ from the filename extension, and we can potentially skip files if the directory structure changes and we forget to update the command. I was going to make a PR for this later, but I guess it's good we can discuss this now.

hyperupcall: I disabled `SC2164` because it was not relevant since Bats automatically applies `-e` mode, and ShellCheck doesn't know that - I'll add a comment for each rule I ignore.

The severity was changed temporarily, just to show that "all warnings pass" - the [next pr](#1453) properly removes it, so everything is tested

test/shim_env_command.bats
jthegedus: Ah okay, well, the result from the `run` command should be checked to ensure that the `::` is not found by grep.

Can you update?

scripts/shellcheck.bash
jthegedus: @hyperupcall the script is the way it is for historical reasons. I would like to keep the definition of files and which shell we test against explicit for now, 1 for execution of shellcheck for each type of `--shell`. We can revisit later once we decide on `sh` vs `bash` as outlined in #1396

I would prefer not to ignore SC rules across the whole execution from the command, but rather use comments and ignore per-occurrence and then per-file.

test/shim_env_command.bats
hyperupcall: Definitely - Gosh I'm kind of embarrassed that I didn't check to replace the test

scripts/shellcheck.bash
jthegedus: With the check at the top of each file we can in future use the single-line shellcheck you proposed above without the exclude leaking into other files. Thanks

test/shim_env_command.bats
jthegedus: The tests have been missing this for a long, you're just the one discovering it, not your fault!

jthegedus (approved):
Thanks!

PR #1457, opened 2023-01-25
chore: improve checkstyle double equals test regex & types

hyperupcall:
# Summary

General fixes for `checkstyle.py`, including a false positive fix for the `no-test-double-equals` rule.

### DIFF
--- a/scripts/checkstyle.py
+++ b/scripts/checkstyle.py
@@ -17,6 +17,11 @@ from typing import List, Dict, Any # compat
 #
 # Lint a particular file
 # $ ./scripts/checkstyle.py ./lib/functions/installs.bash
+#
+# Check to ensure all regexes are working as intended
+# $ ./scripts/checkstyle.py --internal-test-regex
+
+Rule = Dict[str, Any]
 
 class c:
     RED = '\033[91m'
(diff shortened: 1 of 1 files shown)


jthegedus:
If this change fixes a false positive, why does the lint script not show so in the PR checks?

eankeen:
@jthegedus If I am understanding you correctly, the PR checks don't change because no existing code in `main` is matched by the old regular expression. Code from #1456 does match though (in particular, the `[ ${line[0]} == 'usage: '* ]`).

hyperupcall:
@jthegedus If I am understanding you correctly, the PR checks don't change because no existing code in main is matched by the old regular expression. Code from https://github.com/asdf-vm/asdf/pull/1456 does match though (in particular, the [ ${line[0]} == 'usage: '* ]).

jthegedus:
I thought the Workflows executed from the forked code, not what was on the target branch. https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

Which would lead me to think that this change hasn't matched any previous false positives.

hyperupcall:
@jthegedus I don't think we are in disagreement - this change doesn't match any _previous_ false positives, but it will address potential _future_ potential ones ones. Specifically, the false positives are shown in the CI errors of #1456, which yes are executed from the forked code. I think we agree that these changes will fix false positives?

Maybe it's weird that I introduced this fix separately instead of bundling it together with #1456? I'm not sure

Stratus3D:
@hyperupcall I think I understand. As you say the false positive wasn't present in the codebase (since the wrong regex in this `checkstyle.py` script would have prevented it from ever being added).

@jthegedus you can validate this by reverting the change to line 125/old line 118 in `checkstyle.py` and running the file's tests like this: `./scripts/checkstyle.py --internal-test-regex`. It'll print this:

```
no-test-double-equals: Failed negative test:
=> [[ "${lines[0]}" == 'usage: '* ]]
```

Stratus3D (approved):
Code seems fine but it's not clear to me why this check is needed at all.

The description states: `Although it is valid Bash, it reduces consistency and copy-paste-ability`. Is that all there is too it?

hyperupcall:
The check was originally added to help address #1313. Maybe the error message should change to be more descriptive?

jthegedus:
Much better error msg, thanks!

PR #1459, opened 2023-01-26

### LINKED ISSUE #1396: bug: enable shellcheck on bats tests

hyperupcall:
# Summary

This PR is the **eigth** of a multi-part series that fixes various issues with the tests. For more detailed information, see the [first](https://github.com/asdf-vm/asdf/pull/1433) PR.

I disabled SC2164 in places because it was not relevant since Bats automatically applies -e mode, and ShellCheck doesn't know that - I will update this PR to add a comment for each rule I ignore so it's more clear

I disabled `SC2030` and `SC2031` in places because it was not relevant since the tests are self contained (each test is within a subshell, and tests don't access variables from other tests)

I disabled [`SC2012`](https://www.shellcheck.net/wiki/SC2012) to keep the test behavior exactly the same. For each directory, only alphanumeric filenames were used so this warning doesn't really matter as much.

Closes #1396

### DIFF
--- a/scripts/shellcheck.bash
+++ b/scripts/shellcheck.bash
@@ -23,4 +23,4 @@ shellcheck --shell bash --external-sources \
   test/fixtures/dummy_plugin/bin/*
 
 shellcheck --shell bats --external-source \
-  --severity warning test/*.bats
+  test/*.bats
--- a/test/asdf_elvish.bats
+++ b/test/asdf_elvish.bats
@@ -1,4 +1,5 @@
 #!/usr/bin/env bats
+# shellcheck disable=SC2030,SC2031
 
 load test_helpers
 
--- a/test/install_command.bats
+++ b/test/install_command.bats
@@ -98,7 +98,7 @@ teardown() {
   run grep "# asdf-plugin: dummy 1.0.0"$'\n'"# asdf-plugin: dummy 1.1.0" "$ASDF_DIR/shims/dummy"
   [ "$status" -eq 0 ]
 
-  lines_count=$(grep "asdf-plugin: dummy 1.1.0" "$ASDF_DIR/shims/dummy" | wc -l)
+  lines_count=$(grep -c "asdf-plugin: dummy 1.1.0" "$ASDF_DIR/shims/dummy")
   [ "$lines_count" -eq "1" ]
 }
 
--- a/test/latest_command.bats
+++ b/test/latest_command.bats
@@ -17,19 +17,19 @@ teardown() {
 ####################################################
 @test "[latest_command - dummy_plugin] shows latest stable version" {
   run asdf latest dummy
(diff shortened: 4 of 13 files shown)


jthegedus (approved):
Looks good to me. Shellcheck suggestions like replacing `ls` with `find` might be something we want to address in future, but happy to leave for later investigation.

jthegedus:
Same with SC2030/SC2031, while not necessarily applicable to the unique way Bats executes each test in a subshell, addressing the cause means that if that behaviour ever changes, we're not masking issues that might actually matter.

Stratus3D:
> Looks good to me. Shellcheck suggestions like replacing `ls` with `find` might be something we want to address in future, but happy to leave for later investigation.

I think we should address that when we can.

Issue #1435, opened 2023-01-18
docs: track full list of asdf dependencies


jthegedus:
### How can we improve the documentation?

It would be ideal to improve our list of dependencies and also capture the minimum supported versions.

Prior discussion #511

---

## Dependencies list

Here is a non-comprehensive list of dependencies from a quick look through some code:

| tool      | min version | what makes that the min version | why/where do we use it? |
| :-------- | :---------- | :------------------------------ | :---------------- |
| `git`     | `1.7.7.2`   | `git --work-tree`               | used in `asdf plugin update` with `--git-dir` to avoid `cd`. `git -C` is preferred but would require a later `git` version |
| `git`     | `1.8.5.6`   | `git -C`                        | used in **tests** in place of `--git-dir`+`--work-tree` |
| `curl`    |             |                                 | used in `plugin-test` command |
| `awk`     |             |                                 |                   |
| `grep`    |             |                                 |                   |
| `sed`     |             |                                 |                   |
| `tr`      |             |                                 |                   |
| `tail`    |             |                                 |                   |
| `head`    |             |                                 |                   |
| `cat`     |             |                                 |                   |

I am not sure if these tools are in the same class as the above. Should they be treated the same?

| tool      | min version | what makes that the min version | why/where do we use it? |
| :-------- | :---------- | :------------------------------ | :---------------- |
| `read`    |             |                                 |                   |
| `printf`  |             |                                 |                   |
| `rm`      |             |                                 |                   |
| `command` |             | `command -v`                    |                   |

`Bash` is almost certainly considered a dependency, but we will tackle that in another Issue.

### Thoughts

Is it even possible to list all min versions of tools which are different implementations of the same tool for different OSs? Eg: BSD vs GNU grep. Are the versions tracked the same? One assumes not as they could differ in minor/patch releases due to non-breaking changes and bugfixes, right? 🤔 

## Banned Commands

We have already banned some commands, or even just flags on commands because they are not common across most platforms. See - https://github.com/asdf-vm/asdf/blob/master/test/banned_commands.bats

## Potential reduction

I am sure we can reduce this list.

`curl` is used in `plugin-test` to fetch the Rate Limit dynamically to an error message. We can probably get rid of it as the users who see this branch of code is definitely a single digit number.

## Why raise this issue

* it would be good to know
* we already ban certain tools & flags which is some form of tracking
* we don't list the actual dependencies in our documentation site. `curl` is listed but technically not used by asdf core
[...]

PR #1460, opened 2023-01-27
test: vaidate Bats minimum required version in tests

hyperupcall:
# Summary

[According to Bats](https://github.com/bats-core/bats-core/releases/tag/v1.7.0), the just-merged usage of `setup_suite.bash` was added in Bats 1.7. Conveniently, a `bats_require_minimum_version` was also added to assert the minimum Bats version.

This uses that, partially addressing #1435  

The assertion must be in `test_helpers.bash`, because versions under 1.7 won't even automatically source `setup_suite.bash`, making the assertion useless

### DIFF
--- a/test/test_helpers.bash
+++ b/test/test_helpers.bash
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+bats_require_minimum_version 1.7.0
+
 # shellcheck source=lib/utils.bash
 . "$(dirname "$BATS_TEST_DIRNAME")"/lib/utils.bash
 


jthegedus:
I did read about this, and in relation to #1435 have been thinking if we should validate min versions not just in docs but as assertions in the code itself.

This specific update would only be useful for those who aren't using `asdf` to manage their deps in this codebase, as our CI uses the same version of Bats as defined in `.tool-versions`.

jthegedus (approved):
In general, I approve this idea, though it is only useful if:
* people are aware we are using this Bats feature when adding new Bats API usage to our tests
* people are not using `asdf` to manage their development tools for this repo (as defined in `.tool-versions`)

Given this is low impact I will approve and merge as a step towards #1435

PR #1463, opened 2023-01-29
test: Skip Fish test if not installed

hyperupcall:
# Summary

In the same style as #1423 and #1422, this only runs Fish tests if Fish is installed

Partially addresses #465 (part of the WSL errors are Fish not installed errors, even if the error isn't WSL-specific)

### DIFF
--- a/test/asdf_fish.bats
+++ b/test/asdf_fish.bats
@@ -5,6 +5,10 @@ load test_helpers
 
 setup() {
   cd "$(dirname "$BATS_TEST_DIRNAME")"
+
+  if ! command -v fish; then
+    skip "Fish is not installed"
+  fi
 }
 
 cleaned_path() {


jthegedus (approved):
Approving because we have this with the other types of Shells, however, I don't think this contributes to the linked WSL issues because we aren't actually running test cases in a WSL environment (something I do want to address once WSL2 can be run properly in GitHub CI).

Thanks!

jthegedus:
One of my concerns with code like this is it can potentially mask other issues. For instance, suppose that our CI env changes and doesn't properly install Fish, how will we be notified of this error. 

Perhaps this check itself can be wrapped in a way that they aren't run during CI environments.

Another PR that ensures these tests are only skipped in non-CI environments would be ideal.

hyperupcall:
That is a good point - a PR to ensure that the test if always ran in CI would prevent masking errors

Stratus3D:
> One of my concerns with code like this is it can potentially mask other issues. For instance, suppose that our CI env changes and doesn't properly install Fish, how will we be notified of this error.

Yeah, that's a really good point. I'd prefer explicit `DISABLE_FISH` flags for envs that don't have Fish available

PR #1464, opened 2023-01-29
fix: enforce & use consistent function definitions

hyperupcall:
# Summary

In Bash, there are there ways to declare a function:

```sh
fn() { :; }
function fn { :; }
function fn() { :; }
```

This PR ensures that only the first form is used.

### DIFF
--- a/scripts/checkstyle.py
+++ b/scripts/checkstyle.py
@@ -3,7 +3,7 @@ import re
 import os
 import argparse
 from pathlib import Path
-from typing import List, Dict, Any # compat
+from typing import Callable, List, Dict, Any # compat
 
 # This file checks Bash and Shell scripts for violations not found with
 # shellcheck or existing methods. You can use it in several ways:
@@ -33,6 +33,7 @@ class c:
     RESET = '\033[0m'
     BOLD = '\033[1m'
     UNDERLINE = '\033[4m'
+    LINK: Callable[[str, str], str] = lambda href, text: f'\033]8;;{href}\a{text}\033]8;;\a'
 
 def utilGetStrs(line, m):
     return (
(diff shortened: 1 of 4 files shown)


scripts/checkstyle.py
```diff
@@ -140,6 +157,20 @@ def main():
             ],
             'found': 0
         },
+        {
+            'name': 'no-function-keyword',
+            'regex': '^[ \\t]*(?P<match>function .*?(?:\\([ \\t]*\\))?[ \\t]*){',
+            'reason': 'Only allow functions declared like `fn_name() { :; }` for consistency',
```
jthegedus, line 163: Consistency, sure, but a better reasoning is because it is more portable. `function` is not as portable. Shellcheck has a rule for this https://www.shellcheck.net/wiki/SC2113 It is just not showing as an error because it only applies to `--shell sh` not `--shell bash` or `--shell bats`.

jthegedus:
How is this deciding which files to apply to? `lib/asdf.fish` also has a `function ` usage. Why is that not matched?

scripts/checkstyle.py
hyperupcall: Yeah, `function` isn't portable, but since checkstyle only scans Bash/Bats files, I guess I thought it would be better to put styling as a reason. I will change it to add a reference to the SC error too

hyperupcall:
> How is this deciding which files to apply to? lib/asdf.fish also has a function  usage. Why is that not matched?

Only `bash` and `sh`, and `bats` files are matched, if files are directly supplied to `checkstyle.py`. Since there are relatively few lines of Fish, I opted not to do any checking of them.

scripts/checkstyle.py
jthegedus: Yes please. Consistency & portability (as pointed out by Shellcheck in SC2113)

jthegedus:
Renaming the PR to be a `fix:` type as it touches code executed by users, so therefore should be considered as a change in the next release version.

jthegedus (approved):
Thanks

Stratus3D:
Thanks @hyperupcall ! I've been wanting this for years!

PR #1465, opened 2023-01-29
chore: provide `.git-blame-ignore-revs` for easier investigation

hyperupcall:
# Summary

This adds a .git-blame-ignore-revs file so formatting commits show up less frequently, in [GitHub](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) or with hyper-blame (automatically), and with blame using --ignore-rev (configurable). It also works in VSCode with the configuration:

```json
"gitlens.advanced.blame.customArguments": [
  "--ignore-revs-file",
  ".git-blame-ignore-revs"
],
```

It adds the recent commits that have mass-fixed at least tens of files at once, including the first shfmt commit.. Unfortunately, since some of the "Fix ShellCheck errors in tests" were merged as a whole commit, some blames to `scripts/checkstyle.py` will be incorrect.


jthegedus:
Two questions:

* can this support file renames as well?
* can we add this VSCode setup guide to the contribution docs page? - https://asdf-vm.com/contribute/core.html#development

hyperupcall:
> can this support file renames as well?

Assuming Git is able to track the rename (iirc if you do `git mv`, then it _should_ work), then I wouldn't see why not - I just don't think this file is directly relevant to renaming files. And if Git doesn't properly track the rename, then it wouldn't matter, since no previous history would show up anyways.

> can we add this VSCode setup guide to the contribution docs page? - https://asdf-vm.com/contribute/core.html#development

:+1:

docs/contribute/core.md
```diff
@@ -64,6 +64,33 @@ The following is the `.gitignore` file in the `asdf-vm/asdf` repository. We igno
...
+If using an IDE, configure it to enhance the default experience. For example, when using VSCode (with GitLens) configure it like so:
+
+```json:no-line-numbers
+{
+  "gitlens.advanced.blame.customArguments": [
+    "--ignore-revs-file",
+    ".git-blame-ignore-revs"
+  ]
+}
+```
```
jthegedus, line 92: is this a file supposed to be in `.vscode/settings.json`? Can you say which file these settings should go in?

docs/contribute/core.md
```diff
@@ -64,6 +64,33 @@ The following is the `.gitignore` file in the `asdf-vm/asdf` repository. We igno
 
 @[code](../../.gitignore)
 
+### `.git-blame-ignore-revs`
+
+`asdf` uses a `.git-blame-ignore-revs` to reduce noise when running a blame.
```
jthegedus, line 69: After this line, can you link to the official Git documentation for this feature?

hyperupcall:
:+1:

jthegedus:
Thanks.

I ask about renames because there are some old commits which renamed all the files from

```
lib/commands/command-version
# to
lib/commands/command-version.sh
# to
lib/command/command-version.bash
```

Would be good to track those down and include in this if it can support.

jthegedus (approved):

PR #1470, opened 2023-02-03
docs: update format of example asdf commands

ericbn:
# Summary

Update output of `asdf current`. Also fix `asdf where` command example, as it will already output the install path of the current version installed if no specific version is given.


jthegedus (approved):
Thanks

jthegedus (approved):

Issue #1458, opened 2023-01-26
[nushell] support PATH is not a list asdf.nu


mrdev023:
### Describe the Bug

![image](https://user-images.githubusercontent.com/11292703/214783214-ba386605-c00a-4f48-b8ba-ff04a856e40e.png)

### Steps to Reproduce

1. Clone https://github.com/mrdev023/dotfiles_nushell
2. `cd dotfiles_nushell`
3. `./install.sh`

**WARNING: Everything is deleted before each update or add** 
```
FILE ADDED: 
	~/.zoxide.nu
	~/.starship.nu

FOLDER LINKED: 
	.../dotfiles/dotfiles_nushell/nushell -> ~/.config/nushell

FOLDER ADDED: 
        ~/.asdf
```

### Expected Behaviour

asdf command work.

### Actual Behaviour

asdf loading crash.

But, with `let-env PATH = ( $env.PATH | split row ';' )` added before error line fix the problem

### Environment

```shell
nushell -> 0.74.0
zoxide -> 0.9.0
starship -> 1.12.0
```

PR #1471, opened 2023-02-06
fix: nushell PATH conversion to list before filter

hut8:
# Summary

Nushell, at least in recent versions (I did not check ancient versions) needs to have the `$PATH` variable split on the separator character (`;` for Windows, otherwise `:`). Currently, asdf is broken with Nushell as a result.

Fixes: #1458 

## Other Information

None


jthegedus (approved):
Thanks. Sorry for the delay.

Issue #1440, opened 2023-01-19
bug: asdf auto-complete doesn't work for "latest"


akkadaya:
### Describe the Bug

When typing `late` and hitting the tab afterwards, it doesn't auto-complete.  
But, it works for `--all` like hitting the tab after typing `asdf latest --` 

### Steps to Reproduce

1. Type `asdf late`
2. Press the Tab key
3. It doesn't auto-complete

### Expected Behaviour

To auto-complete `latest` in `asdf latest`

### Actual Behaviour

Doesn't auto-complete `latest` in `asdf late`

### Environment

```shell
OS:
Darwin DCS-N6D0QDCV4R-akkad
xxx.xxx 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:03:51 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.8.1 (x86_64-apple-darwin22.0)

ASDF VERSION:
v0.11.1-9a8bb47

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/Users/xxx.xxx/.asdf
ASDF_DIRENV_BIN=/Users/xxx.xxx/.asdf/installs/direnv/2.32.2/bin/direnv

ASDF INSTALLED PLUGINS:
direnv                       https://github.com/asdf-community/asdf-direnv.git master e3622e6
dotnet-core                  https://github.com/emersonsoares/asdf-dotnet-core.git master 4583e4b
java                         https://github.com/halcyon/asdf-java.git master db4cbc0
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master c9e5df4
perl                         https://github.com/ouest/asdf-perl.git master 4edc8ab
python                       https://github.com/danhper/asdf-python.git master 8505457
ruby                         https://github.com/asdf-vm/asdf-ruby.git master 4eb815c
rust                         https://github.com/code-lever/asdf-rust.git master 0c88f99
```

PR #1472, opened 2023-02-07
fix: bash completion for latest command

maristgeek:
# Summary

Adds _latest_ to the list of commands available for bash tab completion.

Fixes: #1440 

## Other Information

N/A


jthegedus (approved):
Thanks!

PR #1475, opened 2023-02-13
feat: Display system on "asdf list" if exist local system version each plugin

eugenejeonme:
# Summary

Check the local system version of each plugin, and if it exists, display the "system" keyword on the screen as shown below.
For example, macOS in Ventura 13.2 has Ruby 2.6.10 installed by default; Will check it and display it in the asdf list.

image
image

## Motivation

The local system version exists, but it is not displayed on the screen in the asdf list.
Sometimes there is confusion when you want to select the local system version.
So I decided to revise it.

## Other Information

It would be better if you could display the details in the asdf list.
However, only the system keyword is displayed for easy comparison with the currently selected version.

We need help from another hero. 😂

### DIFF
--- a/docs/plugins/create.md
+++ b/docs/plugins/create.md
@@ -168,6 +168,11 @@ See also the related hooks:
 - `post_asdf_plugin_remove`
 - `post_asdf_plugin_remove_${plugin_name}`
 
+#### bin/system-version
+
+This can be got the version of the tool installed on the system by asdf.
+If the system-version file isn't implemented, asdf will only find the file with the plugin name in the '/usr/bin/' folder.
+
 ## Extension commands for asdf CLI.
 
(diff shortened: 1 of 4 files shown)


jthegedus:
Thanks for the PR. This is an interesting idea.

Please update the tests for this command to ensure the new functionality is captured.

eugenejeonme:
> Thanks for the PR. This is an interesting idea.
> 
> Please update the tests for this command to ensure the new functionality is captured.

Oh, thanks. 
Do you mean to write the test code? I see. I'll write it. :)

eugenejeonme:
I have seen this issue, #1428 .
I found out that it was an issue related to "asdf list".
So, close this PR. To make a new feature (branch).

Issue #1467, opened 2023-02-02
No version manager name reported in CLI output, consider adding it


olivierlacan:
### Is your feature request related to a problem? Please describe

Hi there, excuse the snarky issue title, but it would be nice if asdf introduced itself in directories where it can't find `.tool-versions` instead of printing a cryptic message (which appears partially broken) without attribution. 

Otherwise this output message can be quite confusing: 

```
$ npm outdated
No version is set for command npm
Consider adding one of the following versions in your config file at
nodejs 16.13.2
```

Clearly there's some missing text here, probably assuming a location to a `.tool-versions` could be interpolated at the end of the second to last line? 

Formatting-wise, showing `nodejs 16.13.2` without any list formatting is an odd choice as well, since at different viewport width it appears to be the end of the sentence above.

### Describe the proposed solution

Change this message to this instead: 

```
$ npm outdated
asdf found no installed version for command `npm`.
You can add a version in `.tool-versions`, the following version is available: nodejs 16.13.2
```

That's of course assuming there's no alternative configuration file, but that's nothing a good link to a documentation page on the website won't fix if needed.

### Describe similar `asdf` features and why they are not sufficient

Since asdf is intercepting this call to `npm` I don't think this relates to other asdf features.

### Describe other workarounds you've considered

The only workaround would be for someone to be aware of `which npm` or `which -a npm` which sadly may be a bit too much to ask in some cases.

In my case, I forgot I installed `asdf` while also using `volta` and a Homebrew-installed version of node. I know, I know, that's a bit silly but I don't think it's that rare considering [this StackOverflow thread](https://stackoverflow.com/questions/74935373/no-version-is-set-for-command-npm-tool-versions-file-missing)

hyperupcall:
Yeah, I agree that it's pretty confusing. I've done some preliminary work in https://github.com/hyperupcall/asdf/commit/3a6d35b263de09aa5db2df5d119b8e4ed39f4f46 to address this

Before I make a PR, I'm still trying to figure out if prepending the terminal messages with `asdf: ` should be a thing enabled by default, or behind an environment variable. Sort of leaning towards the former, with a possible opt-out.

olivierlacan:
@hyperupcall I agree, disabling `asdf: <command>` prepending should definitely be an option, maybe that's something that could be enabled for any existing installs vs. net new ones to avoid breaking any stdout capture folks might have? 

Any way to reliably detect that? Maybe empty install directories?

[...]

Issue #928, opened 2021-04-27
Identify different class of error between **missing version** from **requiring a reshim**


jthegedus:
**Describe the bug**

We get a number of bug reports around the error `No preset version installed for command <CMD>` message. Some of the time the set version for the tool is not installed and this error message is correct. Other times it is not correct because the version is installed as required, but the plugin requires a reshim as a binary was installed using the tool and `asdf` needs to create a shim for the new binary.

We see this with any tool that can install binaries through it's own package manager (Ruby, Golang, Node.js etc)

The most recent example is https://github.com/asdf-vm/asdf/issues/914 is an example of this:

```
No preset version installed for command <golang binary>
Please install a version by running one of the following:

asdf install golang 1.16.2

or add one of the following versions in your config file at HOME/.tool-versions
golang 1.15.1
```

where the `<golang binary>` was installed and **used** before reshimming the plugin (`asdf reshim golang`).

**To Reproduce**

See #914 

**Expected behavior**

As mentioned by OP in https://github.com/asdf-vm/asdf/issues/914#issuecomment-827180264 the error message should notify the user to reshim when the specified version of the tool is installed.

Auto reshimming (#409) would resolve this class of errors. Notably, plugins can run the reshim themselves. The Node.js plugin uses the npm postinstall hook to run the reshim after installation (the Node package manager used must support the npm hooks though so this may be a brittle soln.) 

**Actual behavior**

We get an indication to install the tool instead of reshim.

### Environment

Copied from #914 

```shell
OS:
Darwin Kalorme 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64

SHELL:
zsh 5.8 (x86_64-apple-darwin20.0)

ASDF VERSION:
v0.8.0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/usr/local/opt/asdf

ASDF INSTALLED PLUGINS:
golang                       https://github.com/kennyp/asdf-golang.git
minio                        https://github.com/aeons/asdf-minio.git
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git
postgres                     https://github.com/smashedtoatoms/asdf-postgres.git
protoc                       https://github.com/paxosglobal/asdf-protoc.git
python                       https://github.com/danhper/asdf-python.git
redis                        https://github.com/smashedtoatoms/asdf-redis.git
[...]

PR #1479, opened 2023-02-18
fix: prefix messages with 'asdf: '

hyperupcall:
# Summary

Sometimes asdf prints messages, and it's not obvious that the message is coming from asdf. This fixes that.

It does so by prepending `asdf: ` to every single user-read console message, but only on the `asdf exec` subcommand codepath.

Fixes: #1467

### DIFF
--- a/bin/private/asdf-exec
+++ b/bin/private/asdf-exec
@@ -10,6 +10,6 @@ rm "$(asdf_data_dir)"/shims/*
 "$asdf_dir"/bin/asdf reshim
 shim_name=$(basename "$2")
 
-printf "asdf: now running %s\\n" "$shim_name"
+display "now running %s\\n" "$shim_name"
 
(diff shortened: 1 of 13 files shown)


jthegedus:
I am not sure this is the right solution to this problem. The initial report has a few things going on:

1. `.tool-versions` wasn't interpolated
2. the user found the error output format confusing
3. the user did not know which tool produced this error

I don't think prefixing all outputs with `asdf:` is the best solution here.

I do like the OPs suggestion for including 
>asdf did not find a version...

That is clear which tool produced the error.

And so perhaps this PR can be used to make that change.

That solves 3.

Solving 1 will take more investigation.

Solving 2 I think is related to https://github.com/asdf-vm/asdf/issues/928 more than anything.

hyperupcall:
Yeah you're right, that's a lot better - I'm not sure what I was thinking...

PR #1480, opened 2023-02-19
fix!: rework POSIX entrypoint for greater shell support

### LINKED ISSUE #653: Support for sh/dash
Is there an official list of what shells asdf supports? Is there any chance for dash/sh to be supported, or is it only bash and korn shells?

https://github.com/asdf-vm/asdf/blob/e79016add33720fbb039ec4e98c5959624704bbb/asdf.sh#L1-L4

Seems to be a partial list of support, but fish is also supported since there's an asdf.fish file.

Currently when run under sh you get an error like:
`/home/my_home/.asdf/asdf.sh: 8: /home/my_home/.asdf/asdf.sh: Bad substitution`

My use case is that I help maintain the [Elixir Language Server](https://github.com/elixir-lsp/elixir-ls/) which uses a `sh` script to launch the server, but since many users use asdf to manage their elixir installations we recently started sourcing `asdf` inside that script, but this doesn't always work because `/bin/sh` might be actual `sh` (or dash) instead of `bash`.

Of course the effort involved in changing asdf to be sh-compatible may be large which might make it not worth the effort. So my goal with this issue is to definitively find out if sh/dash is meant to be supported, and if not to have an official list of shells (even if it's not exhaustive) that are supported.

### LINKED ISSUE #1313: Improve consistency with bash syntax

### LINKED ISSUE #1397: bug: enable shellcheck & shfmt for asdf entrypoints

### LINKED ISSUE #1426: bug: Installing asdf without sourcing shell
### Describe the Bug

Hi

I have a server which runs amazon linux and I can only upload a bootstrap script to install the tools I need on the machine

That means I dont have access to sourcing the bashrc 

is there a way to install asdf by copying the bin files or symlink the files to the already available PATH environment?

thanks

### Steps to Reproduce

1. install asdf 
2. asdf command not found after script ends

### Expected Behaviour

asdf command should run

### Actual Behaviour

asdf command not found

### Environment

```shell
OS:
Linux 9a5a81343953 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux

SHELL:
GNU bash, version 4.2.46(2)-release (aarch64-koji-linux-gnu)
Copyright (C) 2011 Free Software Foundation, Inc.

ASDF VERSION:
v0.11.0-6a4f51a

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/root/.asdf

ASDF INSTALLED PLUGINS:
```

_No response_

hyperupcall:
# Summary

Earlier it was discovered that `shfmt` and `shellcheck` were not running properly on some entrypoint files (`asdf.sh` and `lib/asdf.sh`. As #1397 states, they were identified as Bash when they should have been identified as POSIX shell.

This PR updates both shfmt and shellcheck to parse the aforementioned files as `POSIX` with one slight nuance:
- `shfmt` sometimes spuriously prints errors when encountering ZSH-specific syntax (`./asdf.sh:34:16: parameter expansion requires a literal`). Because of this, `shfmt` no longer runs on `asdf.sh`.

I'm not sure if there is still confusion as to whether these files should be treated as (Bash and ZSH) or POSIX, but because we know (from #1138) that users are using shells such as `mksh` (and at least for me, I know I want `asdf` to continue to work if i jump into `dash` or `ash`), I'm treating these files as POSIX shell files. And POSIX compliance isn't too much work since we're only dealing with two files.

Moving attention back to the changes itself, I make the following fixes:

- Remove any Bash-isms or ZSH-isms that are not parseable with a POSIX-compliant shell
  - Previously, `dash` would error when parsing `asdf.sh`
- `ASDF_BIN` and `ASDF_USER_SHIMS` are no longer capitalized (they aren't environment variables), and they are now properly unset before `lib/asdf.sh` is sourced
- Remove `current_script_path="$0"`  (this doesn't work)
  - These files are sourced, so `$0` _usually_ holds the path or name the binary of the currently executing shell (ex. `/usr/bin/bash`, `dash`)
- Remove `_under="$_"`
  - `$_` rarely actually works in practice (writing something as innocuous as `: somevar` before `source .../asdf.sh` breaks this functionality). In my opinion, this flaky behavior is not worth implementing, and I added a workaround for the original Korn shell anyways.
- ~`PATH` is no longer modified (i consider it bad practice)~ (maybe I'll make a separate PR for this)

Other non-essential improvements:

- Behavior according to different shells are clearly segregated according to shell-specific variables (`{BASH,ZSH,KSH}_VERSION`)
- Uses no subshells and only builtins
- Properly prefix any created variables with `_asdf_` and properly `unset` them on errors
- Implement proper error handling and break out of script if any happen to occur
- Prefix all error messages with `asdf: Error` so the source of the error is clear (related to #1467)

I tested this with `dash`, `bash`, `zsh`, `ksh`, and `mksh`.

- Fixes #1397
- Fixes #1313
- Closes #1138
- Closes #1426
  - This is a related issue, but it should really be closed as our 'source'ing behavior is a design choice rather than a bug
- Closes #653

## ~~TODO~~

- [x] Add documentation describing explicit support for `dash`, `sh`, and all POSIX-shells

## Other Information

I'm thinking that `lib/asdf.sh` should be removed. I'm not sure what useful purpose it serves, as its contents can easily be written in `asdf.sh` (same with `lib/asdf.fish`, etc.). Thoughts?

### DIFF
--- a/lib/asdf.sh
+++ b/lib/asdf.sh
@@ -1,20 +1,20 @@
+# shellcheck shell=sh
+
 # The asdf function is a wrapper so we can export variables
 asdf() {
-  local command
-  command="$1"
-  if [ "$#" -gt 0 ]; then
-    shift
-  fi
-
-  case "$command" in
+  case $1 in
   "shell")
-    # commands that need to export variables
+    if ! shift; then
+      printf '%s\n' 'asdf: Error: Failed to shift' >&2
+      return 1
+    fi
+
+    # Invoke command that needs to export variables.
     eval "$(asdf export-shell-version sh "$@")" # asdf_allow: eval
     ;;
   *)
-    # forward other commands to asdf script
-    command asdf "$command" "$@" # asdf_allow: ' asdf '
+    # Forward other commands to asdf script.
+    command asdf "$@" # asdf_allow: ' asdf '
     ;;
-
   esac
 }
--- a/scripts/format.bash
+++ b/scripts/format.bash
@@ -3,10 +3,8 @@
 set -euo pipefail
 
 # check .sh files
-# TODO(jthegedus): unlock this check later
-# TODO  shfmt --language-dialect posix --indent 2 --write \
-# TODO  asdf.sh \
-# TODO  lib/*.sh
+shfmt --language-dialect posix --indent 2 --write \
+  lib/*.sh
 
 # check .bash files
 shfmt --language-dialect bash --indent 2 --write \
--- a/scripts/shfmt.bash
+++ b/scripts/shfmt.bash
@@ -3,10 +3,8 @@
 set -euo pipefail
 
 # check .sh files
-# TODO(jthegedus): unlock this check later
-# TODO  shfmt --language-dialect posix --indent 2 --diff \
-# TODO  asdf.sh \
-# TODO  lib/*.sh
+shfmt --language-dialect posix --indent 2 --diff \
+  lib/*.sh
 
 # check .bash files
 shfmt --language-dialect bash --indent 2 --diff \
--- a/asdf.sh
+++ b/asdf.sh
@@ -1,37 +1,77 @@
-# For Korn shells (ksh, mksh, etc.), capture $_ (the final parameter passed to
-# the last command) straightaway, as it will contain the path to this script.
-# For Bash, ${BASH_SOURCE[0]} will be used to obtain this script's path.
-# For Zsh and others, $0 (the path to the shell or script) will be used.
-_under="$_"
-if [ -z "${ASDF_DIR:-}" ]; then
-  if [ -n "${BASH_SOURCE[0]}" ]; then
-    current_script_path="${BASH_SOURCE[0]}"
-  elif [[ "$_under" == *".sh" ]]; then
-    current_script_path="$_under"
-  else
-    current_script_path="$0"
+# shellcheck shell=sh
+# shellcheck disable=SC1007
+
+# This file is the entrypoint for all POSIX-compatible shells. If `ASDF_DIR` is
+# not already set, this script is able to calculate it, if the shell is either
+# Bash, Zsh, and the or Ksh. For other shells, `ASDF_DIR` must be manually set.
+
+export ASDF_DIR="${ASDF_DIR:-}"
+
+if [ -z "$ASDF_DIR" ]; then
+  if [ -n "$BASH_VERSION" ]; then
+    # Use BASH_SOURCE[0] to obtain the relative path to this source'd file. Since it's
+    # a relative path, 'cd' to its dirname and use '$PWD" to obtain the fullpath.
+    # Use 'builtin cd' to ensure user-defined 'cd()' functions aren't called.
+    # Use variable '_asdf_old_dir' to avoid using subshells.
+
+    _asdf_old_dir=$PWD
+    # shellcheck disable=SC3028,SC3054
+    if ! CDPATH= builtin cd -- "${BASH_SOURCE[0]%/*}"; then
+      printf '%s\n' 'asdf: Error: Failed to cd' >&2
+      unset -v _asdf_old_dir
+      return 1
+    fi
+    ASDF_DIR=$PWD
+    if ! CDPATH= builtin cd -- "$_asdf_old_dir"; then
+      printf '%s\n' 'asdf: Error: Failed to cd' >&2
+      unset -v _asdf_old_dir
+      return 1
+    fi
+    unset -v _asdf_old_dir
+  elif [ -n "$ZSH_VERSION" ]; then
+    # Use '%x' to expand to path of current file. It must be prefixed
+    # with '(%):-', so it expands in non-prompt-string contexts.
+
+    # shellcheck disable=SC2296
+    ASDF_DIR=${(%):-%x}
+    ASDF_DIR=${ASDF_DIR%/*}
+  elif [ -n "$KSH_VERSION" ] && [ -z "$PATHSEP" ]; then
+    # Only the original KornShell (kornshell.com) has a '.sh.file' variable with the path
+    # of the current file. To prevent errors with other variations, such as the MirBSD
+    # Korn shell (mksh), test for 'PATHSEP' which is _not_ set on the original Korn Shell.
+
+    # shellcheck disable=SC2296
+    ASDF_DIR=${.sh.file}
+    ASDF_DIR=${ASDF_DIR%/*}
   fi
-
-  ASDF_DIR="$(dirname "$current_script_path")"
 fi
-export ASDF_DIR
-# shellcheck disable=SC2016
-[ -d "$ASDF_DIR" ] || printf "%s\n" "$ASDF_DIR is not a directory"
 
-# Add asdf to PATH
-#
-# if in $PATH, remove, regardless of if it is in the right place (at the front) or not.
-# replace all occurrences - ${parameter//pattern/string}
-ASDF_BIN="${ASDF_DIR}/bin"
-ASDF_USER_SHIMS="${ASDF_DATA_DIR:-$HOME/.asdf}/shims"
-[[ ":$PATH:" == *":${ASDF_BIN}:"* ]] && PATH="${PATH//$ASDF_BIN:/}"
-[[ ":$PATH:" == *":${ASDF_USER_SHIMS}:"* ]] && PATH="${PATH//$ASDF_USER_SHIMS:/}"
-# add to front of $PATH
-PATH="${ASDF_BIN}:$PATH"
-PATH="${ASDF_USER_SHIMS}:$PATH"
+if [ -z "$ASDF_DIR" ]; then
+  printf "%s\n" "asdf: Error: Source directory could not be calculated. Please set it manually before sourcing this file." >&2
+  return 1
+fi
+
+if [ ! -d "$ASDF_DIR" ]; then
+  printf "%s\n" "asdf: Error: Variable '\$ASDF_DIR' is not a directory: $ASDF_DIR" >&2
+  return 1
+fi
+
+_asdf_bin="${ASDF_DIR}/bin"
+_asdf_shims="${ASDF_DATA_DIR:-$HOME/.asdf}/shims"
+
+# shellcheck disable=SC3060
+if [ -n "$BASH_VERSION" ] || [ -n "$ZSH_VERSION" ]; then
+  case ":$PATH:" in
+    *":${_asdf_bin}:"*) PATH="${PATH//$_asdf_bin:/}" ;;
+  esac
+  case ":$PATH:" in
+    *":${_asdf_shims}:"*) PATH="${PATH//$_asdf_shims:/}" ;;
+  esac
+fi
+
+PATH="${_asdf_bin}:${_asdf_shims}:$PATH"
+unset -v _asdf_bin _asdf_shims
 
 # shellcheck source=lib/asdf.sh
-# Load the asdf wrapper function
 . "${ASDF_DIR}/lib/asdf.sh"
 
-unset _under current_script_path ASDF_BIN ASDF_USER_SHIMS
--- a/asdf.sh
+++ b/asdf.sh
@@ -2,8 +2,8 @@
 # shellcheck disable=SC1007
 
 # This file is the entrypoint for all POSIX-compatible shells. If `ASDF_DIR` is
-# not already set, this script is able to calculate it, if the shell is either
-# Bash, Zsh, and the or Ksh. For other shells, `ASDF_DIR` must be manually set.
+# not already set, this script is able to calculate it, but only if the shell is
+# either Bash, Zsh, and Ksh. For other shells, `ASDF_DIR` must be manually set.
 
 export ASDF_DIR="${ASDF_DIR:-}"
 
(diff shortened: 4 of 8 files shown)


scripts/shfmt.bash
```diff
@@ -3,10 +3,8 @@
 set -euo pipefail
 
 # check .sh files
-# TODO(jthegedus): unlock this check later
-# TODO  shfmt --language-dialect posix --indent 2 --diff \
-# TODO  asdf.sh \
```
jthegedus, line 8: Why remove `asdf.sh`?

scripts/format.bash
```diff
@@ -3,10 +3,8 @@
 set -euo pipefail
 
 # check .sh files
-# TODO(jthegedus): unlock this check later
-# TODO  shfmt --language-dialect posix --indent 2 --write \
-# TODO  asdf.sh \
```
jthegedus, line 8: Why remove `asdf.sh`?

lib/asdf.sh
```diff
@@ -1,20 +1,20 @@
+# shellcheck shell=sh
```
jthegedus, line 1: What is this giving us that the `scripts/shellcheck.bash` isn't already?

jthegedus:
>I'm thinking that lib/asdf.sh should be removed. I'm not sure what useful purpose it serves, as its contents can easily be written in asdf.sh (same with lib/asdf.fish, etc.). Thoughts?

That structure predates me, so not sure of motivation. Perhaps @Stratus3D can shed some light on it.

jthegedus:
This is a monumental change and can potentially break a lot of stuff. I am leaning towards merging other PRs first, releasing them, then merging this after the next release so that people can play with it on `HEAD` without any other changes confusing bug reports.

jthegedus:
>shfmt sometimes spuriously prints errors when encountering ZSH-specific syntax (./asdf.sh:34:16: parameter expansion requires a literal). Because of this, shfmt no longer runs on asdf.sh.

Right. It seems if we are targeting actual `sh` then we should not have ZSH specific syntax or specifically ignore those lines, or something. Turning off the formatter on the whole file for one edge case is not the ideal solution.

jthegedus:
@Stratus3D hasn't replied to https://github.com/asdf-vm/asdf/issues/1397#issuecomment-1367880907 But I believe we're on the same page about treating these files as `sh` and not `bash` :+1:

scripts/shfmt.bash
hyperupcall: This is related to the thing I noted earlier:

> shfmt sometimes spuriously prints errors when encountering ZSH-specific syntax (./asdf.sh:34:16: parameter expansion requires a literal). Because of this, shfmt no longer runs on asdf.sh.

lib/asdf.sh
hyperupcall: Adding that to the top of the file allows editors with shellcheck plugins to know how to treat the file. These sort of comments also obviate the need to pass --shell to each invocation since it can be inferred from the comment.

hyperupcall:
> This is a monumental change and can potentially break a lot of stuff. I am leaning towards merging other PRs first, releasing them, then merging this after the next release so that people can play with it on HEAD without any other changes confusing bug reports.

That sounds good :+1: 

> However, that doesn't solve the partial support in a single file. So perhaps it is best to extract the .zsh specific stuff into it's own script to be tested (or not) idependently, leaving the sh in this file lint-able for sh specifically. It is a very critical piece of code to turn this off for.

Just to be a bit more clear, it was shfmt having the issues parsing the file - shellcheck works fine.

But I definitely see your point about turning off formatting for a particular file. I think there is another way I can express the same thing while keeping the formatter happy

hyperupcall:
Okay so I made the ZSH change, and now shfmt is tripping up again. Here is the change:

```diff
diff --git a/asdf.sh b/asdf.sh
index d7dd4c7..5e66fc4 100644
--- a/asdf.sh
+++ b/asdf.sh
@@ -29,11 +29,10 @@ if [ -z "$ASDF_DIR" ]; then
     fi
     unset -v _asdf_old_dir
   elif [ -n "$ZSH_VERSION" ]; then
-    # Use '%x' to expand to path of current file. It must be prefixed
-    # with '(%):-', so it expands in non-prompt-string contexts.
+    # Use '%x' to expand to path of current file. It must be used with
+    # '-P', since depends on prompt expansion.
 
-    # shellcheck disable=SC2296
-    ASDF_DIR=${(%):-%x}
+    print -P -v ASDF_DIR '%x'
     ASDF_DIR=${ASDF_DIR%/*}
   elif [ -n "$KSH_VERSION" ] && [ -z "$PATHSEP" ]; then
     # Only the original KornShell (kornshell.com) has a '.sh.file' variable with the path
```

With `--language-dialect bash`, `shfmt` shows: `asdf.sh:43:16: invalid parameter name`
With `--language-dialect posix`, `shfmt` shows: `asdf.sh:19:46: arrays are a bash/mksh feature`.

It looks impossible to make the `asdf.sh` file parseable by `shfmt`. Since it's just the formatter and not linter, are you okay with just dropping the format check? I understand it's not ideal, but there doesn't seem to be much of a choice.

hyperupcall:
Hey can this get another look? v0.11.2 was just released a week or so ago, so now is an opportune time to merge as mentioned [here](https://github.com/asdf-vm/asdf/pull/1480#issuecomment-1435966579).

lib/asdf.sh
jthegedus: I would like to be consistent and since we aren't using this in the rest of the codebase to remove it. I would rather just be explicit in the `scripts/*` we use than spreading this information throughout the codebase.

jthegedus:
>It looks impossible to make the asdf.sh file parseable by shfmt. Since it's just the formatter and not linter, are you okay with just dropping the format check?

The errors from `shfmt` certainly seem like linting errors. Call it what you want but I would like to know information like this: `asdf.sh:19:46: arrays are a bash/mksh feature`

I believe this is just revealing that we should have separate `asdf.*` files for each of these other shells, as we do with Fish, Elvish & Nushell.

Just because each of these shells share a lineage with some common syntax or features is a bad reason to keep the code in one place and turn off our automatic validation on such a crucial aspect of our code.

So my suggestion now would be either:

1. Breaking changes:
	- remove the KSH branch of code, we should add `asdf.ksh` later
	- set `shmft` to parse the files as `bash` for now. We should add `asdf.bash` later
	- after we add `asdf.bash`, then change `shfmt` to parse this file as `sh`
2. Non-breaking but less clear:
    - extract each branch in `asdf.sh` into their own scripts. Eg: the KSH branch calls `asdf-entrypoint.ksh`, ZSH branch calls `asdf-entrypoint.zsh`
    - then we can add specific `asdf.ksh`, `asdf.zsh` scripts later and give users a migration path
    

I'm thinking out loud here. We could just turn off `shfmt` I guess... Just given the `asdf.fish` etc it seems that is a better path to go down.

hyperupcall:
> The errors from shfmt certainly seem like linting errors. Call it what you want but I would like to know information like this: asdf.sh:19:46: arrays are a bash/mksh feature

I suppose you could argue that the `shfmt` formatter is printing "linting errors", but it doesn't matter because there are no "issues" that `shfmt` detects that `shellcheck` doesn't:

- With the "arrays are a Bash/mksh feature" (` "${BASH_SOURCE[0]%/*}"`) we [purposefully disable SC3054](https://github.com/asdf-vm/asdf/blob/1d34bf471d95094f7fdb4066c3dd586d9f72c9b4/asdf.sh#L18) because it's a false positive - the code parses on all applicable shells.
- With the "`.sh.file` is an invalid parameter name", ShellCheck catches that too, but again it's an obvious false positive and [SC2296 is supressed](https://github.com/asdf-vm/asdf/blob/1d34bf471d95094f7fdb4066c3dd586d9f72c9b4/asdf.sh#L43) - the code parses on all applicable shells.

So nothing (except formatting) is lost when turning `shfmt` off - ShellCheck already catches everything for us.

hyperupcall:
I also want to mention, that before my changes, there was no clear separation between shells, functionality was broken for both KSH and POSIX Shells, and ShellCheck wasn't being ran on `asdf.sh`. My changes added ShellCheck, improved error messages and fixed all aforementioned issues with full backwards compatibility.

Since my changes only fix things, would it be OK to just merge these changes, and if you really wish to separate things into different files (which I would strongly disagree with), to do it later?

jthegedus:
>So nothing (except formatting) is lost when turning shfmt off - ShellCheck already catches everything for us. So since ShellCheck is active, would we agree the whole asdf.sh thing is fine since ShellCheck will catch any errors that are introduced?

Yes, you're right. Sorry, I lost context of the past conversation as there was a gap in time in which I was able to review and contribute to this conversation.

>Since my changes only fix things, would it be OK to just merge these changes, and if you really wish to separate things into different files (which I would strongly disagree with), to do it in some later PR?

Yes, let's just merge this and we can discuss the other changes in future Issues.

jthegedus (approved):
This is great work, thanks very much @hyperupcall :pray: 

We will merge this and let people test it with `asdf update --head`

Stratus3D (commented):
This looks like a really great set of changes. Thanks for your hard work on this @hyperupcall and @jthegedus !

Stratus3D:
> > I'm thinking that lib/asdf.sh should be removed. I'm not sure what useful purpose it serves, as its contents can easily be written in asdf.sh (same with lib/asdf.fish, etc.). Thoughts?
> 
> That structure predates me, so not sure of motivation. Perhaps @Stratus3D can shed some light on it.
> 
> EDIT: **I would say we shouldn't change this for now.**

I think I've wondered the same thing. I seem to remember thinking all the logic in it could either be moved into `bin/asdf` or `asdf.sh`. It looks like this has already been done though, so good work!

PR #1481, opened 2023-02-19
fix: support nushell v0.75.0

hyperupcall:
# Summary

Ever since Nushell v0.75.0 was released on homebrew, our CI fails (see #1479, #1480, and #1482). This fixes the Nushell tests to be compatible with the latest version

The only necessary change is the `export extern "asdf"` => `export extern main`, which is due to [recent module changes](https://www.nushell.sh/blog/2023-01-31-nushell_0_75.html#main-command-exported-from-module-defines-top-level-module-command-kubouch).

Everything else is just whitespace and improved test reliability. The tests pass on v0.75.0 and earlier (not just v0.75.0).


jthegedus (approved):
Great work, thanks!

jthegedus:
Made this a `fix:` since it touches code executed by users.

PR #1485, opened 2023-02-21
fix: Better handling with paths that include spaces

hyperupcall:
# Summary

I saw that there was a `ASDF_BATS_SPACE_IN_PATH` variable, which seemed to be part of a gradual transition to make the tests work better with whitespace stuff. Probably with help from #1433, it was somewhat trivial to make the proper fixes and remove this variable entirely.

One test was left out because it seems that change requires a fix in the `.tool-versions` parser (related to the `path:~/src/elixir` feature)

### DIFF
--- a/lib/commands/command-plugin-remove.bash
+++ b/lib/commands/command-plugin-remove.bash
@@ -21,7 +21,16 @@ plugin_remove_command() {
   rm -rf "$(asdf_data_dir)/installs/${plugin_name}"
   rm -rf "$(asdf_data_dir)/downloads/${plugin_name}"
 
-  grep -l --null "asdf-plugin: ${plugin_name}" "$(asdf_data_dir)"/shims/* 2>/dev/null | xargs -0 rm -f
+  local is_nullglob_disabled=
+  shopt -q nullglob || is_nullglob_disabled=yes
+  shopt -s nullglob
+  for f in "$(asdf_data_dir)"/shims/*; do
+    if grep -q "asdf-plugin: ${plugin_name}" "$f"; then
+      rm -f "$f"
+    fi
+  done
+  [ "$is_nullglob_disabled" = 'yes' ] && shopt -u nullglob
+  unset -v is_nullglob_disabled
 
   asdf_run_hook "post_asdf_plugin_remove" "$plugin_name"
   asdf_run_hook "post_asdf_plugin_remove_${plugin_name}"
--- a/lib/commands/command-plugin-remove.bash
+++ b/lib/commands/command-plugin-remove.bash
@@ -21,7 +21,7 @@ plugin_remove_command() {
   rm -rf "$(asdf_data_dir)/installs/${plugin_name}"
   rm -rf "$(asdf_data_dir)/downloads/${plugin_name}"
 
-  grep -l "asdf-plugin: ${plugin_name}" "$(asdf_data_dir)"/shims/* 2>/dev/null | xargs rm -f
+  grep -l --null "asdf-plugin: ${plugin_name}" "$(asdf_data_dir)"/shims/* 2>/dev/null | xargs -0 rm -f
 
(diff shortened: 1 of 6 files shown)


test/test_helpers.bash
```diff
@@ -6,11 +6,12 @@ bats_require_minimum_version 1.7.0
...
 setup_asdf_dir() {
-  if [ -n "${ASDF_BATS_SPACE_IN_PATH:-}" ]; then
-    BASE_DIR="$(mktemp -dt "asdf with spaces.XXXX")"
+  if [ "$BATS_TEST_NAME" = 'test_shim_exec_should_use_path_executable_when_specified_version_path-3a-3cpath-3e' ]; then
+    BASE_DIR="$(mktemp -dt "asdf_with_no_spaces.XXXX")"
   else
-    BASE_DIR="$(mktemp -dt asdf.XXXX)"
+    BASE_DIR="$(mktemp -dt "asdf with spaces.XXXX")"
   fi
+
```
jthegedus, line 14: Can we not just override `BASE_DIR` in the test itself instead of doing this conditional setup?

hyperupcall: Yeah that's a much better idea

:+1: Made those changes

jthegedus:
The changes I requested seemed to have introduced a bunch of shellcheck issues we would need to ignore. Are you able to update @hyperupcall ?

hyperupcall:
> The changes I requested seemed to have introduced a bunch of shellcheck issues we would need to ignore. Are you able to resolve them @hyperupcall ?

Wish I mentioned this earlier, but the reason why the Bats test fails is because the newer changes moves the setting of `BASE_DIR` inside of the function. This means that it has no effect since `BASE_DIR` is used in `setup`, which runs before the test.

hyperupcall:
I reverted those changes, and the `BASE_DIR` is now set inside of `setup()`, so things should work. I know it's kind of ugly since the condition depends on the name of the test and there is a separation of sorts - do you think there is a better way?

My first thought would be to move the invocation of `setup_asdf_dir` from within `setup()`, to each individual function, just for the tests in `shim_exec.bats`, but it seems that would make all that code harder to read and more repetitive just for the sake of a single test.

jthegedus:
It is a shame we cannot opt-out of using the setup for a single test in the file. The most clean alternative I can think of is having a separate test file for these specific tests.

If someone thinks of something better, we can discuss then, in the meantime we will merge this.

jthegedus (approved):

lib/commands/command-plugin-remove.bash
```diff
@@ -21,7 +21,16 @@ plugin_remove_command() {
...
+  local is_nullglob_disabled=
+  shopt -q nullglob || is_nullglob_disabled=yes
+  shopt -s nullglob
+  for f in "$(asdf_data_dir)"/shims/*; do
+    if grep -q "asdf-plugin: ${plugin_name}" "$f"; then
+      rm -f "$f"
+    fi
+  done
+  [ "$is_nullglob_disabled" = 'yes' ] && shopt -u nullglob
+  unset -v is_nullglob_disabled
```
Stratus3D, line 33: Is there really not a better way of doing this? This seems like quite a bit of code just to find and delete shims containing a certain line.

hyperupcall: There [was](Ihttps://github.com/hyperupcall/asdf/blob/82b9ef01e26f39b375c1b462979aa7ffd069df8a/lib/commands/command-plugin-remove.bash#L24), but it didn't work on Alpine.

I suppose I could have replaced the nullglob stuff with a `[ -f "$f" ]` in the loop

Stratus3D: I guess I'd prefer `[ -f "$f" ]` in the loop as it feels harder to get wrong - it's easy for someone to forget to unset an option or reset it to the wrong value. In this case it looks correct so not an issue.

PR #1486, opened 2023-02-21
docs: list asdf-nodejs deps for enterprise linux in guide

ygxxii:
# Summary

I saw that there was no "Dependency Installation" for Enterprise Linux.


jthegedus:
Thanks for the PR. I intend to merge it.

Note however, this list is just an example of the dependencies from `asdf-nodejs` and is part of a walkthrough guide, not necessarily indicative of the list of tools required for asdf itself.

jthegedus (approved):
Thanks

PR #1495, opened 2023-03-06
docs: fix Nushell-Homebrew setup instructions

kimufly:
# Summary

Fix the problem that ASDF_NU_DIR does not fight for an invisible newline character at the end of the brew --prefix command

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -277,7 +277,7 @@ Completions are automatically configured
 Add `asdf.nu` to your `~/.config/nushell/config.nu` with:
 
 ```shell:no-line-numbers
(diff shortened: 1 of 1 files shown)


Stratus3D:
Can anyone who uses nushell test this and confirm it fixes the problem?

@kimufly can you write a unit test for this issue?

jthegedus:
We don't really test these setup instructions in any of our CI pipelines. The Nushell tests just directly setup using Nu script instead of each permutation of Shell+asdf installation method.

If we were to add tests I guess they would live https://github.com/asdf-vm/asdf/blob/master/test/asdf_nu.bats but would have to detect OS and execute conditionally.

jthegedus:
I'm going to merge this and we can improve the Nushell test suite over time. There are a few issues already raised in other Issues which are being addressed individually.

PR #1496, opened 2023-03-08
fix: Prepend asdf directories to the PATH in Nushell

etoal83:
# Summary

In the Nushell setup script, I change the paths to asdf shims and bin to be prepended in front of the existing `$PATH` instead of being appended at the end. This change makes asdf plugins have priority over the existing commands in the search path as well as other shells' setup.

### DIFF
--- a/asdf.nu
+++ b/asdf.nu
@@ -7,8 +7,8 @@ def-env configure-asdf [] {
     let asdf_bin_dir = ( $env.ASDF_DIR | path join 'bin' )
 
 
-    let-env PATH = ( $env.PATH | split row (char esep) | where { |p| $p != $shims_dir } | append $shims_dir )
-    let-env PATH = ( $env.PATH | split row (char esep) | where { |p| $p != $asdf_bin_dir } | append $asdf_bin_dir )
+    let-env PATH = ( $env.PATH | split row (char esep) | where { |p| $p != $shims_dir } | prepend $shims_dir )
+    let-env PATH = ( $env.PATH | split row (char esep) | where { |p| $p != $asdf_bin_dir } | prepend $asdf_bin_dir )
 
 }
 


jthegedus:
All our Shell integrations use append due to significant user feedback when changing this to prepend in the past. I think we should stick with append.

What is the specific use case troubling you that requires this?

Stratus3D:
@jthegedus it appears we **prepend** to the asdf paths to `$PATH` in `asdf.sh`. It appears the changes here from @etoal83 make Nushell consistent with Bash/Zsh in this regard. I am in favor of this change if I'm understanding things correctly.

To me it also seems easier to reason about a script that adds something to the beginning of the `$PATH`. When you source it in your `.zshrc`/`.bashrc` you be certain that asdf is added at the beginning of `$PATH`, anything that comes before it in the rc file ends up after asdf on `PATH`, and anything after it in the rc file comes before it on the `PATH`.

Stratus3D (approved):

etoal83:
@jthegedus @Stratus3D 

Thank you for commenting! What I intended with this PR is correctly explained by @Stratus3D .

> What is the specific use case troubling you that requires this?

Appending the path becomes a problem when the same command already exists in the user environment. Because `asdf.nu` appends the `$shims_dir` and `$asdf_bin_dir` paths to the existing `$PATH`, a preinstalled version of executable was used in the command over the asdf version. I think in most situations people installing asdf would prefer to use the asdf version of the executable rather than the existing one.

This PR changes the asdf path setting from/to:

- before: `$PATH:$shims_dir:$asdf_bin_dir`
- after: `$asdf_bin_dir:$shims_dir:$PATH`

Sorry if the use of my word "append" and "prepend" were misleading😣

jthegedus:
Sorry for the confusion, I misunderstood the current state. I should have validated :sweat_smile:

PR #1502, opened 2023-03-16
fix: nushell plugin list all (#1501)

### LINKED ISSUE #1501: Nushell `asdf plugin list all` does not list all plugins
### Describe the Bug

Running `asdf plugin list all` in Nushell does not list all plugins because the regex is broken and won't work for links that start with `https` or don't end with `.git`:
https://github.com/asdf-vm/asdf/blob/0adc6c11fb4f87dbb476f8b61e4cf8fb7613599b/asdf.nu#L102

A fix for the regex could be: `(?P<name>.+)\s+?(?P<installed>[*]?)(?P<repository>(?:git|http|https).+)`. The `.git` does not seem to be required by `git clone` (tested on git version 2.40.0).

### Steps to Reproduce

Just run `asdf plugin list all` in Nushell.

### Expected Behaviour

It should print the whole list of plugins.

### Actual Behaviour

Stops on the first line that does not match the regex.

### Environment

```shell
OS:
Darwin acidghost-MBP.local 22.3.0 Darwin Kernel Version 22.3.0: Mon Jan 30 20:38:37 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.8.1 (x86_64-apple-darwin22.0)

ASDF VERSION:
v0.11.2

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/Users/acidghost/.dotfiles/asdf-vm

ASDF INSTALLED PLUGINS:
golang                       https://github.com/kennyp/asdf-golang.git master f006a12
java                         https://github.com/halcyon/asdf-java.git master 54b7426
julia                        https://github.com/rkyleg/asdf-julia.git master 1788f0f
perl                         https://github.com/ouest/asdf-perl.git master 4edc8ab
ruby                         https://github.com/asdf-vm/asdf-ruby.git master 7a671a2
```

_No response_

acidghost:
# Summary

Fixes #1501.

### DIFF
--- a/test/asdf_nu.bats
+++ b/test/asdf_nu.bats
@@ -9,6 +9,14 @@ setup() {
   if ! command -v nu; then
     skip "Nu is not installed"
   fi
+
+  setup_asdf_dir
+  setup_repo
+  install_dummy_plugin
+}
+
+teardown() {
+  clean_asdf_dir
 }
 
 cleaned_path() {
@@ -109,4 +117,21 @@ cleaned_path() {
 
   result=$(echo "$output" | grep "ASDF INSTALLED PLUGINS:")
   [ "$result" != "" ]
+}
+
+@test "parses the output of asdf plugin list all" {
+  run nu -c "
+    hide-env -i asdf
+    hide-env -i ASDF_DIR
+    let-env PATH = ( '$(cleaned_path)' | split row ':' )
+    let-env ASDF_NU_DIR = '$PWD'
+
+    source asdf.nu
+    asdf plugin list all | to csv -n"
+
+  [ "$status" -eq 0 ]
+  [ "$output" = "\
+bar,false,http://example.com/bar
+dummy,true,http://example.com/dummy
+foo,false,http://example.com/foo" ]
 }
--- a/asdf.nu
+++ b/asdf.nu
@@ -99,7 +99,7 @@ module asdf {
 
     # list all available plugins
     export def "asdf plugin list all" [] {
-        let template = '(?P<name>.+)\s+?(?P<installed>[*]?)(?P<repository>(?:git|http).+\.git)'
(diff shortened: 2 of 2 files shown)


jthegedus:
@acidghost can you add a regression test to https://github.com/asdf-vm/asdf/blob/master/test/asdf_nu.bats ?

acidghost:
What should it test? Check that the output of `asdf plugin list all` is the same for Nushell and bash in terms of repositories?

jthegedus:
It should test the permutations of this regex you're changing to ensure that the command runs successfully for expected repositories.

test/asdf_nu.bats
```diff
@@ -110,3 +110,23 @@ cleaned_path() {
...
+dummy                        *http://example.com/dummy
+foo                           http://example.com/foo
+foos                          https://example.com/foo
+fooss                        *https://example.com/foo
+foog                          git@example.com/foo.git
+foogs                        *git@example.com/foo.git"
+  local nr_lines=7
+
+  run nu -c "
+    let re = '(?P<name>.+)\s+?(?P<installed>[*]?)(?P<repository>(?:git|http|https).+)'
```
jthegedus, line 126: As with the test above, sourcing `asdf.nu` would then let you run the commands we're trying to assert. This way we would be testing the `asdf.nu` and not code duplicated from that script here.

jthegedus (approved):
Looks good, thanks!

jthegedus:
> @jthegedus I think this is what you meant. To fully test this, there should be more types of test repository URLs (e.g. `https`, `git`, etc.). I can add these in another PR as it would require to modify most of the test cases to update the expected values.

Yeah, just submit a follow up with some more test cases. There's probably a way to hack more values into just this test case.

Issue #1503, opened 2023-03-17
Nushell `asdf plugin list --urls` regex is broken


acidghost:
### Describe the Bug

Nushell's `asdf plugin list --urls` has the wrong regex and does not display anything. The regex expects an url of the form `git@[...].git`: https://github.com/asdf-vm/asdf/blob/0adc6c11fb4f87dbb476f8b61e4cf8fb7613599b/asdf.nu#L82

### Steps to Reproduce

Just run `asdf plugin list --urls` in Nushell with plugins installed from a repo url that starts with `https`.

### Expected Behaviour

It shows the table of plugins.

### Actual Behaviour

Does not display anything.

### Environment

```shell
OS:
Darwin acidghost-MBP.local 22.3.0 Darwin Kernel Version 22.3.0: Mon Jan 30 20:38:37 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.8.1 (x86_64-apple-darwin22.0)

ASDF VERSION:
v0.11.3

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/Users/acidghost/.dotfiles/asdf-vm

ASDF INSTALLED PLUGINS:
clojure                      https://github.com/asdf-community/asdf-clojure.git master 490074c
golang                       https://github.com/kennyp/asdf-golang.git master f006a12
java                         https://github.com/halcyon/asdf-java.git master 54b7426
julia                        https://github.com/rkyleg/asdf-julia.git master 1788f0f
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master c9e5df4
perl                         https://github.com/ouest/asdf-perl.git master 4edc8ab
python                       https://github.com/danhper/asdf-python.git master 8505457
ruby                         https://github.com/asdf-vm/asdf-ruby.git master 7a671a2
```

jthegedus:
Hey @acidghost now you have an idea of what the test cases look like from #1502 , I would love the help here for another PR to resolve this, if you have the capacity :pray:

PR #1507, opened 2023-03-21
fix: Nushell plugin list --urls

acidghost:
# Summary

Fixes: #1503.

## Other Information

Also adds tests for the `--refs`flag and its combination with `--urls`.

Note that installed plugins can have a "remote" URL of two forms:
- a remote git URL (i.e. starting with `http[s]|git` and terminating with `.git`)
- an absolute path to a local folder

**N.B.**: At the moment this PR is only testing for plugins added from a local folder (second case above). To test plugins that have been cloned from an *actually remote* repository (first case above) there have to be testing fixtures that clone them first.

### DIFF
--- a/asdf.nu
+++ b/asdf.nu
@@ -79,8 +79,10 @@ module asdf {
     ] {
 
         let params = [
-            {name: 'urls', enabled: $urls, template: '\s+?(?P<repository>git@.+\.git)', flag: '--urls'}
-            {name: 'refs', enabled: $refs, template: '\s+?(?P<branch>\w+)\s+(?P<ref>\w+)', flag: '--refs'}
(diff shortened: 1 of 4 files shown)


jthegedus (approved):
Thanks!

jthegedus:
I am not sure what we can do about external test fixtures for plugins from actual remotes :shrug:

acidghost:
@jthegedus  Well, one solution would be to create some test repositories and actually clone them. Shall I open a new issue?

jthegedus:
An issue to discuss a solution would be best, yes. I would like to explore other alternatives if possible though, as maintaining a whole set of repos for testing is a bit annoying.

PR #1514, opened 2023-03-24
docs: Improve wording of env vars section

hyperupcall:
Improves wording of environmental variables section in two ways:

1. No longer refer to `~/.asdf` as an absolute path

Beginners may take this too literally and do something like `export ASDF_DATA_DIR=~/.asdf`, which won't work since tilda expansion doesn't occur here (since the `~/` isn't the first two characters in the word, the `AS` in `ASDF_DATA_DIR` is).

To prevent misconceptions or potential issues, this has been removed.

2. Update description of `ASDF_DIR`

This variable no longer defaults to `~/.asdf`, but instead is [dynamically set](https://github.com/asdf-vm/asdf/blob/05a8bb940a9ce1a83dcda2a2bf52dc3d0786bd54/lib/utils.bash#L27) and always refers to the true location of the source files.

### DIFF
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -134,8 +134,8 @@ Disabling the plugin short-name repository does not remove plugins previously in
 
 - `ASDF_CONFIG_FILE` - Defaults to `~/.asdfrc` as described above. Can be set to any location.
 - `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` - The filename of the file storing the tool names and versions. Defaults to `.tool-versions`. Can be any valid filename. Typically you should not override the default value unless you know you want asdf to ignore `.tool-versions` files.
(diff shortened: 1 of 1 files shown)


hyperupcall:
That's a nice improvement! Once I finally finish up #1351, the extra space will make it easier to write and read the new behavior.

I did change the wording of `ASDF_DATA_DIR` to be more accurate - in nearly all cases (since `$HOME` is [always set](https://github.com/asdf-vm/asdf/blob/bbcbddcdd4ffa0f49c3772b66d87331420fa5727/lib/utils.bash#L44)), the value will be `$HOME/.asdf`. Which is good because otherwise `asdf` would require superuser priviledges if installed through the AUR.

jthegedus (approved):

PR #1516, opened 2023-03-26
fix: lint & style errors in `bin/asdf`

hyperupcall:
# Summary

I noticed that `bin/asdf` wasn't added to the checkstyle script. This does that, runs `checkstyle --fix`, and makes miscellaneous changes to the types and comments in the checkstyle script.

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -16,9 +16,9 @@ find_cmd() {
   done
 
   if [ -f "$cmd_dir/$cmd_name" ]; then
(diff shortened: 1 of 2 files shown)


jthegedus (approved):
LGTM, thanks

jthegedus:
@hyperupcall It would be great if this checkstyle script accepted the file globs as args so we could use it in a Bash script similar to how we use Shellcheck and Shfmt so issues of missing files like this are all visible in the same way.

Just a thought.

PR #1517, opened 2023-03-26
test: skip tests for uninstalled shells when run locally

hyperupcall:
# Summary

Before:
```sh
 - exports ASDF_DIR (skipped: Elvish version is not at least 0.18. Found ..)
```

After:
```sh
 - exports ASDF_DIR (skipped: Elvish not installed)
```

## Other Information

This also makes sure that CI fails if a particular shell isn't installed.

### DIFF
--- a/test/asdf_elvish.bats
+++ b/test/asdf_elvish.bats
@@ -8,12 +8,16 @@ setup() {
   export XDG_DATA_HOME=
   export XDG_DATA_DIRS=
 
+  if ! command -v elvish &>/dev/null && [ -z "$GITHUB_ACTIONS" ]; then
+    skip 'Elvish not installed'
+  fi
+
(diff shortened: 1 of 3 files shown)


jthegedus:
I think we discussed this previously with Fish, but it would be good if each of these test skips for missing Shell types only skipped locally and always erred in the CI environment so we know when we break our CI Shell setup process.

Can you update this to skip this test when in CI?

hyperupcall:
Ooh, right yeah I can do that

jthegedus (approved):
Great, thanks!

Stratus3D:
Looks great! I like the `&& [ -z "$GITHUB_ACTIONS" ]` condition ensuring it'll still run in CI.

PR #1518, opened 2023-03-26
fix: assign default values to all internal variables

hyperupcall:
# Summary

As mentioned [here](https://github.com/asdf-vm/asdf/pull/1513#discussion_r1147979242), this fixes the behavior of defaulting `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` to the correct version. This is also done to `$ASDF_DATA_DIR`.

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -5,9 +5,6 @@ GREP_OPTIONS="--color=never"
 # shellcheck disable=SC2034
 GREP_COLORS=
 
-ASDF_DIR=${ASDF_DIR:-''}
-ASDF_DATA_DIR=${ASDF_DATA_DIR:-''}
-
 asdf_version() {
   local version git_rev
   version="v$(cat "$(asdf_dir)/version.txt")"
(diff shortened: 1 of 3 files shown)


jthegedus (approved):
LGTM

jthegedus:
One assumes that the fact all our tests passed without changes that this is not a breaking change :thinking:

hyperupcall:
Unless I'm missing something, I don't think it's a breaking change. The environment variables are a part of the "public" API as you mention, but as far as I can tell, the behavior of the env vars isn't changing, the thing that changes is where it is set (if at all).

If asdf launches a subprocess like executing a plugin, the value of the environment variable is the same, so if people were depending on it, the value doesn't change. And the fix is that previous unset environment variables are now set, so that doesn't break anything. I suppose theoretically someone plugin somewhere could be relying on the fact that one or more of these variables were incorrectly unset or set, but empty - I don't personally care about that because if those were breaking changes, then too many changes would be breaking change due to Hyrum's Law.

jthegedus:
Thanks for the input, `fix:` it is then!

lib/utils.bash
```diff
@@ -19,21 +16,12 @@ asdf_version() {
...
-    local current_script_path=${BASH_SOURCE[0]}
-    export ASDF_DIR
-    ASDF_DIR=$(
-      cd "$(dirname "$(dirname "$current_script_path")")" || exit
-      printf '%s\n' "$PWD"
-    )
-  fi
-
-  printf "%s\n" "$ASDF_DIR"
+asdf_default_tool_versions_filename() {
```
Stratus3D, line 19: nitpick: The default tool versions file name is `.tool-versions`. If the user has customized it by setting `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` then this function won't return the default filename. I think `asdf_tool_versions_filename` would be a better function name.

PR #1523, opened 2023-03-28
feat: Add search subcommand

tooosay:
# Summary

I have added a search subcommand. 
which searches if the plugin exists from asdf-vm/asdf-plugins repository.

eg:
```
$ asdf search vim
neovim
vim
```
command-search.sh is a very simple shell script.

Also, I have added help that denotes the search subcommand.

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -60,6 +60,8 @@ asdf shim-versions <command>            List the plugins and versions that
 asdf update                             Update asdf to the latest stable release
 asdf update --head                      Update asdf to the latest on the master branch
 
+asdf search <name>                      Searches if the command exists. Not fuzzy find.
+
 RESOURCES
 GitHub: https://github.com/asdf-vm/asdf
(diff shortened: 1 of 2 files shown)


hyperupcall:
I'm not sure if there was a discussion on how much this feature is needed, but about the code, I'm pretty sure the api.github.* endpoints rate limited if no token if provided?

jthegedus:
I do want to add search in the future but that would be on the pathway to removing the plugins repository, not to further search the plugins repo.

The plugin repo is already cloned locally to people's machines, so `curl`ing the GitHub API is not preferred over grepping the existing local file.

`asdf plugin list all | grep "<tool>"` currently gives users this same functionality.

Given these factors we will not be merging this implementation of search.

@tooosay thanks for taking the time to contribute. Help is always welcome and we have many open tickets and PRs that need rework if you find yourself with some time.

PR #1524, opened 2023-03-29
Salmon

hyperupcall:
# Summary

As per [this comment](https://github.com/asdf-vm/asdf/pull/1521#issuecomment-1486619699), this matches the behavior of Fish with Bash, Nushell, and all the other shells. Along with a few other things

There were some bugs with the previous Fish implementation that are now fixed:

- If `ASDF_DIR` was set to an empty string, the value is used (unlike in Bash/Zsh, Elvish, Nushell, which replaces it with a correct one)
- `ASDF_DIR` is now always an absolute path to the asdf directory (previously it could have been a `.`). (Since it's an absolute path, we now export it so that asdf can use it instead of recalculating)
- Depending on version, either `PATH` or `fish_user_paths` (through `fish_add_path`) would be modified. Now, only the latter is potentially modified

It seems that `fish_add_path` potentially modifies the placement of an entry in `$fish_user_paths`, even if `--move` is not specified, so I opted to prepend to `$fish_user_paths` manually, so the behavior is consistent with all the other shells.

Now we also use `fish_indent` to make sure Fish scripts are formatted nicely (even if we have relatively few of them)

### DIFF
--- a/scripts/format.bash
+++ b/scripts/format.bash
@@ -23,3 +23,6 @@ shfmt --language-dialect bash --indent 2 --write \
 # check .bats files
 shfmt --language-dialect bats --indent 2 --write \
   test/*.bats
+
+# check .fish files
+fish_indent --write ./**/*.fish
--- a/scripts/shfmt.bash
+++ b/scripts/shfmt.bash
@@ -23,3 +23,6 @@ shfmt --language-dialect bash --indent 2 --diff \
 # check .bats files
 shfmt --language-dialect bats --indent 2 --diff \
   test/*.bats
+
+# check .fish files
+fish_indent --check ./**/*.fish
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,6 @@ insert_final_newline = true
 
 [*.py]
 indent_size = 4
+
+[*.fish]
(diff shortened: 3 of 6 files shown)


scripts/shfmt.bash
```diff
@@ -23,3 +23,6 @@ shfmt --language-dialect bash --indent 2 --diff \
 # check .bats files
 shfmt --language-dialect bats --indent 2 --diff \
   test/*.bats
+
+# check .fish files
+fish_indent --check ./**/*.fish
```
jthegedus, line 28: :tada:

scripts/format.bash
```diff
@@ -23,3 +23,6 @@ shfmt --language-dialect bash --indent 2 --write \
 # check .bats files
 shfmt --language-dialect bats --indent 2 --write \
   test/*.bats
+
+# check .fish files
+fish_indent --write ./**/*.fish
```
jthegedus, line 28: :tada:

jthegedus (approved):
I don't use fish anymore so haven't tested this locally myself, but LGTM

jthegedus:
Merging to get people testing this.

PR #1531, opened 2023-04-06
ci: pin all shell versions for Ubuntu test environment

jthegedus:
# Summary

Currently we pin Shell versions for Elvish & Nushell. We do not for Fish or Powershell.

We should be consistent. So this PR introduces pinning to Fish & Powershell.

Homebrew installation on the macOS runner wasn't using any pinned versions. We will leave this as is for now.


jthegedus:
I guess most Homebrew users use whatever is "latest" in Homebrew anyway, so it is probably a "better" test to represent user.

In future I would like to expand the CI workflows for each OS to:
- test Shell minimum supported versions with a pinned set of versions
- test Shell latest versions with a pinned set of versions
- test whatever the OS package manager defaults to

Stratus3D:
Looks good to me!

PR #1532, opened 2023-04-06
feat: Support configurable `ASDF_CONCURRENCY`

### LINKED ISSUE #1097: ASDF_CONCURRENCY should be configurable

There's no way to configure the value for ASDF_CONCURRENCY today, making it almost useless for plugin maintainers, the current behavior is to max out all cores, a situation rarely desirable

We should provide a way for configuring ASDF_CONCURRENCY via env vars or via `$HOME/.asdfrc` (both preferably)

ASDF_CONCURRENCY just gets the max number of processors/cores

The other way for plugin maintainers to provide customizable concurrency settings is to provide their CONCURRENCY env var, which makes the ecosystem more fragmented.

hyperupcall:
# Summary

Fixes #1097

## Other Information

Use by either setting:

- environment variable `ASDF_CURRENCY`
- config key `concurrency`

### DIFF
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -134,6 +134,16 @@ Disabling the plugin short-name repository does not remove plugins previously in
 
 :::
 
+### `concurrency`
+
+The default concurrency to use.
+
+| Options                                                                                                 | Description                                                  |
+| :------------------------------------------------------------------------------------------------------ | :----------------------------------------------------------- |
+| integer | Number of cores to use when compiling the source code |
+
+Note that if the environment variable `ASDF_CONCURRENCY` is set, that takes precedence.
+
 ## Environment Variables
 
 Setting environment variables varies depending on your system and Shell. Default locations depend upon your installation location and method (Git clone, Homebrew, AUR).
@@ -172,6 +182,11 @@ The location where `asdf` will install plugins, shims and tool versions. Can be 
 
 - Default: `$HOME/.asdf`
 - Usage: `export ASDF_DATA_DIR=/home/john_doe/.asdf`
+
+### `ASDF_CONCURRENCY`
+
+- Default: Output of `nproc`
+- Usage: `export ASDF_CONCURRENCY=32`
 
 ## Internal Configuration
 
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -151,45 +151,67 @@ Setting environment variables varies depending on your system and Shell. Default
 
 Environment variables should generally be set before sourcing `asdf.sh`/`asdf.fish` etc. For Elvish set above `use asdf`.
 
-The following examples assume:
-
-- an installation to `$HOME/.asdf`
-- a Bash Shell
+The following describe usage with a Bash Shell.
 
 ### `ASDF_CONFIG_FILE`
 
 Path to the `.asdfrc` configuration file. Can be set to any location. Must be an absolute path.
 
-- Default: `$HOME/.asdfrc`
+- If Unset: `$HOME/.asdfrc` will be used.
 - Usage: `export ASDF_CONFIG_FILE=/home/john_doe/.config/asdf/.asdfrc`
 
 ### `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME`
 
 The filename of the file storing the tool names and versions. Can be any valid filename. Typically, you should not set this value unless you want to ignore `.tool-versions` files.
 
-- Default: `.tool-versions`
+- If Unset: `.tool-versions` will be used.
 - Usage: `export ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=tool_versions`
 
 ### `ASDF_DIR`
 
 The location of `asdf` core scripts. Can be set to any location. Must be an absolute path.
 
-- Default: `$HOME/.asdf` (always the parent directory of the `bin/asdf` executable)
+- If Unset: the parent directory of the `bin/asdf` executable is used.
 - Usage: `export ASDF_DIR=/home/john_doe/.config/asdf`
 
 ### `ASDF_DATA_DIR`
 
 The location where `asdf` will install plugins, shims and tool versions. Can be set to any location. Must be an absolute path.
 
-- Default: `$HOME/.asdf`
+- If Unset: `$HOME/.asdf` if it exists, or else the value of `ASDF_DIR`
 - Usage: `export ASDF_DATA_DIR=/home/john_doe/.asdf`
 
 ### `ASDF_CONCURRENCY`
 
-Number of cores to use when compiling the source code. If unset, the asdf config `concurrency` value is used. If set, this environment variable takes precedence over the asdf config `concurrency` value. 
+Number of cores to use when compiling the source code. If set, this value takes precedence over the asdf config `concurrency` value. 
 
-- Default: unset environment variable
+- If Unset: the asdf config `concurrency` value is used.
 - Usage: `export ASDF_CONCURRENCY=32`
+
+## Full Configuration Example
+
+Following a simple asdf setup with:
+
+- a Bash Shell
+- an installation location of `$HOME/.asdf`
+- installed via Git
+- NO environment variables set 
+- NO custom `.asdfrc` file
+
+would result in the following outcomes:
+
+| Configuration                         | Value            | Calculated by                                                                                                                                      |
+| :------------------------------------ | :--------------- | :------------------------------------------------------------------------------------------------------------------------------------------------- |
+| config file location                  | `$HOME/.asdfrc`  | `ASDF_CONFIG_FILE` is empty, so use `$HOME/.asdfrc`                                                                                                |
+| default tool versions filename        | `.tool-versions` | `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` is empty, so use `.tool-versions`                                                                            |
+| asdf dir                              | `$HOME/.asdf`    | `ASDF_DIR` is empty, so use parent dir of `bin/asdf`                                                                                               |
+| asdf data dir                         | `$HOME/.asdf`    | `ASDF_DATA_DIR` is empty so use `$HOME/.asdf` as `$HOME` exists.                                                                                   |
+| concurrency                           | `auto`           | `ASDF_CONCURRENCY` is empty, so rely on `concurrency` value from the [default configuration](https://github.com/asdf-vm/asdf/blob/master/defaults) |
+| legacy_version_file                   | `no`             | No custom `.asdfrc`, so use the [default configuration](https://github.com/asdf-vm/asdf/blob/master/defaults)                                      |
+| use_release_candidates                | `no`             | No custom `.asdfrc`, so use the [default configuration](https://github.com/asdf-vm/asdf/blob/master/defaults)                                      |
+| always_keep_download                  | `no`             | No custom `.asdfrc`, so use the [default configuration](https://github.com/asdf-vm/asdf/blob/master/defaults)                                      |
+| plugin_repository_last_check_duration | `60`             | No custom `.asdfrc`, so use the [default configuration](https://github.com/asdf-vm/asdf/blob/master/defaults)                                      |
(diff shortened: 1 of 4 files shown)


docs/manage/configuration.md
```diff
@@ -134,6 +134,16 @@ Disabling the plugin short-name repository does not remove plugins previously in
...
 
+### `concurrency`
+
+The default concurrency to use.
+
+| Options                                                                                                 | Description                                                  |
+| :------------------------------------------------------------------------------------------------------ | :----------------------------------------------------------- |
+| integer | Number of cores to use when compiling the source code |
+
+Note that if the environment variable `ASDF_CONCURRENCY` is set, that takes precedence.
```
jthegedus, line 145: Note: `ASDF_CONCURRENCY` take precedence if set.

docs/manage/configuration.md
```diff
@@ -134,6 +134,16 @@ Disabling the plugin short-name repository does not remove plugins previously in
 
 :::
 
+### `concurrency`
+
+The default concurrency to use.
+
+| Options                                                                                                 | Description                                                  |
+| :------------------------------------------------------------------------------------------------------ | :----------------------------------------------------------- |
+| integer | Number of cores to use when compiling the source code |
```
jthegedus, line 143: See comment below about defaults.

jthegedus:
Thanks for working on this. I left some comments about handling defaults similar to the rest of the config values.

lib/functions/installs.bash
```diff
@@ -25,7 +25,17 @@ install_command() {
...
 
 get_concurrency() {
-  if command -v nproc >/dev/null 2>&1; then
+  if [ -n "$ASDF_CONCURRENCY" ]; then
+    printf "%s\n" "$ASDF_CONCURRENCY"
+    return
+  fi
+
+  local asdf_concurrency=
+  asdf_concurrency=$(get_asdf_config_value 'concurrency')
```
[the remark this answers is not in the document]
jthegedus: Thinking on this further, `auto` might be the better default value so that we rely on `nproc` and co and only when the user explicitly sets `ASDF_CONCURRENCY` or `concurrency` will we use what they provide.

hyperupcall:
Thank you for your feedback - It's good to be consistent and I'll apply the feedback later today

lib/functions/installs.bash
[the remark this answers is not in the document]
hyperupcall: I like the `auto` special value - agree its good to make it the default

jthegedus:
Ping me in a comment when this is ready for re-review

hyperupcall:
@jthegedus

docs/manage/configuration.md
```diff
@@ -173,6 +184,11 @@ The location where `asdf` will install plugins, shims and tool versions. Can be
 - Default: `$HOME/.asdf`
 - Usage: `export ASDF_DATA_DIR=/home/john_doe/.asdf`
 
+### `ASDF_CONCURRENCY`
+
+- Default: Output of `nproc`, then `sysctl hw.ncpu`, then `/proc/cpuinfo`
```
jthegedus, line 189: The resultant behaviour of this env var with the `concurrency` config value has these defaults, but this env var defaults to an empty env var, no?

jthegedus:
The implementation here looks good. Only a docs nitpick, though I am happy to merge and fix docs later if we don't reach a consensus on the docs nit.

lib/functions/installs.bash
```diff
@@ -25,15 +25,27 @@ install_command() {
...
+      sysctl -n hw.ncpu
+    elif [ -f /proc/cpuinfo ]; then
+      grep -c processor /proc/cpuinfo
+    else
+      printf '%s\n' '1'
+    fi
+  else
+    printf '%s\n' "$asdf_concurrency"
+  fi
+
```
jthegedus, line 48: What do you think about:

```shell
  local asdf_concurrency=

  if [ -n "$ASDF_CONCURRENCY" ]; then
    asdf_concurrency="$ASDF_CONCURRENCY"
  else
    asdf_concurrency=$(get_asdf_config_value 'concurrency')
  fi

  if [ "$asdf_concurrency" = 'auto' ]; then
    if command -v nproc &>/dev/null; then
      asdf_concurrency=$(nproc)
    elif command -v sysctl &>/dev/null && sysctl hw.ncpu &>/dev/null; then
      asdf_concurrency=$(sysctl -n hw.ncpu)
    elif [ -f /proc/cpuinfo ]; then
      asdf_concurrency=$(grep -c processor /proc/cpuinfo)
    else
	    asdf_concurrency="1"
    fi
  fi

  printf '%s\n' "$asdf_concurrency"
```

This way there is a single "return" point in the function?

docs/manage/configuration.md
hyperupcall: Yeah good point - this environment variable defaults to an empty variable. Maybe the code should be modified so that it behaves like other `ASDF_` environment variables and it gets set (internally). Once we do that, we can then also show it on `asdf info`, since I believe it's not printed on there either.

hyperupcall:
lgtm! ^.^

hyperupcall:
I'm looking at the `Default: unset environment variable`, line and it just seems odd to write that here. All of these variables are by default unset externally, except for I believe `ASDF_DIR` and `ASDF_DATA_DIR` on some shells. When I say externally, I mean in the user's environment, in their current shell.

The fact that the environment is unset is an implementation detail because the environment variable is set only within the `asdf` Bash process. So I think writing that it's unset is unhelpful at best and confusing at worst, since it's inconsistent with the other environment variables

jthegedus:
The other env vars are certainly more straightforward compared to this one. These are how they are all handled.

#### `ASDF_CONFIG_FILE`

https://github.com/asdf-vm/asdf/blob/a1e858d2542691adabf9b066add86f16e759a90c/lib/utils.bash#L23-L25

#### `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME`

https://github.com/asdf-vm/asdf/blob/a1e858d2542691adabf9b066add86f16e759a90c/lib/utils.bash#L19-L21

#### `ASDF_DIR`

https://github.com/asdf-vm/asdf/blob/a1e858d2542691adabf9b066add86f16e759a90c/lib/utils.bash#L41-L51

#### `ASDF_DATA_DIR`

https://github.com/asdf-vm/asdf/blob/a1e858d2542691adabf9b066add86f16e759a90c/lib/utils.bash#L27-L39

#### `ASDF_CONCURRENCY` - Proposed

https://github.com/hyperupcall/asdf/blob/da1f5ba0263d9656884b49fe164b0d86cc59e3d2/lib/functions/installs.bash#L27-L49

```shell
get_concurrency() {
  local asdf_concurrency=

  if [ -n "$ASDF_CONCURRENCY" ]; then
    asdf_concurrency="$ASDF_CONCURRENCY"
  else
    asdf_concurrency=$(get_asdf_config_value 'concurrency')
  fi

  if [ "$asdf_concurrency" = 'auto' ]; then
    if command -v nproc &>/dev/null; then
      asdf_concurrency=$(nproc)
    elif command -v sysctl &>/dev/null && sysctl hw.ncpu &>/dev/null; then
      asdf_concurrency=$(sysctl -n hw.ncpu)
    elif [ -f /proc/cpuinfo ]; then
      asdf_concurrency=$(grep -c processor /proc/cpuinfo)
    else
      asdf_concurrency="1"
    fi
  fi

  printf "%s\n" "$asdf_concurrency"
}
```

Most of these don't interact with the `.asdfrc` configuration file. Perhaps we should change the language of all of these env vars docs. Instead of describing the value with "Default: abc" we describe the affect with "If Unset: abc".

Perhaps something like this:

---

## Environment Variables

Setting environment variables varies depending on your system and Shell. Default locations depend upon your installation location and method (Git clone, Homebrew, AUR).

Environment variables should generally be set before sourcing `asdf.sh`/`asdf.fish` etc. For Elvish set above `use asdf`.

The following describe usage with a Bash Shell.

### `ASDF_CONFIG_FILE`

Path to the `.asdfrc` configuration file. Can be set to any location. Must be an absolute path.

- If Unset: `$HOME/.asdfrc` will be used.
- Usage: `export ASDF_CONFIG_FILE=/home/john_doe/.config/asdf/.asdfrc`

### `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME`

The filename of the file storing the tool names and versions. Can be any valid filename. Typically, you should not set this value unless you want to ignore `.tool-versions` files.

- If Unset: `.tool-versions` will be used.
- Usage: `export ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=tool_versions`

### `ASDF_DIR`

The location of `asdf` core scripts. Can be set to any location. Must be an absolute path.

- If Unset: the parent directory of the `bin/asdf` executable is used.
- Usage: `export ASDF_DIR=/home/john_doe/.config/asdf`

### `ASDF_DATA_DIR`

The location where `asdf` will install plugins, shims and tool versions. Can be set to any location. Must be an absolute path.

- If Unset: `$HOME/.asdf` if it exists, or else the value of `ASDF_DIR`
- Usage: `export ASDF_DATA_DIR=/home/john_doe/.asdf`

### `ASDF_CONCURRENCY`

Number of cores to use when compiling the source code. If set, this value takes precedence over the asdf config `concurrency` value. 

- If Unset: the asdf config `concurrency` value is used.
- Usage: `export ASDF_CONCURRENCY=32`

## Full Configuration Example

Following a simple asdf setup in a Bash Shell with an installation location of `$HOME/.asdf` and NO environment variables set would result in the following outcomes:

| Configuration | Value  | Calculated by |
| :----- | :----- | :----- |
| config file location | `$HOME/.asdfrc` | `ASDF_CONFIG_FILE` is empty, so use `$HOME/.asdfrc` |
| default tool versions filename | `.tool-versions` | `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` is empty, so use `.tool-versions` |
| asdf dir | `$HOME/.asdf` | `ASDF_DIR` is empty, so use parent dir of `bin/asdf` |
| asdf data dir | `$HOME/.asdf` | `ASDF_DATA_DIR` is empty so use `$HOME/.asdf` as `$HOME` exists. |
| concurrency | `auto` | `ASDF_CONCURRENCY` is empty, so rely on `concurrency` value from the default configuration ([defaults](https://github.com/asdf-vm/asdf/blob/master/defaults)) |
| legacy_version_file | `no` | from the default configuration ([defaults](https://github.com/asdf-vm/asdf/blob/master/defaults)) |
| use_release_candidates | `no` | from the default configuration ([defaults](https://github.com/asdf-vm/asdf/blob/master/defaults)) |
| always_keep_download | `no` | from the default configuration ([defaults](https://github.com/asdf-vm/asdf/blob/master/defaults)) |
| plugin_repository_last_check_duration | `60` | from the default configuration ([defaults](https://github.com/asdf-vm/asdf/blob/master/defaults)) |
| disable_plugin_short_name_repository | `no` | from the default configuration ([defaults](https://github.com/asdf-vm/asdf/blob/master/defaults)) |

---

@hyperupcall thoughts?

jthegedus:
Merging, any further feedback we can address in a follow up PR :bow:

hyperupcall:
Sounds good, get this in first :+1:

Stratus3D:
I've been out of the loop on this - why does asdf need to even manage concurrency? Why can't plugins just determine this on their own using existing tools outside of asdf?

jthegedus:
>why does asdf need to even manage concurrency

It doesn't, and probably shouldn't, but it has done so for ~7 years.

>Why can't plugins just determine this on their own using existing tools outside of asdf

They can, but the fact it became part of core is an indicator that plugin authors don't want that responsibility. It's the same issue we have with people requesting we provide `sort_versions`, `get_arch` etc type of utility functions to the plugins. The authors don't want to rewrite the same code across multiple plugins.

---

During the initial growth of the project, I don't think there was a clear plan or consensus on how to handle public APIs. People requesting features and raising PRs meant activity and growth of the tool, so maintainers were perhaps not as stringent as they should have been. I think the existence of the plugin repo is another example of this.

I'm happy to delete and remove code and features. But if we're making larger determinations about the plugin API and our intent to remove things, we should also carve out a clear picture of a goal-state we're aiming towards. It would then help justify to plugin authors and contributors our intentions and perhaps ease the pain of removing features.

PR #1533, opened 2023-04-06
fix: enforce consistent shell redirection format

hyperupcall:
# Summary

Make redirections more consistent

### DIFF
--- a/lib/commands/reshim.bash
+++ b/lib/commands/reshim.bash
@@ -11,7 +11,7 @@ remove_shim_for_version() {
   local count_installed
   count_installed=$(list_installed_versions "$plugin_name" | wc -l)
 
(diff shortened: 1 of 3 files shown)


jthegedus:
I have no objection to this change. I would like to get @Stratus3D opinion.

jthegedus:
>@hyperupcall &> isn't POSIX is it? I assume we can only enforce this on Bash files.

I guess if it isn't, we would just want to ensure `checkstyle.py` only asserts/fixes this rule for `.bash` files.

jthegedus:
> so this new rule only applies to our POSIX shell files

@hyperupcall Do you mean non-POSIX? We want to apply this to `.bash` and NOT `.sh`, right?

Sorry, I am getting confused.

hyperupcall:
@jthegedus My apologies, that was me - fixed

Stratus3D:
@jthegedus you can merge whenever you are ready.

PR #1525, opened 2023-03-31
fix!: Remove files containing only `asdf` wrapper functions

hyperupcall:
# Summary

To improve consistency with Nushell, Elvish, and PowerShell Core, this moves the wrapper functions for Fish and Bash/Zsh/POSIX Shell to their respective entrypoints.

PR #1538, opened 2023-04-12
chore: rm format script config for removed files

jthegedus:
# Summary

This updates the `scripts/format.bash` script to no longer try and format files removed in #1525


Stratus3D:
Thanks!

PR #1539, opened 2023-04-13
chore: only ban readlink -f

Stratus3D:
Adding this in preparation for https://github.com/asdf-vm/asdf/pull/667


hyperupcall:
late but lgtm :)

Issue #1542, opened 2023-04-16
bug: extra star in plugin version in fish shell


jpalardy:
### Describe the Bug

if a specific version of a plugin is the current one, it gets flagged with a star:

```bash
> asdf local elixir
*1.14.4-otp-25  system

# tab
> asdf local elixir \*1.14.4-otp-25
version *1.14.4-otp-25 is not installed for elixir

# remove "*" manually
> asdf local elixir 1.14.4-otp-25
# works
```

this affects other completions using the same `__fish_asdf_list_versions` function
- reshim
- global
- local
- shell
- where
- uninstall

### Steps to Reproduce

(plugin version must be pre-selected from `.tool-versions` or manually)

1. type `asdf local`
2. tab select one of the installed plugin
3. tab-select version with `*`
4. press `enter`

### Expected Behaviour

completed version does not contain `*`

### Actual Behaviour

completed version contains extra `*`

### Environment

```shell
OS:
Darwin enoshima 21.6.0 Darwin Kernel Version 21.6.0: Thu Mar  9 20:08:59 PST 2023; root:xnu-8020.240.18.700.8~1/RELEASE_X86_64 x86_64

SHELL:
fish, version 3.6.1

ASDF VERSION:
v0.11.3

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/usr/local/opt/asdf/libexec

ASDF INSTALLED PLUGINS:
elixir                       https://github.com/asdf-vm/asdf-elixir.git master d08f506
erlang                       https://github.com/asdf-vm/asdf-erlang.git master 0463971
golang                       https://github.com/kennyp/asdf-golang.git master f006a12
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master 05f99c3
zig                          https://github.com/cheetah/asdf-zig.git master d88f6db
```

jpalardy:
problem function, ([context](https://github.com/asdf-vm/asdf/blob/master/completions/asdf.fish#L34))

```fish
function __fish_asdf_list_versions -a plugin
    asdf list $plugin 2> /dev/null | sed -e 's/^[[:space:]]*//'
end
```

suggested fix (tested locally)
``` fish
function __fish_asdf_list_versions -a plugin
    asdf list $plugin 2> /dev/null | sed -e 's/^[[:space:]*]*//'
end
# extra "literal" star -----------------------------------^
```

PR #1543, opened 2023-04-17
fix: remove leading asterick in Fish completion

hyperupcall:
# Summary

Fixes: #1542

The recommended fix was good, but this uses only builtins


jthegedus (approved):
Thanks

PR #1544, opened 2023-04-17
fix: rename internal function `asdf_tool_versions_filename`

hyperupcall:
# Summary

Applied suggestions from [here](https://github.com/asdf-vm/asdf/pull/1485#discussion_r1164088836) and [here](https://github.com/asdf-vm/asdf/pull/1518#discussion_r1159820722).

PR #1545, opened 2023-04-17
fix: improve readability of the non-set `nullglob` guard

hyperupcall:
Split from #1544


jthegedus:
Addresses feedback in https://github.com/asdf-vm/asdf/pull/1485#discussion_r1164088836

jthegedus (approved):
Thanks!

Issue #244, opened 2017-10-23
Add a reinstall command


nikolay:
Sometimes, especially during development and troubleshooting, it would be nice to be able to reinstall a version in a similar manner to Homebrew. In its primitive form, it could be just `uninstall` and `install` chain, but take further, it could be done transactionally, too, i.e. a failure to install won't remove the previous instance.

Stratus3D:
Good feature. I like the idea of transactional reinstalls.

Stratus3D:
Should this be a separate command or just a flag to the `install` command as

mig4:
I started looking into this, unfortunately it doesn't look like transactional properties can be achieved without some support from the plugins.

While many of them are simple (download + extract) and this could be done by installing into temp dir, then uninstalling the old one, then moving the temp location into place, there are plugins like [clojure](https://github.com/halcyon/asdf-clojure) which injects the install dir location into some installed scripts (so moving it after installation will make them invalid).

Also a lot of plugins use upstream installers or `./configure && make && make install` so it's hard to tell whether moving directories after install would be safe (most likely it would require passing additional options to the installer, like `DESTDIR` with autotools).

Stratus3D:
I wonder if we could work around this issue by doing something like this:

1. Move the currently installed version and copy it out of it's installed directory and into a temp directory (making it unusable)
2. Tell the plugin to install that same version, with the name of the original directory containing the old version (so that if it succeeds it's in the right spot)
3. If the new version installs successfully delete the temp dir containing the old install. Otherwise delete the directory containing the failed new install (or copy it elsewhere for debugging) and copy the original installation directory back to the original location (restoring the original version and making it usable again).

Would that work? And if so do you think we will run into any problems with this approach? Thanks for taking the time to look into this!

Aethelflaed:
I had to reinstall erlang after installing a few dependencies to compile it again and I can say that at least asdf-erlang wouldn't work this way, as it keeps its own record of which version is installed, so moving the install directory doesn't work.

(Tried on Archlinux).

I saw there are some plugin-test, maybe it would be interesting to add some tests for this feature before shipping it, so those plugins have the time to adapt ?

Stratus3D:
[...]

PR #1546, opened 2023-04-18
Support force (re)installing

ashmckenzie:
# Summary

This PR introduces a new `ASDF_` environment variable called `ASDF_FORCE_INSTALL` which can be set to `true` to force a (re)install.

Taking Ruby as an example, it's desirable to want to re-install (instead of uninstall and install) when you want to apply a new set of Ruby patches, but don't want to lose the currently installed Ruby gems.

### DIFF
--- a/lib/functions/installs.bash
+++ b/lib/functions/installs.bash
@@ -179,7 +179,7 @@ install_tool_version() {
   concurrency=$(get_concurrency)
   trap 'handle_cancel $install_path' INT
 
-  if [ -d "$install_path" ]; then
+  if [ -d "$install_path" ] && [ "${ASDF_FORCE_INSTALL}" != "true" ]; then
     printf "%s %s is already installed\n" "$plugin_name" "$full_version"
   else
 


ashmckenzie:
@Stratus3D Please may I have your thoughts on this approach? I haven't added documentation or updated any required tests as yet as I wanted to see if you would be in support of such an idea.

hyperupcall:
I think by the very least, the reinstall should be implemented with some sort of flag. Going through #244, it seems this has been mentioned [here](https://github.com/asdf-vm/asdf/issues/244#issuecomment-392593804).

Stratus3D:
@ashmckenzie sorry for the late response here. I agree with @hyperupcall that if we support this it should be done via flag for consistency, but I'm not sure this change is sufficient. As mentioned on https://github.com/asdf-vm/asdf/issues/244 if we support this it should be transactional, so if something fails the user doesn't lose there currently installed version (or worse, end up with a broken install).

However asdf has been rewritten in Go (https://github.com/asdf-vm/asdf/releases/tag/v0.16.0) so this PR cannot be merged. I'm going to close it now.

Please continue discussion on https://github.com/asdf-vm/asdf/issues/244. Overall I'm inclined to not add a reinstall command/flag.

PR #1547, opened 2023-04-18
docs: fix uninstall instructions for Fish Shell

vkrol:
# Summary

Fix uninstall instructions for Fish Shell.

## Other Information

The syntax used is not supported in Fish Shell https://stackoverflow.com/questions/75492179/how-to-convert-export-line-from-bash-to-fishvariable-expansion-shortcuts. I'm not sure if the new version is the most correct, but it at least works.


jthegedus (approved):
Thanks!

PR #1549, opened 2023-04-20
chore: unify lint/fix scripts for improved usability locally & in CI

jthegedus:
# Summary

* Consolidate `shellcheck`, `shfmt` & `format` into a single script with flags:
  * `./scripts/lint.bash --check`
  * `./scripts/lint.bash --fix`
  * This script also:
    * enforces that the script is run from the repository root
    * enforces a single flag to be passed
    * Warns for missing tools to format uncommon code when run locally, errors when missing in CI. EG: `fish_indent` is used to format and lint `.fish` files, we can't expect every dev to have every tool, so it is *not* required for local execution. CI still requires this tool to exist.
* simplify CI dependency installation into a `scripts/install_dependencies.bash` file.
  * this helps be consistent with versions of the tools we test against. Dependencies installed in the _lint_ CI workflow were different to the _test_ CI workflow.
* update some file filters for CI filtering with `dorny/paths-filter` as recent changes have affected this filter.
* update `scripts/test.bash` to enforce the script is run from the repository root.
* marked TODO items for linting elvish, nushell & powershell as we should lint and format them as we do Fish with `fish_indent`. A brief look had unclear results for tooling for doing this, so marked as TODO and hopefully people will see these and PR a fix.

## Other Information

Although non-critical this should reduce oversights with CI workflow divergence which is important.

Adding this level of complexity to the `scripts/*.bash` files is annoying, but it should be clear when these aren't working as expected, so hopefully they're not error prone.


jthegedus:
@hyperupcall thoughts?

hyperupcall:
The changes are great!

It's good we directly use the minimum Python version rather than the one bundled with the distribution - this will make sure that things stay compatible.

And with the scripts being moved to their separate files (from the YAML), it makes things significantly easier to read.

Having a single script for formatting and linting makes things easier to check for people less familiar with the codebase. Although I sort of liked having the ability to run each tool separately, the script can always be changed so a flag can be passed to run a specific tool (shellcheck, fish_indent, etc.).

The only thing I think that should probably be changed is:
- In `install_dependencies.bash`, there is a comment about env vars, and variables like `ELVISH_SEMVER` are set. But, those aren't environment variables nor are they marked as env vars later, nor are they passed at environment variables to other processes. So probably best to uncapitalize them
- I guess it could be nice for some scripts can have little `info()` or `error()` functions to make things a bit prettier to read, but I guess that's just personal preference and it doesn't really matter

jthegedus:
>Although I sort of liked having the ability to run each tool separately, the script can always be changed so a flag can be passed to run a specific tool (shellcheck, fish_indent, etc.).

I prefer no options to opt-out of checking everything.

>So probably best to uncapitalize them

Good point :+1: 

>I guess it could be nice for some scripts can have little info() or error() functions to make things a bit prettier to read, but I guess that's just personal preference and it doesn't really matter

Happy to accept changes after merge :)

PR #1552, opened 2023-04-26
feat: sync command (for install plugins and tools from .tool-versions)

xataz:
# Summary

If you install a new computer and asdf, It's really long to add all the plugins and versions of tools you need. Same if a project uses specific versions of tools.
With this feature, you can add all plugins and install all required versions of tools in on single command, from the .tool-versions file.

The name of command is temporary, it will have to be discussed. I chose `sync`, because we synchronize the versions.

## Other Information

Example :
```shell
# Sync local .tool-versions (example in asdf repository)
$ asdf sync # or asdf sync --local
Preparing downloading 1.8.2 from https://github.com/bats-core/bats-core.git
* Downloading Bats-core(version:1.8.2) to /home/mamounet/.asdf/downloads/bats/1.8.2
Cloning into '/home/mamounet/.asdf/downloads/bats/1.8.2'...
[...]
shellcheck-v0.9.0/shellcheck
Downloading shfmt from https://github.com/mvdan/sh/releases/download/v3.6.0/shfmt_v3.6.0_linux_amd64


# Sync global ~/.tool-versions
$ asdf sync --global
nodejs
nodejs 19.8.1 is already installed
bat
bat 0.23.0 is already installed
[...]
yq
yq 4.33.3 is already installed
restic
restic 0.15.2 is already installed
marp-cli
marp-cli 1.5.1 is already installed
```

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -60,6 +60,8 @@ asdf shim-versions <command>            List the plugins and versions that
 asdf update                             Update asdf to the latest stable release
 asdf update --head                      Update asdf to the latest on the master branch
 
+asdf sync [--local | --global]          Add all plugins and install require version
+
 RESOURCES
 GitHub: https://github.com/asdf-vm/asdf
 Docs:   https://asdf-vm.com
--- a/lib/commands/command-plugin-list-all.bash
+++ b/lib/commands/command-plugin-list-all.bash
@@ -1,29 +1,3 @@
 # -*- sh -*-
 
-plugin_list_all_command() {
-  initialize_or_update_plugin_repository
-
-  local plugins_index_path
-  plugins_index_path="$(asdf_data_dir)/repository/plugins"
-
-  local plugins_local_path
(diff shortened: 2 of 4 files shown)


jthegedus:
Hi @xataz 

We have received many PRs and Issues in the past about automatically installing plugins. We do not believe the core tool is in a good place to add this feature yet, if at all. It is something we are considering for the future, but not something we wish to solve this way with the plugin repository.

The crux of the matter is that versions pinning is shared in projects but plugins are pinned to the machine.

`asdf install` will install all the tools once the plugins have been added, so half the functionality of this command already exists.

Other users have written short Shell scripts to automatically install their plugins from a `.tool-versions`, though we don't advise that as you should vet the plugins before installing them. The code of plugins is not considered secure by default.

For these reasons I am closing this PR. Thanks for taking the time to contribute. We have many open issues of bugs which need reproductions and fixes for if you wish to contribute further. Thanks

PR #1555, opened 2023-05-03
fix: Make sure that files with + in their name get escaped

patrickomatic:
# Summary

Hi, I'm working on a program with `++` in the name, similar to `g++` and I noticed that calling it produces this grep error.

```
 grep: repetition-operator operand invalid
```

So included is just a change that escapes the core regexp characters that could occur in a filename


hyperupcall:
This change seems to have no effect on how files with `+` are handled? I change the code to `local key=$2`, run the test suite, and everything passes?

And the code seems to indicate that this is only for keys of the config file (`get_asdf_config_value_from_file`)?

PR #1566, opened 2023-05-30
performance: only create dirs if they don't already exist

mattyo161:
# Summary

When I was doing some debugging the `mkdir -p` was more expensive then checking if the dir exist and then `mkdir` if needed. Discovered while using the `nodejs` plugin, this shaved 50-100 ms in my rough testing.

Fixes: No Issue

## Other Information

Thank you so much for this library, I found this to help my local performance so thought I might share.

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+[[ "${ASDF_DEBUG}" == "1" ]] && set -x || true
 
 # shellcheck source=lib/utils.bash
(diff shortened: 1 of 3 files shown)


hyperupcall:
@mattyo161 Thanks for this improvement - it looks like CI is failing due to some violations. But you only need to worry about the first one because the last two are new and unrelated and fixed in #1571.

jthegedus (approved):
Going to merge this one, still would like some docs in a follow up PR for `ASDF_DEBUG`. Cheers!

PR #1571, opened 2023-06-04
fix: Add ShellCheck directives to find sourced files

hyperupcall:
# Summary

Looking at [CI runs](https://github.com/asdf-vm/asdf/actions/runs/5130651562/jobs/9229710004?pr=1566), it looks like the lack of these directives is causing linting to fail. ~~This suggests that on newer versions, ShellCheck bumped errors like [SC1091](https://www.shellcheck.net/wiki/SC1091) from info to error.~~


jthegedus (approved):
Good catch, thanks!

Issue #1565, opened 2023-05-29
Documenting differences between rtx and asdf


hyperupcall:
I thought it would be useful to document the differences between asdf and other meta version managers (that's what I'm it) like rtx and others that seem to be cropping up recently.

I know this repository is probably not the best place to put it, so I created [version-manager/documents](https://github.com/version-manager/documents) to put this information. For now, it includes tables for [`.tool-versions`](https://github.com/version-manager/documents/blob/main/comparisons/tool-versions.md) compatability and a [general comparison](https://github.com/version-manager/documents/blob/main/comparisons/meta-version-managers.md) across meta version managers. Later, I plan to include scripts to reproduciably measure performance (In the future, I will be making asdf faster so I want to chart improvements). It may also be cool to list languages that each tool supports.

So if anyone (especially maintainers) have any feedback or suggestions or what you would like to see, that would be greatly appreciated! I also opened a similar issue on the rtx repository (https://github.com/jdxcode/rtx/issues/588), for maybe making it easier to talk about stuff.

jdxcode:
I think this is great. A good opportunity to standardize tools and explain differences to users that won't be nearly as familiar as people like us are. It reminds me of https://github.com/shadowspawn/node-version-usage which is a similar useful overview. Another comparison it reminds me of is [this one for chezmoi](https://www.chezmoi.io/comparison-table/).

A couple of random notes/thoughts:

1. rtx should probably be PATH/shim since it can work just as well with shims—the rtx docs don't currently communicate this as well as they should though. It's been a sticking point for folks. I plan to show shims+PATH as 2 methods with their own pros/cons rather than promoting one over the other as it is today.
2. performance sounds incredibly useful, though it might be tough to get a real apples to apples comparison since the tools work differently in a lot of ways. What I would love to see is a chart comparing them over time built automatically in a github action. I'd be more than happy to help contribute to this effort.
3. the `!-<difference>` syntax I am pretty sure I will modify relatively soon. This hasn't happened yet but I think it'll be more like `sub-1:nodejs@18.0.0`, `sub-0.1:nodejs@18.0.0`. What you have is correct today though—and it marks it as experimental so changing it later shouldn't be a big surprise to anyone.

hyperupcall:
[...]

PR #1572, opened 2023-06-07
Initialize new Rust project

Stratus3D:


Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "asdf-core-rust"
+version = "0.1.0"
+edition = "2021"
+
+description = "Extendable version manager with support for Ruby, Node.js, Elixir, Erlang & more - now in Rust!"
```
Stratus3D, line 6: Copied from the Github interface (with an addition). Not sure this will actually get displayed anywhere.

Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+homepage = "https://asdf-vm.com/"
+license = "MIT"
+keywords = ["version-manager", "asdf-vm", "multiple-languages", "shell"]
+categories = ["command-line-interface", "command-line-utilities", "config", "development-tools"]
```
Stratus3D, line 10: I also added these for good measure, but I think they are only used by crates.io, which I'm not sure we will be using.

Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[[bin]]
+name = "asdf"
```
Stratus3D, line 15: Make the default binary name `asdf` instead of `asdf-core-rust`.

Stratus3D:
Opened on wrong repo :(

hyperupcall:
:0 Is this code public? I'd like to track it

Stratus3D:
@hyperupcall not yet, as you can see we're just getting started. It will be made public BEFORE it replaces anything in this repo. We will need a lot of folks to test the new Rust implementation before we would feel comfortable replacing the current Bash implementation.

I opened this PR here by accident because I had my local Git remotes setup wrong for the private fork.

hyperupcall:
I see - sounds cool

hyperupcall:
Yes, see more information at https://github.com/asdf-vm/asdf/issues/1565, https://github.com/version-manager/documents, or the rtx project. They explain things in more detail

PR #1583, opened 2023-06-28
fix(fish)!: don't resolve symlinks for ASDF_DIR

MageJohn:
# Summary

In the fish entry point, only make the ASDF_DIR path absolute, and don't resolve symlinks.

## Other Information

Without this change, a `brew upgrade` causes all the shims to be broken.

When asdf is installed by homebrew, it is placed in a versioned directory and then symlinked to `/usr/local/opt/asdf`. If the symlink is resolved, ASDF_DIR contains the versioned directory. This is a problem because shims use the value of ASDF_DIR to get an absolute path to the asdf executable.

Not resolving symlinks is in line with the `asdf.sh` entry point.

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -1,5 +1,5 @@
 if test -z $ASDF_DIR
-    set ASDF_DIR (realpath (dirname (status filename)))
+    set ASDF_DIR (realpath --no-symlinks (dirname (status filename)))
 end
 set --export ASDF_DIR $ASDF_DIR
 


danhper (approved):
Thanks! Looks good to me!

evanleck:
I'm not sure if this patch caused it but after upgrading I keep getting this output in my shell (fish):

```
realpath: illegal option -- -
usage: realpath [-q] [path ...]
```

MageJohn:
@evanleck I think your right, this probably is the culprit. It looks like, despite there being a fish builtin `realpath` command that supports the `--no-symlinks` flag, it first uses the system `realpath`. If that doesn't support the flag (which is the case on Mac), then there's a problem. I'll open a PR to fix this.

evanleck:
@MageJohn thank you 🙏

PR #1591, opened 2023-07-11
docs: Do not hint that the query for `bin/latest-stable` may be empty

### LINKED ISSUE #1576: Incorrect call signature on bin/latest-stable script
### How can we improve the documentation?

From code: query never be empty as it set default value once it is empty.

https://github.com/asdf-vm/asdf/blob/5b7d0fea0a10681d89dd7bf4010e0a39e6696841/lib/functions/versions.bash#L136

But on document said 

> Call signature from asdf core
> The script should accept a single argument, the filter query.
> 
> "${plugin_path}"/bin/latest-stable ""
> "${plugin_path}"/bin/latest-stable "$query"

ref: https://asdf-vm.com/plugins/create.html#bin-latest-stable

hyperupcall:
# Summary

Fixes #1576

### DIFF
--- a/docs/plugins/create.md
+++ b/docs/plugins/create.md
@@ -278,7 +278,6 @@ Determine the latest stable version of a tool. If absent, the asdf core will `ta
 The script should accept a single argument, the filter query.
 
 ```bash
-"${plugin_path}"/bin/latest-stable ""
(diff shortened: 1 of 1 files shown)


jthegedus (approved):
Thanks

PR #1592, opened 2023-07-11
fix: no longer write temporary files to home directory

### LINKED ISSUE #1556: bug: `asdf global ... system` needs write access to the entire HOME directory when `.tool-versions` already exists there
### Describe the Bug

I removed write access on my $HOME a long time ago, to explicitly allowlist the files / directories required by the tools I use, and to avoid the tools that have weird or unenforcable writing habits.

Asdf looks like it needs write access to the home directory because it uses `sed -i` to modify `.tool-versions` when `.tool-versions` exists
 
```
$ asdf global python system
sed: couldn't open temporary file /home/lestephane/sedWp3RW1: Permission denied
```

I'd expect write access to `.tool-versions` is all asdf needs for me to run this command.

### Steps to Reproduce

```
$ truncate -s 0 $HOME/.tool-versions
$ chmod a-w $HOME
$ asdf global python system
$ asdf global python system
sed: couldn't open temporary file /home/lestephane/sediRUEgd: Permission denied
```

### Expected Behaviour

No error on the second call to `asdf global python system`

### Actual Behaviour

Second call fails

### Environment

```shell
...
ASDF VERSION:
v0.11.3
...
```

N/A

### LINKED ISSUE #1575: bug: Need to set a temporary directory for sed commands when setting global version. (When installed via Nix)
### Describe the Bug

On a nix system the main ASDF static script files are in a read only directory under the nix store which is generated on building the environment. All files that require write permissions and are generated are taken care by XDG and ASDF environment variables storing then under the users home directory..

However, there is one issue that I have found which is setting ASDF global. 
The issue here is that sed wants to create a temporary file under the nix store.

A fix would be to make it use the system's temporary directory and/or add an ASDF environment variable to set it manually 
(maybe .asdfrc aswell)  

### Steps to Reproduce

1. setup nix 
2. install asdf via nixpkgs
3. install asdf-python plugin
4. install python 3.11.3 via asdf
5. asdf global python python 3.11.3

### Expected Behaviour

To set python (or other specified plugin and version ) as the global version

### Actual Behaviour

```
asdf global python 3.11.3
sed: couldn't open temporary file /nix/store/496bar1jwx2a646wnyqzhvqw8iq3pz8i-home-manager-files/sed3ylRgw: Permission denied
```

### Environment

```shell
> asdf info

OS:
Darwin Andrews-MBP 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:53:44 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T8103 arm64

SHELL:
zsh 5.9 (aarch64-apple-darwin21.6.0)

ASDF VERSION:
v0.11.3

ASDF ENVIRONMENT VARIABLES:
ASDF_DATA_DIR=/Users/andrew/.local/share/asdf
ASDF_DIR=/nix/store/hymd5xkvyjr76ppw3pwf2plnzyp0yspa-asdf-vm-0.11.3/share/asdf-vm
ASDF_CONFIG_FILE=/Users/andrew/.config/asdf/asdfrc

ASDF INSTALLED PLUGINS:
nodejs
python
rust

# --------------------------------------------------------

 > nix-info -m
 - system: `"aarch64-darwin”`
 - host os: `Darwin 22.5.0, macOS 13.4`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.13.3`
 - channels(andrew): `””`
 - channels(root): `””`
 - nixpkgs: `/nix/store/w18hyx1dwk4pa591rjyv5g6w0ayf3qcz-source`
# ------------------------------------------------------------------

> echo $TMPDIR
/var/folders/tg/wd1yw6m12mz7vyb1nw2wz9n80000gn/T/

# ------------------------------------------------------------------
> sed --version
sed (GNU sed) 4.9

# ------------------------------------------------------------------

> which sed
/run/current-system/sw/bin/sed

 which is a symbolic link to the nix store 
/run/current-system/sw/bin/sed -> /nix/store/c3g4kbjz47jksvzid2qarxnx8mqdacr1-gnused-4.9/bin/sed

# -------------------------------------------------------------------
The directory Sed tried to write to is in fact root of nix home-managers Applications directory under nix store.  
/nix/store/496bar1jwx2a646wnyqzhvqw8iq3pz8i-home-manager-files/

this contains "Applications/Home manager Apps" which is symbolically linked to the main macOS Applications directory. I have no idea why sed is trying to temporary store data here.
```

seems to be most if not all plugins 
I think its asdf itself.

hyperupcall:
# Summary

No longer write temporary files to home directory.

Fixes #1556
Fixes #1575

### DIFF
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -61,8 +61,15 @@ version_command() {
   done
 
   if [ -f "$file" ] && grep "^$plugin_name " "$file" >/dev/null; then
-    sed -i.bak -e "s|^$plugin_name .*$|$plugin_name ${resolved_versions[*]}|" "$file"
-    rm -f "$file".bak
+    local temp_dir
+    temp_dir=${TMPDIR:-/tmp}
+
+    local temp_tool_versions_file
+    temp_tool_versions_file=$(mktemp "$temp_dir/asdf-tool-versions-file.XXXXXX")
+
+    cp -f "$file" "$temp_tool_versions_file"
+    sed -e "s|^$plugin_name .*$|$plugin_name ${resolved_versions[*]}|" "$temp_tool_versions_file" >"$file"
+    rm -f "$temp_tool_versions_file"
   else
     # Add a trailing newline at the end of the file if missing
     [[ -f "$file" && -n "$(tail -c1 "$file")" ]] && printf '\n' >>"$file"


jthegedus (approved):
Nice solution 👍

PR #1594, opened 2023-07-12
fix: Set default shell version values on POSIX entrypoint

kurochan:
# Summary
Add default values for `BASH_VERSION`, `ZSH_VERSION`, `KSH_VERSION`.

Shell which sets option `nounset` occurs following error.
```bash
$ /opt/homebrew/opt/asdf/libexec/asdf.sh:68: BASH_VERSION: parameter not set
```

### DIFF
--- a/asdf.sh
+++ b/asdf.sh
@@ -8,7 +8,7 @@
 export ASDF_DIR="${ASDF_DIR:-}"
 
 if [ -z "$ASDF_DIR" ]; then
-  if [ -n "$BASH_VERSION" ]; then
+  if [ -n "${BASH_VERSION:-}" ]; then
     # Use BASH_SOURCE[0] to obtain the relative path to this source'd file. Since it's
     # a relative path, 'cd' to its dirname and use '$PWD' to obtain the fullpath.
     # Use 'builtin cd' to ensure user-defined 'cd()' functions aren't called.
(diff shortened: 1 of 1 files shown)


hyperupcall:
Nice catch!

To properly address this, the variables cannot be exported and should not be assigned to because this changes the shell environment.

So the variable substitutions have to be inline each if statement

kurochan:
@hyperupcall 
Thanks for the review. It was indeed causes breaking changes 😱
Fixed at https://github.com/asdf-vm/asdf/pull/1594/commits/ebe7310e08252ea03c110f5328c45e1a6ea78879 .

hyperupcall:
LGTM! Waiting for other maintainers to give feedback / take action

hannestyden:
Happy to see that this fix is available already so I don't have to file it myself. 🙇:blush:

kurochan:
But unfortunately, it has not yet been approved and released 😢

hyperupcall:
I would merge this if I could - the maintainers seem to be merging things less frequently these days. For now, I have cherry-picked this commit (and others from other PRs) into [fox-incubating/asdf](https://github.com/fox-incubating/asdf), one of my forks.

All I can do is release a new version from my fork. I plan to add [other features](https://github.com/fox-incubating/asdf/issues/1) as well.

jthegedus (approved):
Thanks! And thanks for the review @hyperupcall 🙇‍♂️

PR #1597, opened 2023-07-14
fix!: `plugin list` exit code 0 when no plugins are installed

### LINKED ISSUE #903: `asdf plugin list` should return a 0 exit code when no plugins are installed
I'm using ansible to bootstrap a development environment.  Part of this process is installing asdf along with some plugins.  Prior to installing a plugin, I run `asdf plugin list` to see if that plugin is already installed.  When no plugins are installed, `asdf plugin list` will return a non-zero exit code - which ansible recognizes as a failure.  To me, having no plugins and attempting to list all installed plugins should just return an empty list - not a failure code.

It's not a big deal though - the workaround I'm doing is:

```
  - name: get asdf plugin list
    command: 
      cmd: "asdf plugin list"
    register: asdf_plugin_list
    failed_when: asdf_plugin_list.rc != 0 and 'No plugins installed' not in asdf_plugin_list.stderr
```

Buuut, I wouldn't mind throwing up a PR to address this if you're willing!

hyperupcall:
# Summary

Closes #903

I do not consider this a breaking change because the fixed behavior falls into line with existing package managers. If one needs to test if no packages are installed, the better option is to parse standard output.

### DIFF
--- a/lib/functions/plugins.bash
+++ b/lib/functions/plugins.bash
@@ -42,7 +42,7 @@ plugin_list_command() {
     ) | awk '{ if (NF > 1) { printf("%-28s", $1) ; $1="" }; print $0}'
   else
     display_error 'No plugins installed'
-    exit 1
+    exit 0
   fi
 }
 


lib/functions/plugins.bash
```diff
@@ -42,7 +42,7 @@ plugin_list_command() {
     ) | awk '{ if (NF > 1) { printf("%-28s", $1) ; $1="" }; print $0}'
   else
     display_error 'No plugins installed'
```
Tony-Sol, line 44:
```suggestion
    printf "No plugins installed\n"
```
since it's no longer an error, i think it's better to use stdout instead of stderr

hyperupcall: In the future it may be worth changing this depending on if standard output is a tty. But I don' think it's worth changing that right now

jthegedus:
Same reasoning in the other PR. Exit codes and formatted outputs from commands are our public API contract. I would consider this breaking. Happy to discuss in depth in a specific issue about what is breaking vs not, but in the short term I am pretty well convinced that status codes changes are breaking changes.

jthegedus (approved):
Thanks. This probably should have broken a test or two 😅

jthegedus:
Bumped the wrong button, sorry 😅

hyperupcall:
> Same reasoning in the other PR. Exit codes and formatted outputs from commands are our public API contract. I would consider this breaking. Happy to discuss in depth in a specific issue about what is breaking vs not, but in the short term I am pretty well convinced that exit codes changes are breaking changes.

I don't really feel very strongly about it in this particular case, so making this breaking is definitely fine by me

PR #1598, opened 2023-07-14
fix!: exit success when adding an already-added plugin

### LINKED ISSUE #841: `asdf plugin add` should exit code 0 in most cases
**Is your feature request related to a problem? Please describe.**
`asdf plugin add` exits with code 2 if the plugin is already added. This isn't a great user experience for automation. From the perspective of the final result, the operation succeeded.

**Describe the solution you'd like**
Exit with code `0` if the plugin is already installed. Possibly check/warn if the current plugin is outdated?

**Describe similar asdf features and why they are not sufficient**
n/a

**Describe workarounds you've considered**
```
asdf plugin add <name> || true
```

unfortunately this workaround is problematic in that if it fails for a different and legitimate reason, the script will not be aware.

**Additional context**
n/a

hyperupcall:
# Summary

Closes #841 

I do not consider this a breaking change because the fixed behavior does not break currently existing scripts that have the workaround (ie `(exit 2) || :` has the same affect as `(exit 0) || :`)

### DIFF
--- a/lib/functions/plugins.bash
+++ b/lib/functions/plugins.bash
@@ -79,8 +79,8 @@ plugin_add_command() {
   mkdir -p "$(asdf_data_dir)/plugins"
 
   if [ -d "$plugin_path" ]; then
-    display_error "Plugin named $plugin_name already added"
-    exit 2
+    printf '%s\n' "Plugin named $plugin_name already added"
+    exit 0
   else
     asdf_run_hook "pre_asdf_plugin_add" "$plugin_name"
     asdf_run_hook "pre_asdf_plugin_add_${plugin_name}"
--- a/test/plugin_add_command.bats
+++ b/test/plugin_add_command.bats
@@ -97,12 +97,12 @@ teardown() {
   [ "$output" = "dummy" ]
 }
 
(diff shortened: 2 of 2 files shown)


lib/functions/plugins.bash
```diff
@@ -79,8 +79,8 @@ plugin_add_command() {
   mkdir -p "$(asdf_data_dir)/plugins"
 
   if [ -d "$plugin_path" ]; then
-    display_error "Plugin named $plugin_name already added"
-    exit 2
+    printf '%s\n' "Plugin named $plugin_name already added" >&2
```
Tony-Sol, line 82: looks like it should be either without passing to stderr (better imho) or without changes

hyperupcall: Agreed, it would be better to print to standard output here

jthegedus:
I agree with this change.

This is a breaking change as people have scripts dependent on the output here. It was added specifically because people wanted to know the difference of these states.

I am updating the PR title conventional commit to reflect this

```diff
-fix:
+fix!:
```

jthegedus (approved):
Thanks, LGTM. Finally! 😌

PR #1607, opened 2023-07-29
fix: improve lint and test scripts

hyperupcall:
# Summary

- Makes prints semantically dry
- Use physical directory when comparing to see if script is running in `asdf` repository. (Unfortunately `set -o physical` does not work with `$PWD`)
- Other small fixes


jthegedus (approved):
Thanks. Sorry for the delay

PR #1637, opened 2023-09-11
fix(fish): use builtin realpath over system one

MageJohn:
# Summary

Fixes the issue raised in a PR comment here: https://github.com/asdf-vm/asdf/pull/1583#issuecomment-1714212545

Fish treats their `realpath` builtin as a fallback in case a system version doesn't exist. Not all versions of `realpath` support the `--no-symlinks` flag, which `asdf.fish` uses. This PR ensures that the builtin version of `realpath` is always used.

## Other Information

Just as a note for posterity, the `--no-symlinks` flag for the builtin `realpath` was added in fish 3.2.0, released in March 1, 2021. 2 years old is good, but some slower Linux distros may lack support. If that proves to be a problem, then perhaps a different solution can be found. I think long term though the builtin will offer the best support across different systems.

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -1,5 +1,5 @@
 if test -z $ASDF_DIR
-    set ASDF_DIR (realpath --no-symlinks (dirname (status filename)))
+    set ASDF_DIR (builtin realpath --no-symlinks (dirname (status filename)))
 end
 set --export ASDF_DIR $ASDF_DIR
 


cr0t:
These changes also resolve the issue on macOS because system `realpath` doesn't support the given options.

I just updated to the latest `asdf` version 0.13.0 and started to get this error at the beginning of my sessions:

```console
realpath: illegal option -- -
usage: realpath [-q] [path ...]
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
> ~
```

@MageJohn patch fixes the problem.

jthegedus (approved):
Great contribution, thanks!

jthegedus:
You can use this change by pulling the latest code with `asdf update --head`

rebebop:
Any idea when will this be released? `asdf update --head` doesn't work if installed via `brew`.

kblcuk:
@rebebop at least according to github releases, 0.13.1 was release 4 minutes ago, so give it some time :)
https://github.com/asdf-vm/asdf/releases/tag/v0.13.1

cr0t:
@rebebop if you're in hurry, you can edit `asdf.fish` file manually, it's here: `/usr/local/opt/asdf/libexec/asdf.fish`.

Just add `builtin` on the 2nd line to get this:

```fish
if test -z $ASDF_DIR
    set ASDF_DIR (builtin realpath --no-symlinks (dirname (status filename)))
end
# ...
```

rebebop:
> @rebebop at least according to github releases, 0.13.1 was release 4 minutes ago, so give it some time :) [`v0.13.1` (release)](https://github.com/asdf-vm/asdf/releases/tag/v0.13.1)

@kblcuk All good :), when I posted, which was an hour ago, it was obviously not released so was only wondering.
@cr0t Thanks for the snippet

PR #1643, opened 2023-09-17
fix: plugin extension commands to not require `bin/` directory

### LINKED ISSUE #726: Extension commands do not work without a bin directory
**Describe the bug**

The [extension commands plugin api](https://asdf-vm.com/#/plugins-create?id=extension-commands-for-asdf-cli) does not work if the plugin does not have a `bin` directory.

This seems to be related to https://github.com/asdf-vm/asdf/blob/master/bin/asdf#L51 which checks for the existence of the `bin` directory before looking for a plugin command. This will be misleading for plugins that only have extension commands.

Also see #523 

**To Reproduce**

1. Create a new plugin `foo` with a `lib/commands/command.bash` file that echo's something
2. Symlink the plugin directory into the asdf plugins directory as `foo`
3. Run `asdf foo`
4. You'll see an unknown command error

**Expected behavior**

The plugin command is found and works correctly.

**Actual behavior**

Get the error message `Unknown command: asdf foo` + standard help text

### Environment

**OS**: macOS

**asdf version**: 0.7.8

cc @jthegedus

### LINKED ISSUE #729: no Ruby script found in input (LoadError) when asdf called directly
The asdf integration in the JetBrains projects calls asdf.sh directly, and periodically, I'll run into the issue below.

`
➜ . /usr/local/Cellar/asdf/0.7.8/asdf.sh && asdf local ruby 2.5.7 && ruby /Users/eugene.cook/.asdf/shims/rails console
Traceback (most recent call last):
/Users/eugene.cook/.asdf/installs/ruby/2.5.7/bin/ruby: no Ruby script found in input (LoadError)
`
I've tried removing the shims directory completely and triggering a reshim, without success.

### LINKED ISSUE #1365: bug: can no longer store global tool-versions files in subdirectories conveniently
### Describe the Bug

This commit (https://github.com/asdf-vm/asdf/commit/711ad991043a1980fa264098f29e78f2ecafd610) mean that storing global versions files in a sub-directory no longer works in combination with local versions files in the current folder.

e.g. I've always stored all my global tool-versions files in a subdirectory of $HOME called asdf-tool-versions and set `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=asdf-tool-versions/.tool-versions.$1` (where $1 is obviously variable)

This way, `asdf global x y` would update `$HOME/asdf-tool-versions/.tool-versions.x` but `asdf local x y` would create a file called `.tool-versions` in the current directory .

After the above mentioned commit, either every local directory needs to also have a `asdf-tool-versions` sub-directory, or all global configs must be in the root of $HOME.

One option maybe to add ASDF_DEFAULT_GLOBAL_FILE_FOLDER or suchlike, or just revert.

### Steps to Reproduce

1. Create directory in home folder called asdf-tool-versions
2. Export ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=asdf-tool-versions/.tool-versions
3. Change to a directory other than home e.g. /tmp
4. `asdf global x y` should change `$HOME/asdf-tool-versions/.tool-versions`
5. `asdf local x y` results in `/opt/asdf-vm/lib/functions/versions.bash: line 67: /tmp/asdf-tool-versions/.tool-versions.default: No such file or directory`

### Expected Behaviour

`asdf local x y` should change `/tmp/.tool-versions`

### Actual Behaviour

`asdf local x y` results in `/opt/asdf-vm/lib/functions/versions.bash: line 67: /tmp/asdf-tool-versions/.tool-versions.default: No such file or directory`

### Environment

```shell
❯ asdf info
OS:
Linux jaws 5.15.76-1-MANJARO #1 SMP PREEMPT Sat Oct 29 14:22:16 UTC 2022 x86_64 GNU/Linux

SHELL:
zsh 5.9 (x86_64-pc-linux-gnu)

ASDF VERSION:
v0.10.2

ASDF ENVIRONMENT VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=asdf-tool-versions/.tool-versions.default
ASDF_DIR=/opt/asdf-vm
```

Not plugin specific

hyperupcall:
# Summary

This is a bug.

Closes #726
Closes #1365 (duplicate)
Closes #729 (duplicate, solved, stale)

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -48,15 +48,13 @@ find_asdf_cmd() {
 
 find_plugin_cmd() {
   local ASDF_CMD_FILE args_offset
-  if [ -d "$(get_plugin_path "$1")/bin" ]; then
-    local result=
-    result="$(find_cmd "$(get_plugin_path "$1")/lib/commands" "${@:2}")"
-    ASDF_CMD_FILE=${result% *}
-    args_offset=${result##* }
-    if [ -n "$ASDF_CMD_FILE" ]; then
-      args_offset=$((args_offset + 1)) # since the first argument is the plugin name
-      printf "%s %s\n" "$ASDF_CMD_FILE" "$args_offset"
-    fi
+  local result=
+  result="$(find_cmd "$(get_plugin_path "$1")/lib/commands" "${@:2}")"
+  ASDF_CMD_FILE=${result% *}
+  args_offset=${result##* }
+  if [ -n "$ASDF_CMD_FILE" ]; then
+    args_offset=$((args_offset + 1)) # since the first argument is the plugin name
+    printf "%s %s\n" "$ASDF_CMD_FILE" "$args_offset"
   fi
 }
 


jthegedus (approved):
See comments here

PR #1644, opened 2023-09-17
wip

hyperupcall:
# Summary


jthegedus:
Nice.

Since this would run across many plugins, so logging the specific plugin that doesn't support this would be ideal.

hyperupcall:
This is ready to be merged

Issue #1657, opened 2023-09-30
bug: asdf crashes when run from a symlink


cornfeedhobo:
### Describe the Bug

`asdf` makes heavy use of `$(dirname "$(dirname "$0")")` to resolve script location. This fails in the case of being run from a symlinked location.

### Steps to Reproduce

1. `mkdir ~/.local/bin`
2. `export PATH="${HOME}/.local/bin:${PATH}"`
3. `git clone https://github.com/asdf-vm/asdf.git ~/.asdf`
4. `cd ~/.local/bin`
5. `ln -s ~/.asdf/bin/asdf asdf`

### Expected Behaviour

I expect `asdf` to gracefully discover it's real root directory

### Actual Behaviour

```plain
asdf: line 5: /home/cornfeedhobo/.local/bin/lib/utils.bash: No such file or directory
asdf: line 71: asdf_tool_versions_filename: command not found
asdf: line 74: asdf_config_file: command not found
asdf: line 77: asdf_data_dir: command not found
asdf: line 80: asdf_dir: command not found
asdf: line 28: asdf_dir: command not found
asdf: line 113: asdf_dir: command not found
asdf: line 116: /lib/commands/command-help.bash: No such file or directory
```

### Environment

```shell
asdf can't run, so no info right now
```

hyperupcall:
This is not the correct way to install asdf - see the [website](https://asdf-vm.com/guide/getting-started.html#_3-install-asdf) to see how to properly install asdf

hyperupcall:
It would be good to print a warning in this case, even if it's the incorrect instalation procedure

cornfeedhobo:
@hyperupcall I've already got a branch up to fix this. I've seen it in [pyenv](https://github.com/pyenv/pyenv/pull/1697) and [goenv](https://github.com/go-nv/goenv/pull/146). This makes it much easier for distro package maintainers.

Issue #1164, opened 2022-01-20
Add easier shell integration with a subcommand that can be used in eval statement


onnimonni:
### Is your feature request related to a problem? Please describe

There's so many multiple ways to enable asdf for different shells.

For example I'm interested in replacing `nodenv` with `asdf` in the company I'm now working.

In our project readme we have now the following steps:
> Add the following line to your shell's rc/profile file and restart your shell:
> ```eval "$(nodenv init -)"```

Which is much cleaner than checking if people are using bash/zsh/fish separately and providing multiple installation steps like we need with asdf.

### Describe the proposed solution

Add `init` subcommand to asdf which prints out the paths or commands that can be added for corresponding shell:

Example output using fish shell
```
$ nodenv init
# Load nodenv automatically by appending
# the following to ~/.config/fish/config.fish:

status --is-interactive; and source (nodenv init -|psub)
```

Example output using bash:
```
# Load nodenv automatically by appending
# the following to ~/.bash_profile:

eval "$(nodenv init -)"
```

### Describe similar `asdf` features and why they are not sufficient

The third section in [getting started page](https://asdf-vm.com/guide/getting-started.html#_3-install-asdf) is sufficient but this would make the adoption little bit smoother.

### Describe other workarounds you've considered

Pointing people to asdf getting started page from our documentation.

hyperupcall:
@jthegedus @Stratus3D do either (or both) of you have interest in this? I don't want to implement this if there is no desire for it.

hyperupcall:
Tackling this would make asdf easier to package for distros. Related to #1343, #666.

Stratus3D:
I think you make a good point @onnimonni , however, I think this feature should only be implemented in the new Go version (PR https://github.com/asdf-vm/asdf/pull/1805)

PR #1658, opened 2023-09-30
fix: asdf not working when bin is symlinked

cornfeedhobo:
# Summary

This PR seeks to add support for resolving `asdf_dir` when `bin/asdf` is symlinked to a non-standard location.

Fixes: #1657

### DIFF
--- a/bin/asdf
+++ b/bin/asdf
@@ -1,8 +1,47 @@
 #!/usr/bin/env bash
 set -o pipefail
 
+if type -p realpath > /dev/null; then
+  absdir() {
+    local path
+    path="$(realpath "$1")"
+    printf "%s\n" "${path%/*}"
+  }
+elif test -n "$(type -p greadlink readlink | head -1)"; then
+  absdir() {
+    local name path readlink
+    path="$1"
+    readlink="$(type -p greadlink readlink | head -1)"
+    (
+      while [ -n "${path}" ]; do
+        pathdir="${path%/*}"
+        if [ "${pathdir}" != "${path}" ]; then
+          cd "${pathdir}" || exit 1
+        fi
+        name="${path##*/}"
+        path="$($readlink "${name}")"
+      done
+      printf "%s\n" "${PWD}"
+    )
+  }
+else
+  printf "Unable to find readlink or realpath" >&2
+  exit 1
+fi
+
+asdf_dir() {
+  if [ -z "$ASDF_DIR" ]; then
+    local bin_dir root_dir
+    bin_dir="$(absdir "${BASH_SOURCE[0]}")"
+    root_dir="${bin_dir%/*}"
+    printf '%s\n' "$root_dir"
+  else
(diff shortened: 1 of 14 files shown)


hyperupcall:
Thanks for working on a fix! - although these changes fix the `source`ing issues, `asdf` will still break in subtle ways because full setup depends on an `asdf` function loaded into the shell. And with this solution, the user doesn't know that they failed to load the `asdf` function in the shell. Although the errors are fixed, it is misleading because the user will get errors elesewhere.

I think a better solution would be to warn if the user has not sourced `asdf.sh` (or `asdf.fish`, etc.), and exit completely. That will fix the hard to understand `source` errors and give the user direct instructions on how to properly fix the issue

cornfeedhobo:
@hyperupcall continuing the thread here..

> It would be good to print a warning (and exit) in this case

When checking for `realpath` or `readlink` support, I can have the final condition test if the directory has the expected structure or error out saying that symlink support doesn't work without those tools. I kind of already did, but it's simplistic in it's current form.

cornfeedhobo:
> will still break in subtle ways

I don't see how.

- All functions in `asdf` are called by asdf itself, so the subshell is already loaded correctly
- If plugins are currently written to be naive and not use `asdf_dir`, then how would they break from this?
- This doesn't alter the location that plugins are installed or how they are managed, so any logic based on `dirname $0` should still work.

Could you give an example of where you expect this would break? Maybe I can test it out or something?

hyperupcall:
I don't think having the `realpath` or `readlink` code is a good idea. For each shell, we already [have logic](https://github.com/asdf-vm/asdf/blob/master/asdf.sh) to calculate the correct directory paths and stuff. The `realpath` stuff seems extraneous. And I'm not sure directory check you mentioned will catch the case that the user has not sourced the equivalent of `asdf.sh` for their shell. 

Instead, I was thinking of having some `ASDF_INTERNAL_FUNCEXEC` that we always pass into the `bin/asdf.sh` from the shell functions. In `bin/asdf.sh`, if that environment variable is not present, then we print the error. This seems like the most simple solution

hyperupcall:
> I don't see how.

There are various commands that set [an environment variable in the user's current interactive shell](https://github.com/asdf-vm/asdf/blob/4085e5542bac824ea124610ad247c2f90d1b8d93/asdf.sh#L137). Unfortunately, I think this would make your approach a no-go since this case would be impossible to handle since no logic would be ran in the user's interactive shell environment.

cornfeedhobo:
I see. That's truly unfortunate. `asdf`'s internal management of updates and the non-standard loading are the only things blocking adoption of this tool. I keep revisiting this every few years wondering when I'll be able to ditch the `*env` suite I've been using.

hyperupcall:
It is unfortunate - if #1164 is fixed, then perhaps we would move forward with something like this thenceforth.

PR #1682, opened 2023-11-22
fix: make clearer help usage

mrjk:
# Summary

Fixes: help usage is difficult to read and quite bulky

## Other Information

I propose this PR that just reorganize a bit command help message in a more logical way (to me at least). Also, there is more spacing and indentation that make the whole mere readable and understandable. Feel free to comment if you have other opinion on this topic :)

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -1,65 +1,81 @@
-MANAGE PLUGINS
-asdf plugin add <name> [<git-url>]      Add a plugin from the plugin repo OR,
-                                        add a Git repo as a plugin by
-                                        specifying the name and repo url
-asdf plugin list [--urls] [--refs]      List installed plugins. Optionally show
-                                        git urls and git-ref
-asdf plugin list all                    List plugins registered on asdf-plugins
-                                        repository with URLs
-asdf plugin remove <name>               Remove plugin and package versions
-asdf plugin update <name> [<git-ref>]   Update a plugin to latest commit on
-                                        default branch or a particular git-ref
-asdf plugin update --all                Update all plugins to latest commit on
-                                        default branch
+* CONTEXT PACKAGES
+---------------------
+  asdf current                            Display current version set or being
+                                          used for all packages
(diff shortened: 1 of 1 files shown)


hyperupcall:
Currently, the help menu is organized alphabetically, by category. And I do prefer older way of styling the category title (by capitalization). Changing the help menu is a big change, because many are used to the positions of (or memorize) certain commands, and this would disrupt that. In my opinion, I think it should only be changed if there is consensus. But I'm not a maintainer, only a frequent contributor.

mrjk:
Thanks for your feedback :) This commit changes only 4 things:

1) Add dashes after titles
2) Indent content from title
3) Add one extra space before "sub-sections" and titles
4) Reorganize sections order

I feel a good `--help` is the one you don't have to look every time you want to use the program; if the help message is confusing, either people won't be able to use easily the program, either they find something else. Also, reformatting help message is probably the less disruptive change in an app lifecycle, especially if it can improve user comprehension :p

Anyway, I'm completely open to keep only some of these change, but eventually, I would go AT LEAST with point 1 and 2, and check if point 3 or 4 are "too disruptive".

Tell me what fit the best to the community :)

mrjk:
Hi there, should I assume my PR is refused ?

Thanks

hyperupcall:
@mrjk This project isn't maintained as much as it used to be. Sometimes even important PRs take a bit longer to be reviewed. I'm only someone who used to frequently contribute.

Stratus3D:
Hi @mrjk , I'm sorry for the late response here. Thanks for the PR! Unfortunately, asdf has been rewritten in Go (https://github.com/asdf-vm/asdf/releases/tag/v0.16.0) so this PR cannot be applied. I'm going to close it now.

I'm also not sure these changes should be made to the help output regardless the text does not use standardized terminology like "tools", "plugins", and "versions" but rather "packages", which aren't actually a thing in asdf.

Issue #1684, opened 2023-11-24
bug: `asdf latest <plugin>` and `asdf latest --all` are sometimes inconsistent


cj81499:
### Describe the Bug

Using asdf `v0.13.1` (latest at time of writing), `asdf latest <plugin>` and `asdf latest --all` are sometimes inconsistent.

### Steps to Reproduce

```
# install asdf-bazel-buildtools plugin
# use ssh
$ asdf plugin add git@github.com:cj81499/asdf-bazel-buildtools.git
# or use https
$ asdf plugin add https://github.com/cj81499/asdf-bazel-buildtools.git

$ asdf latest bazel-buildtools  
6.1.1

$ asdf latest --all
...
bazel-buildtools        v6.4.0  installed
...

# observe inconsistency
```

### Expected Behaviour

Both commands should report the latest version of bazel-buildtools (`v6.4.0` at time of writing)

### Actual Behaviour

`asdf latest --all` correctly reports the latest version (v6.4.0)

`asdf latest bazel-buildtools` incorrectly reports an old version (6.1.1)

I believe this is due to `asdf latest --all` and `asdf latest <plugin>` using a different default filter. I intend to open a PR to fix shortly.

See https://github.com/cj81499/asdf-bazel-buildtools/issues/6

### Environment

```shell
OS:
Linux cjacobson-dev1 5.4.0-131-generic #147-Ubuntu SMP Fri Oct 14 17:07:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
zsh 5.8 (x86_64-ubuntu-linux-gnu)

BASH VERSION:
5.0.17(1)-release

ASDF VERSION:
v0.13.1-0586b37

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/home/cjacobson/.asdf
ASDF_DIR=/home/cjacobson/.asdf
ASDF_CONFIG_FILE=/home/cjacobson/.asdfrc

ASDF INSTALLED PLUGINS:
...
bazel-buildtools             git@github.com:cj81499/asdf-bazel-buildtools.git main ce6b4b1
...
```

### asdf plugins affected (if relevant)

https://github.com/cj81499/asdf-bazel-buildtools

Stratus3D:
[...]

PR #1685, opened 2023-11-24
fix: make asdf latest output consistent

cj81499:
# Summary

Modify the default query used by `asdf latest` to allow for "v" prefixes.

Additionally, perform a small refactor to extract some duplicated logic to a helper function.

Fixes: #1684

## Other Information

I believe that `asdf latest --all` uses an empty query (it simply doesn't pass a query argument), which is different than `asdf latest <plugin>`.
I think both should use the same default query, but I'm not sure which (`"v?[0-9]"` vs `""`) should be preferred. I'd like guidance on which should be preferred.


cj81499:
I'm having trouble reproducing the test failures locally (in fact, I'm having trouble getting a clean run of the tests on the default branch).

I'd appreciate help getting the tests running/passing.

hyperupcall:
The tests that usually fail are in `update_command.bats`. When I run `bats test`, 5 tests fail out of 368:

```text
update_command.bats
 ✓ asdf update --head should checkout the master branch
 ✗ asdf update should checkout the latest non-RC tag
   (in test file test/update_command.bats, line 41)
     `[ "$status" -eq 0 ]' failed
   Cloning into '/tmp/asdf.UZh0/home/.asdf'...
   done.
 ✗ asdf update should checkout the latest tag when configured with use_release_candidates = yes
   (in test file test/update_command.bats, line 54)
     `[ "$status" -eq 0 ]' failed
   Cloning into '/tmp/asdf.OpSy/home/.asdf'...
   done.
 ✓ asdf update is a noop for when updates are disabled
 ✓ asdf update is a noop for non-git repos
 ✓ asdf update fails with exit code 1
 ✗ asdf update should not remove plugin versions
   (in test file test/update_command.bats, line 85)
     `[ "$status" -eq 0 ]' failed
   Cloning into '/tmp/asdf.RWrB/home/.asdf'...
   done.
 ✗ asdf update should not remove plugins
   (in test file test/update_command.bats, line 95)
     `[ "$status" -eq 0 ]' failed
   Cloning into '/tmp/asdf.dOCk/home/.asdf'...
   done.
 ✗ asdf update should not remove shims
   (in test file test/update_command.bats, line 106)
     `[ "$status" -eq 0 ]' failed
   Cloning into '/tmp/asdf.n4EO/home/.asdf'...
   done.
```

If you get other failures that aren't due to your changes, it might be useful to file an issue about that.

Stratus3D:
Hi @cj81499 thanks for the PR, these look like good changes. I'm sorry I was not able to review sooner. I have rewritten asdf in Golang and I believe I fixed this issue. The regex for latest version is now defined in a single place in the `versions` package - https://github.com/asdf-vm/asdf/blob/3a9912e839b0cc68ce4886a0365eaef3b7f9ab7e/internal/versions/versions.go#L27. I'm going to close this PR as I don't think it makes sense to merge now.

cj81499:
I don't use `asdf` much anymore, so that's fine. fwiw, if this is fixed, you probably want to close #1685 :)

Issue #974, opened 2021-06-17
Need to specify branch name for plugin testing


hickey:
**Describe the bug**

Both GitHub and GitLab are moving away from using the `master` branch name. In the `command_plugin_test.bash` script the branch name is hard coded. [ref](https://github.com/asdf-vm/asdf/blob/79dc43fb5ee2210a5e0284b8a3bfce85d4d39e16/lib/commands/command-plugin-test.bash#L9)

**To Reproduce**

Developing a new plugin for [vcluster](https://github.com/loft-sh/vcluster/) and the repo that I am developing in uses `main` as the primary branch. Creating a new repo and committing any plugin should show the same results of unable to check out `master` when `asdf plugin test PLUGIN URL_TO_REPO` is executed. 

I found #385 and attempted to use `./.git` to see if I could _trick_ the plugin test command to use what is already checked out, but as you can see in the output below, `master` is still attempting to be checked out.

**Expected behavior**

Ability to specify the branch name that plugin test will run against.

**Actual behavior**

Branch name is hard coded and developer is forced into using `master` as the branch name. 

### Environment

```shell
❯ asdf info
OS:
Darwin rebo.local 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:33 PDT 2021; root:xnu-7195.121.3~9/RELEASE_X86_64 x86_64

SHELL:
zsh 5.8 (x86_64-apple-darwin20.0)

ASDF VERSION:
v0.8.0-c6145d0

ASDF ENVIRONMENT VARIABLES:
ASDF_DIR=/Users/hickey/.asdf

ASDF INSTALLED PLUGINS:
aria2                        https://github.com/asdf-community/asdf-aria2.git
bat                          https://gitlab.com/wt0f/asdf-bat.git
direnv                       https://github.com/asdf-community/asdf-direnv.git
dive                         https://github.com/looztra/asdf-dive.git
dyff                         https://gitlab.com/wt0f/asdf-dyff.git
eksctl                       https://github.com/elementalvoid/asdf-eksctl.git
fd                           https://gitlab.com/wt0f/asdf-fd.git
helm                         https://github.com/Antiarchitect/asdf-helm.git
julia                        https://github.com/rkyleg/asdf-julia.git
k9s                          https://github.com/looztra/asdf-k9s.git
kind                         https://github.com/johnlayton/asdf-kind.git
kubectl                      https://github.com/Banno/asdf-kubectl.git
[...]

PR #1694, opened 2023-12-11
fix: `plugin test` git-ref default to the default branch of the plugin's repository

javiergarea:
# Summary

It implements the strategy followed by #800 to retrieve the plugin's default branch instead of relying on master.

Fixes: #974

### DIFF
--- a/lib/commands/command-plugin-test.bash
+++ b/lib/commands/command-plugin-test.bash
@@ -14,7 +14,8 @@ plugin_test_command() {
   local plugin_url="$2"
   shift 2
 
-  local plugin_gitref="master"
+  plugin_remote_default_branch=$(git ls-remote --symref $plugin_url HEAD | awk '{ sub(/refs\/heads\//, ""); print $2; exit }')
+  local plugin_gitref=${3:-${plugin_remote_default_branch}}
   local tool_version
   local interpret_args_literally
   local skip_next_arg
--- a/lib/commands/command-plugin-test.bash
+++ b/lib/commands/command-plugin-test.bash
@@ -14,8 +14,20 @@ plugin_test_command() {
   local plugin_url="$2"
   shift 2
 
-  plugin_remote_default_branch=$(git ls-remote --symref $plugin_url HEAD | awk '{ sub(/refs\/heads\//, ""); print $2; exit }')
-  local plugin_gitref=${3:-${plugin_remote_default_branch}}
+  local exit_code
+  local TEST_DIR
+
+  fail_test() {
+    printf "FAILED: %s\n" "$1"
+    rm -rf "$TEST_DIR"
+    exit 1
+  }
+
+  if [ -z "$plugin_name" ] || [ -z "$plugin_url" ]; then
+    fail_test "please provide a plugin name and url"
+  fi
+
+  local plugin_gitref
   local tool_version
   local interpret_args_literally
   local skip_next_arg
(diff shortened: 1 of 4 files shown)


lib/commands/command-plugin-test.bash
```diff
@@ -14,7 +14,8 @@ plugin_test_command() {
   local plugin_url="$2"
   shift 2
 
-  local plugin_gitref="master"
+  plugin_remote_default_branch=$(git ls-remote --symref $plugin_url HEAD | awk '{ sub(/refs\/heads\//, ""); print $2; exit }')
```
hyperupcall, line 17:
```suggestion
  plugin_remote_default_branch=$(git ls-remote --symref "$plugin_url" HEAD | awk '{ sub(/refs\/heads\//, ""); print $2; exit }')
```

hyperupcall (commented):
I'm not a maintainer, only a frequent contributor, ~~but your changes look good to me~~ thanks for the patch - I noticed an error with it:

This code changes the behavior in cases where `plugin_name` and `plugin_url` are empty. Instead of a nice error message, a weird `git` error will show instead. Since this code executes git and is now a point of failure (ex. if `plugin_url` is empty), this code should idealy be placed [at least after this error check](https://github.com/asdf-vm/asdf/blob/ca87d66c66da61201c4573a8e1a58707d9a8df44/lib/commands/command-plugin-test.bash#L62) or right before `plugin_gitref` is used.

Another thing to note, is that there is [another place in the code with similar functionality](https://github.com/asdf-vm/asdf/blob/27f7ef78529649534b8383daa58e4b370b3cbd7f/lib/functions/plugins.bash#L135), and this sort of thing could be extracted into it's own function. That said, I'm not sure if that is necessary right now, and this kind of thing can be tracked in #1697 anyways.

lib/commands/command-plugin-test.bash
javiergarea: Thank you so much for your review!

Would it be OK if I just move the check to be executed before?
```diff
diff --git a/lib/commands/command-plugin-test.bash b/lib/commands/command-plugin-test.bash
index 16f5c1d..0cb2659 100644
--- a/lib/commands/command-plugin-test.bash
+++ b/lib/commands/command-plugin-test.bash
@@ -14,6 +14,19 @@ plugin_test_command() {
   local plugin_url="$2"
   shift 2
 
+  local exit_code
+  local TEST_DIR
+
+  fail_test() {
+    printf "FAILED: %s\n" "$1"
+    rm -rf "$TEST_DIR"
+    exit 1
+  }
+
+  if [ -z "$plugin_name" ] || [ -z "$plugin_url" ]; then
+    fail_test "please provide a plugin name and url"
+  fi
+
   plugin_remote_default_branch=$(git ls-remote --symref $plugin_url HEAD | awk '{ sub(/refs\/heads\//, ""); print $2; exit }')
   local plugin_gitref=${3:-${plugin_remote_default_branch}}
   local tool_version
@@ -50,19 +63,6 @@ plugin_test_command() {
     set -- "${SHELL:-sh}" -c "$1"
   fi
 
-  local exit_code
-  local TEST_DIR
-
-  fail_test() {
-    printf "FAILED: %s\n" "$1"
-    rm -rf "$TEST_DIR"
-    exit 1
-  }
-
-  if [ -z "$plugin_name" ] || [ -z "$plugin_url" ]; then
-    fail_test "please provide a plugin name and url"
-  fi
-
   TEST_DIR=$(mktemp -dt asdf.XXXX)
   cp -R "$(asdf_dir)/bin" "$TEST_DIR"
   cp -R "$(asdf_dir)/lib" "$TEST_DIR"
(END)
```

hyperupcall: If it were me, I would move the `plugin_gitref` assignment to right after the `case` statement because I think the argument parsing logic is more readable when it is at the top of the file - it would also be consistent with the other `command-*` files, I believe.

javiergarea: I see your point. Nevertheless, I'm finding it challenging to maintain the suggested ordering elegantly. The current setup with the `plugin_gitref` assignment first setting the default value aligns with the logic of the case statement, which overwrites it when a value for `--asdf-plugin-gitref` is provided. Any insights on how to approach it?

hyperupcall: How about checking if it `plugin_gitref` is empty with `[ -z`? That way, if no value or an empty value is supplied, we fallback to calculating it on our own

javiergarea: Makes sense to me. I commited some changes. Let me know if this version fits :)

hyperupcall: That works - the new changes look good to me

jthegedus (approved):
Thanks all for raising and reviewing :pray:

Issue #1629, opened 2023-09-04
bug: fish path set-up breaks fish_add_path


tjvc:
### Describe the Bug

The fish [set-up script](https://github.com/asdf-vm/asdf/blob/master/asdf.fish) sets `fish_user_paths` as a global variable. This breaks the expected behaviour of this variable, which is a universal variable by default. [Universal variables persist across fish sessions; global variables are specific to the current session](https://fishshell.com/docs/current/language.html#variable-scope). Global variables, when present, [take precedence over universal variables of the same name](https://fishshell.com/docs/current/faq.html#why-doesn-t-set-ux-exported-universal-variables-seem-to-work). The impact of this is that users calling `fish_add_path /some/path` unexpectedly find that the assignment is not persisted across sessions.

This has previously been raised as an [issue](https://github.com/fish-shell/fish-shell/issues/9819) on fish itself, and noted as a [comment](https://github.com/asdf-vm/asdf/issues/1346#issuecomment-1570051540) on another issue in this repo, but I think it's deserving of its own issue.

### Steps to Reproduce

1. Install the asdf fish set-up script
2. Start a fish shell
3. Update `fish_user_paths` with `fish_add_path /some/path`
4. Start another fish shell

### Expected Behaviour

The path assignment is persisted across sessions.

### Actual Behaviour

The path assignment is only applied in the current session.

### Environment

```shell
OS:
Darwin XX379KM0VT 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 arm64

SHELL:
fish, version 3.6.0

BASH VERSION:
3.2.57(1)-release

ASDF VERSION:
v0.12.0

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/thom/.asdf
ASDF_DIR=/opt/homebrew/Cellar/asdf/0.12.0/libexec
ASDF_CONFIG_FILE=/Users/thom/.asdfrc

ASDF INSTALLED PLUGINS:
nodejs                       https://github.com/asdf-vm/asdf-nodejs.git master c9e5df4
python                       https://github.com/danhper/asdf-python.git master 8505457
ruby                         https://github.com/asdf-vm/asdf-ruby.git master 16bc8ac
[...]

PR #1699, opened 2023-12-22
fix: use universal scope for fish_user_paths in fish shell integration

bakotaco:
# Summary

This PR addresses issue described in #1629. Currently the Fish shell setup script sets fish_user_paths as a global variable.  This conflicts with the expected universal nature of fish_user_paths, leading to path assignments not persisting across sessions when using fish_add_path.

Fixes:
- fish-shell/fish-shell#9819
- #1629

## Other Information

This ensures that any paths added via fish_add_path are retained across all Fish shell sessions, aligning with the standard behavior of fish_add_path as outlined in its [documentation](https://fishshell.com/docs/current/tutorial.html).

### DIFF
--- a/asdf.fish
+++ b/asdf.fish
@@ -13,10 +13,10 @@ end
 # Do not use fish_add_path (added in Fish 3.2) because it
 # potentially changes the order of items in fish_user_paths
 if not contains $_asdf_bin $fish_user_paths
-    set --global --prepend fish_user_paths $_asdf_bin
+    set --universal --prepend fish_user_paths $_asdf_bin
 end
 if not contains $_asdf_shims $fish_user_paths
-    set --global --prepend fish_user_paths $_asdf_shims
+    set --universal --prepend fish_user_paths $_asdf_shims
 end
 set --erase _asdf_bin
 set --erase _asdf_shims


jthegedus (approved):
Looks good to me, thanks!

jthegedus:
And of course when I have the time to go through PRs for the first time this year GitHub "experiences issues". Will merge tomorrow. Cheers!

jthegedus:
There seem to be some other fish issues, but we can deal with them separately :+1:

hyperupcall:
This MR is a breaking change in cases where a user uses `asdf` in a script, as shown by the failing tests. Because `--universal-` persists the variable modification, there needs to be better isolation for the fish test as well.

jthegedus:
> This MR is a breaking change in cases where a user uses `asdf` in a script, as shown by the failing tests. Because `--universal` persists the variable modification, there needs to be better isolation for the fish test as well, which wasn't done either.

Yeah, I would like to get this fixed before performing a new release. Help welcome as my environment is currently unable to run fish without significant effort :sweat_smile:

bakotaco:
apologies, but after talking to @faho (long time fish contributor) I came to the conclusion this was actually an incorrect fix... I created a followup PR with a proper fix and explanation: #1709

PR #1704, opened 2024-01-03
Slightly nicer messages

toothbrush:
# Summary

I realise i didn't ask permission before opening this PR, so i'm happy to close it or remove bits and pieces as folks see fit.

Having said that, i'm addressing some very small paper cuts i have when using asdf.

* Make it so the output of `asdf latest foo` can directly `>> .tool-versions`
* Tell me _how_ to add a missing plugin – it's infrequent enough that i keep forgetting how

Let me know what you think! 🙌

### DIFF
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -209,6 +209,7 @@ latest_all() {
     done
   else
     printf "%s\n" 'No plugins installed'
+    printf "%s\n" 'Try: asdf plugin add PLUGIN'
   fi
   exit 0
 }
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -165,7 +165,7 @@ latest_command() {
     fi
   fi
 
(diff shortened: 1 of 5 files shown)


lib/functions/versions.bash
```diff
@@ -165,7 +165,7 @@ latest_command() {
     fi
   fi
 
-  printf "%s\n" "$versions"
+  printf "%s %s\n" "$plugin_name" "$versions"
```
jdx, line 168: this is a breaking change

toothbrush: Yes, if folks are using the output of `foo=$(asdf latest bar)` in their scripts, that would break.  I'd be happy to add a conditional, with a new configurable option in `~/.asdfrc` like, say, `asdf_latest_include_plugin_name = yes` which defaults to the old behaviour, unless someone opts in?

(of course with a nicer name than `asdf_latest_include_plugin_name` 😅)

PR #1713, opened 2024-01-23
feat: Make `asdf latest` more ergonomic by outputting plugin name

toothbrush:
# Summary

I have a friction point: I frequently find myself wanting to query the latest version of a plugin, and subsequently pinning it in my current directory.

At the moment, i have to manually munge the output of `asdf latest`, because it gives just the version string, but i need to add the plugin's name as a prefix.

Maybe a shell session will illustrate my friction point:

Previously:

```
$ asdf latest ruby 2
2.7.8
$ asdf latest ruby 2 > .tool-versions
$ # ... pain / manual editing / argh
$ asdf install
ruby 2.7.8 installed
```

After this change:

```
$ echo "version_with_plugin_name = yes" > ~/.asdfrc
$ asdf latest ruby 2
ruby 2.7.8
$ asdf latest ruby 2 > .tool-versions
$ asdf install
ruby 2.7.8 installed
$ # no pain!
```

## Other Information

By default, the behaviour of `asdf latest` is unchanged.  However, i've introduced a config option `version_with_plugin_name` & documentation to opt into my proposed new behaviour.  This should make the transition seamless and non-breaking for those that want it.  I'm happy to modify the name of the config option to something clearer.

This is an improved version of #1704.

### DIFF
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -65,6 +65,7 @@ always_keep_download = no
 plugin_repository_last_check_duration = 60
 disable_plugin_short_name_repository = no
 concurrency = auto
+version_with_plugin_name = no
 ```
 
 ### `legacy_version_file`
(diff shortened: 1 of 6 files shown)


toothbrush:
Hm, i've just realised that the `latest_command` function is used internally in asdf as well... So probably this approach isn't great after all.  That's a shame.

I have also learned that `asdf local ruby latest:2` would do almost what i want – the only issue there is that it doesn't automatically install missing versions but just exits with an error.

I think i'll close this PR and think about a different approach.  Sorry for the noise.

Issue #216, opened 2017-08-08
Add command to install latest stable version of a tool


Stratus3D:
Right now it's up to plugins to decide how they list versions. Some list only stable releases, others may list all releases. Some may print out the versions sorted incorrectly. This prevents us from just doing `asdf list-all "$tool" | tail -1` to get the latest stable version of a tool.

We need a command that allows us to install the latest stable version of a tool. In order to have such a command we will need to have a `latest-stable` callback script in each plugin that returns the latest stable version of a tool.  Then we can add a  `--latest` flag to the install command so we can do `asdf install ruby --latest` to install the version of Ruby returned from the Ruby plugin's `latest-stable` callback. The callback script would need to return the latest stable version of the tool for the given architecture. Some plugins have different versions available for different architectures, so if a plugin does include the `latest-stable` it must chose to the right version for the architecture.

This `latest-stable` callback script would start out optional, but could eventually be made a requirement by adding an assertion to the `plugin-test` command.

Created when reviewing https://github.com/thoughtbot/laptop/pull/502

prasannavl:
Once the referred issues are resolved, what would be nice is to have an package cleanup command, that retains only the latest minor versions by default, but still allowing the cleanup for major versions from an input flag, and also optionally a way to hold certain packages manually. 

While versioning semantics may differ, asdf could just follow semantic versioning, with optional plugin API that can be used to override this behaviour for custom semantics by the plugins.

This should provide the fully functionality needed to maintain a sort of a rolling release model.

NightMachinary:
Has any progress happened since 2017?

quintrino:
I have a small method I've been using in a ruby file that might be of help to you. 

```
def latest(language)
  `asdf plugin-add #{language} 2> /dev/null`
  versions = `asdf list-all #{language} 2> /dev/null`.split("\n")
  stock_versions = versions.select { |v| v.count('a-zA-Z').zero? }
  latest_version = stock_versions.max_by { |v| Gem::Version.new(v) }
  `asdf install #{language} #{latest_version}`
end
```

So far it's proven functional for Ruby, Rust, Node, Postgres, Elixir and Elm, but other languages may break it. 

(This is slightly modified from the original that I use which just returned the version number but should when run in a ruby file/executable install the latest version)

NightMachinary:
@quintrino Thanks! I adapted that into a script:
```ruby
#!/usr/bin/env ruby
def latest(language)
[...]

PR #1714, opened 2024-01-23
feat: Auto-install missing versions in asdf local.

toothbrush:
# Summary

I frequently want to be able to upgrade my `.tool-versions` straight to the newest version of a plugin with the `asdf local` command.  At the moment, the first repository i perform this on, i'll get an error that the latest version isn't installed:

```shellsession
$ asdf local ruby latest
version 3.3.0 is not installed for ruby ❌
```

Instead of having to read the latest version from the error output, then type, e.g., `asdf install ruby 3.3.0`, then redo my command, i'd like to be able to have `asdf local` do the installation automatically.  This will save me a tedious manual step.

With this change, i can do it all in one go:

```shellsession
$ echo "local_command_auto_install_missing_version = yes" >> .asdfrc
$ asdf local ruby latest
...installs Ruby 3.3.0...
$ cat .tool-versions
ruby 3.3.0 🎉
```

## Other Information

Apologies for the PR spam.  Hopefully this one will be acceptable.  I previously misunderstood a few things about how asdf works under the hood – i think this is a better way to address what i actually wanted, namely less manual version swizzling 😁 

Also, i'm happy to find a better name for the config option.

### DIFF
--- a/docs/manage/configuration.md
+++ b/docs/manage/configuration.md
@@ -65,6 +65,7 @@ always_keep_download = no
 plugin_repository_last_check_duration = 60
 disable_plugin_short_name_repository = no
 concurrency = auto
+local_command_auto_install_missing_version = no
 ```
 
 ### `legacy_version_file`
@@ -151,6 +152,17 @@ The default number of cores to use during compilation.
 | `auto`  | Calculate the number of cores using `nproc`, then `sysctl hw.ncpu`, then `/proc/cpuinfo` or else `1` |
 
 Note: the environment variable `ASDF_CONCURRENCY` take precedence if set.
+
+### `local_command_auto_install_missing_version`
+
+Configure the `asdf local` command with the ability to auto-install missing versions.
+
+For example, if you want your `.tool-versions` file to reflect `ruby 2.7.8`, you can invoke `asdf local ruby 2.7.8`.  By default, if Ruby 2.7.8 is not yet installed, this command will exit with an error. However, if you set `local_command_auto_install_missing_version = yes` in `~/.asdfrc`, Ruby 2.7.8 will be downloaded & installed if necessary, and then `.tool-versions` will be updated.
+
+| Options                                                    | Description                                                |
+| :--------------------------------------------------------- | :--------------------------------------------------------- |
+| `no` <Badge type="tip" text="default" vertical="middle" /> | Only update .tool-versions if version is already available |
+| `yes`                                                      | Install version if missing, then update .tool-versions     |
 
 ### Plugin Hooks
 
--- a/lib/commands/command-local.bash
+++ b/lib/commands/command-local.bash
@@ -2,5 +2,9 @@
 
 # shellcheck source=lib/commands/version_commands.bash
 . "$(dirname "$ASDF_CMD_FILE")/version_commands.bash"
+# shellcheck source=lib/commands/reshim.bash
+. "$(dirname "$ASDF_CMD_FILE")/reshim.bash"
+# shellcheck source=lib/functions/installs.bash
+. "$(dirname "$(dirname "$0")")/lib/functions/installs.bash"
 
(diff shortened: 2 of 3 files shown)


feld:
The maintainer won't add automatic plugin installs because they're disillusioned into thinking this is a security threat, so this -- which is a logical extension of the idea -- is probably going to be rejected. So many of us have wanted this for years.

https://github.com/asdf-vm/asdf/issues/276

HashNuke:
@feld The context for #216 and this issue are different. Issue-216 was auto-installing plugins. This one is for auto-installing the latest version if missing.

HashNuke (changes requested):
Thank you for the PR. Notes below.

* Can we add appropriate tests for this change? The relevant file to do so would be `test/version_commands.bats`
* Can we please update this to use a command-line option instead of the configuration file?

```shell
asdf local ruby latest --install-missing
```

I would prefer that the user clearly knows that they are forcing an install.

I haven't tested the PR. I'll do that after the changes are done.

toothbrush:
>     * Can we add appropriate tests for this change? The relevant file to do so would be `test/version_commands.bats`
> 
>     * Can we please update this to use a command-line option instead of the configuration file?

Hey Akash,  thanks for your response!  Indeed i'm happy to add tests and a command-line flag if you prefer that approach.  Before doing the work i think it'd be good to discuss the UX though.

My sense is that if we provide both `--install-missing` and a config file option, that caters to people who want lots of control, as well as people like me who decide they always want auto-installation of a latest version of a known plugin.  If we have a CLI flag _instead_ of the config file option, i believe we will not have improved the UX at all.

Consider, before the change:

```
$ asdf local ruby latest
version 3.3.0 is not installed for ruby ❌
# Ah, whoops, i guess 3.3.0 is released, i'll install it manually:
$ asdf install ruby 3.3.0
...installs Ruby 3.3.0...
# Rerun previous command:
$ asdf local ruby latest
...now it works, .tool-versions correctly updated 👍
```

If we add only a `--install-missing` flag:

```
$ asdf local ruby latest
version 3.3.0 is not installed for ruby ❌
# Ah, whoops, i guess 3.3.0 is released, i'll install it -- what was the flag again? 🤔
$ asdf local ruby latest --install-missing
...installs Ruby 3.3.0, .tool-versions correctly updated 👍
```

It's true that this is only 2 commands versus 3 separate actions before, but arguably (my personal view, i guess others may disagree) it will actually be harder to remember this `--install-missing` flag you have to type out only sometimes.  Personally i'd probably not use it and just do the old-fashioned 3-command incantation instead.

Does my concern make sense?  From my point of view there's not much value in adding only a CLI flag without support for setting it persistently in `.asdfrc`.  I'd happily do the work to support both styles, though -- i think you're right that a CLI flag would be good to have.  After all, if it's disabled-by-default and preserving current behaviour, that seems like we're maximally empowering users to make their own choice about auto-installation of latest versions.

> I would prefer that the user clearly knows that they are forcing an install.

That's a good point.  We could consider adding some output such as `Latest ruby version is 3.3.0, which not installed.  Installing now...` in the case that the auto-install flag is set?

Stratus3D:
Sorry for the late reply here. Thanks for the PR @toothbrush ! Unfortunately, asdf has been rewritten in Go (https://github.com/asdf-vm/asdf/releases/tag/v0.16.0) so this PR cannot be applied. I'm going to close it now. 

Additionally, the `asdf local` and `asdf global` commands has been replaced by an `asdf set` command in v0.16.0, so this feature will have to be reworked for asdf 0.16.0. I'm not opposed to adding this feature, but I think we should go with @HashNuke's example of a ` --install-missing` flag for this behavior. I think `asdf set  --install-missing` would sufficiently cover the use cases you have in mind. 

@toothbrush can you open up a new feature request issue and link to this PR?

Issue #1711, opened 2024-01-21
bug: getting "You have configured asdf to preserve downloaded files" when I don't


Mellbourn:
### Describe the Bug

Every time I install a new version using asdf, I always get the warning "You have configured asdf to preserve downloaded files"
To start with I had no `.asdfrc` file, so that was strange.
Then I even created a `.asdfrc` file with an explicit setting `always_keep_download = no` but the warning persists.

Content of my `$HOME/.asdfrc` file:
```
legacy_version_file = yes
always_keep_download = no
```

### Steps to Reproduce

1. install something, e.g. 
```
> asdf install direnv 2.31.0
```

### Expected Behaviour

The following output only:
```
∗ Downloading and installing direnv...
The installation was successful!
```

### Actual Behaviour

I always get two rows of warnings at the end of any installation:
```
∗ Downloading and installing direnv...
The installation was successful!
asdf: Warn: You have configured asdf to preserve downloaded files (with always_keep_download=yes or --keep-download). But
asdf: Warn: the current plugin (direnv) does not support that. Downloaded files will not be preserved.
```

### Environment

```shell
OS:
Darwin Klass-MacBook-Pro-16-2023.local 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:55:06 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6020 arm64

SHELL:
zsh 5.9 (x86_64-apple-darwin23.0)

BASH VERSION:
5.2.26(1)-release

ASDF VERSION:
v0.14.0-ccdd47d

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/klas.mellbourn/.asdf
ASDF_DIR=/Users/klas.mellbourn/.asdf
ASDF_CONFIG_FILE=/Users/klas.mellbourn/.asdfrc
```

### asdf plugins affected (if relevant)

ASDF INSTALLED PLUGINS:
bundler                      git@github.com:jonathanmorley/asdf-bundler.git master a461ac4
direnv                       git@github.com:asdf-community/asdf-direnv.git master a2219c2
java                         git@github.com:halcyon/asdf-java.git master fc28b48
lazydocker                   git@github.com:comdotlinux/asdf-lazydocker.git master abb6f71
nodejs                       git@github.com:asdf-vm/asdf-nodejs.git master c5b7c40
[...]

PR #1721, opened 2024-02-19
fix: keep_download condition

Birdie0:
# Summary

Swap directory and argument checks so it won't result in misleading warning.

Fixes: #1711

## Other Information

~~Also replaced `[ ! a = b ]` with `[ a != b ]` as it's more readable.~~ Inverted condition for better readability.

### DIFF
--- a/lib/functions/installs.bash
+++ b/lib/functions/installs.bash
@@ -237,11 +237,11 @@ install_tool_version() {
       # Remove download directory if --keep-download flag or always_keep_download config setting are not set
       if [ -d "$download_path" ]; then
         always_keep_download=$(get_asdf_config_value "always_keep_download")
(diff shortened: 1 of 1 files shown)


hyperupcall:
Thanks - I'll take a look at this soon. I made the original PR and want things to pass the additional test cases I added in an offline branch

Birdie0:
@hyperupcall Any luck?

hyperupcall:
@Birdie0 Unfortunately, I can no longer boot up the computer that had those test cases. Until I am able to diagnose and fix it, I won't be able to check.

But your changes look good to me and if anything, is likely to be more correct than the existing logic anyways. I don't think my personal issues is blocking your PR. I'm not a maintainer or anything, but these days it seems more difficult to get PRs in, which is why I have closed many of my existing PRs.

Birdie0:
resolved by #1756

PR #1722, opened 2024-02-19
test: unset ASDF_DATA_DIR and ASDF_CONFIG_FILE during test

bootleq:
# Summary

The purpose of this small change is very similar to 6d8d77f which unset `ASDF_DIR`.

In user's login shell with asdf, `ASDF_DATA_DIR` and `ASDF_CONFIG_FILE` will be set / exported when calling shimmed commands, in test workflow the command is `bats`.

For quick reproduce, login with ASDF_DIR set to `~/.asdf` and having shimmed `bats` command, then run tests:
- `bats test/uninstall_command.bats --filter 'uninstall command executes configured pre hook'`
- `bats test/list_command.bats --filter 'list_command with plugin should list installed versions'`

will fail due to aforementioned problem.


chrisjpalmer:
+1 I was just running unit tests... had this exact problem. Great PR

Stratus3D (approved):

Stratus3D:
Sorry for the late review here. Looks great @bootleq ! I've encountered this issue myself, never had a chance to fix it.

Issue #929, opened 2021-04-29
install order should follow the order in .tool-versions


joshlartz:
**Describe the bug**

Some plugins depend on others to be installed first. Despite being expressed that way in the .tool-versions file, asdf alphabetizes the install regardless. 

This feels like it should be the default behavior, but at the very least there should be a flag for it. 

**To Reproduce**

```shell
# .tool-versions
python 3.9.4
awscli 2.2.0
```
`asdf install`

**Expected behavior**

The .tool-versions file is installed top to bottom.

**Actual behavior**

The installation happens in alphabetical order.

> asdf install
> * Downloading awscli release 2.2.0...
> ~/.asdf/installs/awscli/2.2.0 ~
> .asdf/plugins/awscli/bin/../lib/utils.bash: line 86: python: command not found
> .asdf/plugins/awscli/bin/../lib/utils.bash: line 87: ./venv/bin/activate: No such file or directory
> .asdf/plugins/awscli/bin/../lib/utils.bash: line 88: pip: command not found
> .asdf/plugins/awscli/bin/../lib/utils.bash: line 89: pip: command not found
> .asdf/plugins/awscli/bin/../lib/utils.bash: line 90: pip: command not found
> .asdf/plugins/awscli/bin/../lib/utils.bash: line 91: deactivate: command not found
> ~
> awscli 2.2.0 installation was successful!
> Downloading python-build...

### Environment

```shell
OS:
Darwin C02WQ018HTDD 20.4.0 Darwin Kernel Version 20.4.0: Fri Mar  5 01:14:14 PST 2021; root:xnu-7195.101.1~3/RELEASE_X86_64 x86_64

SHELL:
zsh 5.8 (x86_64-apple-darwin20.0)

ASDF VERSION:
v0.8.0
```

jthegedus:
> @jthegedus not sure if this should be another issue, but in the regards of awscli plugin, it uses a global version of python instead of a local version, is there a way to use a locally defined tool inside a plugin?

It depends on the plugin implementation. Ideally, each tool installed is a standalone, pre-packaged binary. If it's not, the plugin may build the tool from source (tools with git-ref installation support likely do this).

As I understand, AWS CLI v2+ ships it's own version of python in pre-built packages. So either the plugin needs to handle installation of v2+ different to v1 and no longer rely on the global.

As for

>is there a way to use a locally defined tool inside a plugin?

[...]

PR #1723, opened 2024-02-22
feat: install dependencies in .tool-versions order

chrisjpalmer:
# UPDATE

## Summary

This PR changes the behaviour of `asdf install` to install tools in the order they are listed in `.tool-versions`.

The purpose is to prevent installation failures due to dependencies between plugins. An example is the `poetry` `python` plugins. When running `asdf install` for these, the `poetry` plugin will typically fail because python is not available ([see here](https://github.com/asdf-vm/asdf/issues/929#issuecomment-1181820200)). This requires the user to manually run `asdf install python xxx` first and then continue with the rest of the install.

Rather than maintain the correct plugin order as part of the tool, this responsibility can be forwarded to users of the tool via `.tool-versions` file.  `asdf install` will respect the order listed in `.tool-versions` when installing the tools.

## Other Information

1. Tests have been written to verify the new behaviour works. All the old tests are passing locally except for two, but they seem unrelated.
    - `asdf update --head should checkout the master branch`
    - `plugin_list_all should sync repo when check_duration set to 0`
3. The behaviour of the tool is to start in the deepest directory (which is the cwd) and recurse upwards to shallower directories.
    - install tools in the `.tool-versions` file one by one
    - install any tools from legacy file versions if a plugin is installed for one, and it didn't appear in `.tool-versions` file
    - continue to recurse to shallower directories until a tool for every plugin has been installed

## Example

Given the following `.tool-versions`:
```
nodejs 18.16.1
python 3.10.13
poetry 1.4.2
```

**before**
```
asdf install

kubectl 1.24.16 is already installed
nodejs 18.16.1 is already installed
No preset version installed for command python3
Please install a version by running one of the following:

asdf install python 3.10.13

or add one of the following versions in your config file at /Users/xxx/workspace/service-developer-portal/.tool-versions
python 3.10.2

Cleanup: Something went wrong!

48 /Users/xxx/.asdf/plugins/poetry/bin/install: POETRY_HOME=$install_path python3 - --version "$version" $flags
```

**after**
```
asdf install

nodejs 18.16.1 is already installed
python-build 3.10.13 /Users/xxx/.asdf/installs/python/3.10.13
python-build: use openssl from homebrew
python-build: use readline from homebrew
Downloading Python-3.10.13.tar.xz...
-> https://www.python.org/ftp/python/3.10.13/Python-3.10.13.tar.xz
Installing Python-3.10.13...
python-build: use readline from homebrew
python-build: use zlib from xcode sdk
Installed Python-3.10.13 to /Users/xxx/.asdf/installs/python/3.10.13
asdf: Warn: You have configured asdf to preserve downloaded files (with always_keep_download=yes or --keep-download). But
asdf: Warn: the current plugin (python) does not support that. Downloaded files will not be preserved.
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

It will add the `poetry` command to Poetry's bin directory, located at:

/Users/xxx/.asdf/installs/poetry/1.4.2/bin

You can uninstall at any time by executing this script with the --uninstall option,
and these changes will be reverted.

Installing Poetry (1.4.2): Done

Poetry (1.4.2) is installed now. Great!

To get started you need Poetry's bin directory (/Users/xxx/.asdf/installs/poetry/1.4.2/bin) in your `PATH`
environment variable.

Add `export PATH="/Users/xxx/.asdf/installs/poetry/1.4.2/bin:$PATH"` to your shell configuration file.

Alternatively, you can call Poetry explicitly with `/Users/xxx/.asdf/installs/poetry/1.4.2/bin/poetry`.

You can test that everything is set up by executing:

`poetry --version`

Configuring poetry to behave properly with asdf ...
Running: "poetry config virtualenvs.prefer-active-python true".

('Configuration file exists at /Users/xxx/Library/Application Support/pypoetry, reusing this directory.\n\nConsider moving TOML configuration files to /Users/xxx/Library/Preferences/pypoetry, as support for the legacy directory will be removed in an upcoming release.',)
asdf: Warn: You have configured asdf to preserve downloaded files (with always_keep_download=yes or --keep-download). But
asdf: Warn: the current plugin (poetry) does not support that. Downloaded files will not be preserved.
Either specify a tool & version in the command
OR add .tool-versions file in this directory
or in a parent directory
```

# ... old PR description

## Summary

This PR allows `asdf` to install tools in the order they are listed in `.tool-versions`. It works by the command `asdf install --ordered-install` to be provided.

The purpose is to prevent installation failures due to dependencies between plugins. An example is the `poetry` `python` plugins. When running `asdf install` for these, the `poetry` plugin will typically fail because python is not available ([see here](https://github.com/asdf-vm/asdf/issues/929#issuecomment-1181820200)). This requires the user to manually run `asdf install python xxx` first and then continue with the rest of the install.

The PR takes a simple approach to resolve this problem. Rather than maintain the correct plugin order as part of the tool, which is much too difficult, allow developers to maintain the order in`.tool-versions` file. Then make the `asdf` tool follow that order when `--ordered-install` is provided.

## Other Information

I have not written any tests, nor have I done extensive testing to verify this doesn't break any old functionality.
I also have not addressed this change for multiple levels of `.tool-versions` files in higher directories.
I can look into these things, if people like this change and we think its worth pursuing. Do let me know.

Our company uses `asdf` extensively for streamlining developer workflows so we are interested in improving the tool if possible.

Thanks!


jfly (commented):
Excited to see this pr! Why even add the option? I'd vote for just making this the unconditional behavior.

HashNuke:
@chrisjpalmer Thanks for this very useful PR + the notes on the tests and the scenarios not handled.

Ordered install can be the default behaviour.

But I have not given any thought to how we could handle order across `.tool-versions` files (especially `$HOME/.tool-versions` and `${pwd}/.tool-versions`). Open to ideas.

HashNuke (changes requested):
Added notes in the comments but repeating here.

Very useful PR ❤️

1. Ordered install can be the default behaviour
2. Open to ideas to handle scenarios for situations where there is a .tool-versions in the $HOME and the current directory.
3. Tests can be added once we do the above.

HashNuke:
Changed this to a draft PR since this is work in progress.

chrisjpalmer:
Thanks Akash. This sounds good.

I have some initial thoughts on how we can support multiple `.tool-versions` files in higher directories.
First I wanted to confirm the current behvaviour. 

### Current Process
Based on some testing and reverse engineering the code, I can see that ASDF does the following steps:


  1. Check `$PWD/.tool-versions` to see if there are any versions who don't have a corresponding plugin installed on the system. If there are any versions without a corresponding plugin, a message is issued and the command exits early.
     - _Note: Higher level directories with `.tool-versions` are not searched._ 
  2. Iterate plugins in alpha order. For each plugin:
     - Recursively search `.tool-versions` files for a corresponding version - starting with `$PWD/.tool-versions`, going up a directory each time.
     - As soon as a match is found, install that version and move on to the next plugin
     - If no match is found, search the `.tool-versions` file pointed to by `$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` .


Let me know if you agree with the above analysis. Assuming I'm right, then the key takeaway is that the `.tool-versions` files are recursively searched and the default `.tool-versions` file is considered last.

### New Process
To emulate the same behaviour whilst now considering the version order in `.tool-versions` I believe we need the following  process:

  1. Check `$PWD/.tool-versions` to see if there are any versions who don't have a corresponding plugin installed... **(keep this the same)**
  2. Recursively scan `.tool-versions` files starting with `$PWD/.tool-versions`, going up a directory each time. For each `.tool-versions` file:
      - Install versions in the order they appear in the file. 
      - Store the plugins we have installed in a variable to keep track of what we have already covered off - this allows us to skip version installs for same plugin versions as we traverse higher directories.
  3. Search ends when one of two conditions is true:
      - there are no more directories to search (this assumes we have already searched `$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` as well)
      - we have satisfied every plugin that is installed on the system 
  
Just for sanity checking its important to note the following about this new process:
  - versions in earlier `.tool-versions` take precedence over ones in higher directories - this is the same as the previous process which resolves the version for each plugin in earlier `.tool-versions` files before searching higher
  - versions without a corresponding plugin will not contribute to the installation process. This is the same as the previous process because, the previous installation process is driven by the list of plugins that _do exist_ and hence will also ignore versions that have no corresponding plugin.

---

Let me know what you think :) Chris

lib/functions/installs.bash
```diff
@@ -85,6 +85,9 @@ install_local_tool_versions() {
   local tool_versions_path
   tool_versions_path=$(find_tool_versions)
 
+  local ordered_install
+  ordered_install=$1
+
```
chrisjpalmer, line 90: can remove this

lib/functions/installs.bash
```diff
@@ -117,30 +121,146 @@ install_local_tool_versions() {
...
+  display_debug "install_directory_tools_recursive '$search_path': entered with plugins_installed='$plugins_installed'"
+
+  while [ "$search_path" != "/" ]; do
+    # install tools from files in current directory
+    display_debug "--------------------------------------------------------------------------------------------------------------"
+    tools_installed=$(install_directory_tools "$search_path" "$plugins_installed" "$tools_installed")
+    display_debug "install_directory_tools_recursive '$search_path': install_directory_tools returned tools_installed='$tools_installed'"
+
+    # terminate if all tools are installed
+    if [[ -n $(stringlist_same_length "$plugins_installed" "$tools_installed") ]]; then
```
chrisjpalmer, line 153: perhaps the validation should be tools_installed contains all entries in plugins_installed... this would be a bit stronger

lib/functions/installs.bash
```diff
@@ -117,30 +121,146 @@ install_local_tool_versions() {
...
+    fi
+
+    # install the version
+    display_none $(install_tool_version "$plugin_name" "$plugin_version")
+    tools_installed=$(echo "$tools_installed $plugin_name" | awk '{$1=$1};1')
+
+    display_debug "install_directory_tools '$search_path': installed '$plugin_name':'$plugin_version' new state of tools_installed='$tools_installed'"
+  done <<< $tool_versions
+
+  # TODO when should we resolve tool version from the environment ?
```
chrisjpalmer, line 211: I believe you should be able to resolve the tool version from the environment... yet to add this feature

chrisjpalmer:
Hi there. I have actually gone ahead and implemented this. Current status is:
- basic code is in place
- replaced old install process with ordered install process
- no unit tests written
- hand tested on my machine with `.tool-version` files
- have not tested legacy install files

chrisjpalmer:
Hey @HashNuke just checking in to see if you are happy with the progress on this PR ?

chrisjpalmer:
### Update
I ran the unit test suite.. I have one unit test to fix: 
- `install_command installs multiple tool versions when they are specified in a .tool-versions file `

I added a new unit test which verifies the ordered install behaviour:
- `install_command installs plugins in order`

Note: these are also failing but I believe they are unrelated:
- `asdf update --head should checkout the master branch`
- `plugin_list_all should sync repo when check_duration set to 0`

### TODO
- Fix the failing unit test

### For consideration
I was chatting with a colleague of mine, and we were thinking that in addition to having tools be installed in .tool-version order, it might be good to consider the directory structure as well when installing tools. Suppose you have the following structure:

```
/
  - .tool-versions: python xxx
       /subdir
          - .tool-versions: poetry yyy
 ```

Currently in both the previous and new install process, `poetry` will be installed before `python` because `asdf install` installs tools in the deepest directory and the recurses upwards to shallower directories. However, there is an argument that it might be more useful to do this the opposite way. In a monorepo for example a user may specify a general tool in the root directory, and then sub directories could be free to declare additional tools. It may be possible that those additional tools have a dependency on tools declared in the root directory. Therefore it could make sense to actually modify `asdf install`'s behaviour to recurse from the shallower directory downwards.

Thoughts on this ?

Thanks

**Comment by chrisjpalmer** [comment]
Hi there guys... I think this is ready for review now :)

File: lib/functions/installs.bash [hunk 4]
```diff
@@ -117,30 +117,199 @@ install_local_tool_versions() {
...
+  # still got some tools to install...
+  # lets see if $ASDF_DEFAULT_TOOL_VERSIONS_FILENAME as actually
+  # an absolute path
+  if [ -f "$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME" ]; then
+    display_debug "--------------------------------------------------------------------------------------------------------------"
+    display_debug "attempting to treat \$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME as an absolute path: $ASDF_DEFAULT_TOOL_VERSIONS_FILENAME"
+    tools_installed=$(_install_directory_tools "" "$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME" "$plugins_installed" "$tools_installed")
+    display_debug "install_directory_tools_recursive '$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME': install_directory_tools returned tools_installed='$tools_installed'"
+  fi
+
```
chrisjpalmer, line 168: there is no test for this, however this is behaviour that was in asdf already. I know because people at my work were using it like this.

chrisjpalmer:
@HashNuke or @jfly would you mind taking a look to confirm you are happy with progress ?

Stratus3D:
>  1. Check $PWD/.tool-versions to see if there are any versions who don't have a corresponding plugin installed on the system. If there are any versions without a corresponding plugin, a message is issued and the command exits early.
>           Note: Higher level directories with .tool-versions are not searched.

If this is the case I think this might be a bug in the old process (probably out of scope for this PR).

> Recursively scan .tool-versions files starting with $PWD/.tool-versions, going up a directory each time. For each .tool-versions file:
>
>    Install versions in the order they appear in the file.
>    Store the plugins we have installed in a variable to keep track of what we have already covered off - this allows us to skip version installs for same plugin versions as we traverse higher directories.

For the new process we might actually want to start at the uppermost .tool-versions file, installing versions in order, and work our way down to the current directory. I feel like the home/root dir is typically used for system wide stuff that the user just happens to be using asdf for (ie version doesn't vary between dirs). This is definitely a nitpick. No need to implement right now.

> versions without a corresponding plugin will not contribute to the installation process. This is the same as the previous process because, the previous installation process is driven by the list of plugins that do exist and hence will also ignore versions that have no corresponding plugin.

Interesting, I wonder if that is best. I don't think we should be doing anything until all required plugins (ie all those referenced in all .tool-version files at and above the current dir) are installed. Again, definitely out of scope for this PR but I think it may have ended up this way by accident rather than by design. Any idea why we might want to silently ignore missing plugins at this step?

lib/functions/installs.bash
Stratus3D: The docs are clear this is only the **filename** of the file - https://asdf-vm.com/manage/configuration.html#asdf-default-tool-versions-filename

It is not an absolute path to a file and shouldn't be treated like it is here (lines 180-181 should be all that's needed). Was there any existing code that treated it like this?

lib/functions/installs.bash
```diff
@@ -117,30 +117,199 @@ install_local_tool_versions() {
...
+install_directory_tools() {
+  local search_path=$1
+  local plugins_installed=$2
+  local tools_installed=$3
+  display_debug "install_directory_tools '$search_path': starting install. tools_installed='$tools_installed'"
+
+  # install tools from .tool-versions
+  # install order is the order listed in .tool-versions
+  file_name=$(asdf_tool_versions_filename)
+  tools_installed=$(_install_directory_tools "$search_path" "$file_name" "$plugins_installed" "$tools_installed")
```
Stratus3D, line 181: I'm used to wrapping everything in quotes. Is there a reason the `$()` on these two lines aren't quoted? Like this:

```
file_name="$(asdf_tool_versions_filename)"
```

lib/functions/installs.bash
```diff
@@ -117,30 +117,199 @@ install_local_tool_versions() {
...
+  local tools_installed=$4
+
+  local tool_versions
+  if ! [[ -f "$search_path/$file_name" ]]; then
+    display_debug "_install_directory_tools '$search_path': exiting early... $file_name file not found"
+    printf "%s\n" "$tools_installed"
+    return 0
+  fi
+
+  tool_versions=$(strip_tool_version_comments "$search_path/$file_name" | awk '{$1=$1};1')
```
Stratus3D, line 207: Probably want this `$()` to be quoted right?

lib/functions/installs.bash
```diff
@@ -117,30 +117,199 @@ install_local_tool_versions() {
...
+
+    # go up a directory
+    search_path=$(dirname "$search_path")
+  done
+
+  # still got some tools to install...
+  # lets see if $ASDF_DEFAULT_TOOL_VERSIONS_FILENAME as actually
+  # an absolute path
+  if [ -f "$ASDF_DEFAULT_TOOL_VERSIONS_FILENAME" ]; then
+    display_debug "--------------------------------------------------------------------------------------------------------------"
```
Stratus3D, line 163: Maybe create a `display_debug_hr` function so we don't have to hardcode this long string in multiple places?

lib/functions/installs.bash
```diff
@@ -253,3 +422,57 @@ install_tool_version() {
...
+
+display_debug() {
+  if [[ $DEBUG = "true" ]]; then
+    printf "debug: %s\n" "$1" >&2
+  fi
+}
+
+display_none() {
+  printf ""
+}
```
Stratus3D, line 478: Might be good to put these in `utils.sh`.

Stratus3D (commented):
Thanks for the PR @chrisjpalmer ! Overall I think this is a good change. 

Having been away from Bash for a little while I'm a little paranoid of unquoted variables and don't remember exactly when things should be quoted or unquoted. I'll test out this PR branch on my machine today and see how it goes.

lib/functions/installs.bash
chrisjpalmer: happy to remove this.

lib/functions/installs.bash
chrisjpalmer: No reason. Will wrap it in quotes no problem

lib/functions/installs.bash
chrisjpalmer: when I tested asdf on of our company repos I noticed that we were setting `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME` to an absolute path and it was working with asdf. Then I tested my new change and it obviously stopped working. It seems like it could have been one of those "accident by design" things.

lib/functions/installs.bash
chrisjpalmer: I am no expert on bash and you could be right. I just found a lot of the time in installs.bash, there aren't many lines with `="$(...)"`. Should they be quoted ?

lib/functions/installs.bash
```diff
@@ -117,30 +117,199 @@ install_local_tool_versions() {
...
+_install_directory_tools_legacy() {
+  local search_path=$1
+  local plugins_installed=$2
+  local tools_installed=$3
+
+  display_debug "_install_directory_tools_legacy '$search_path': resolving legacy files"
+  local plugin_name
+  for plugin_name in $plugins_installed; do
+    # skip if plugin is installed already
+    if [[ -n $(stringlist_contains "$tool_versions" "$plugin_name") ]]; then
```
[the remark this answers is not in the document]
chrisjpalmer: Thanks fixed this now.

lib/functions/installs.bash
chrisjpalmer: Done

lib/functions/installs.bash
chrisjpalmer: Done

chrisjpalmer:
Thanks @Stratus3D for your review. Appreciate that time you put into it. I have tried to address all your comments and respond to anything that I still wasn't sure about.

> > 1. Check $PWD/.tool-versions to see if there are any versions who don't have a corresponding plugin installed on the system. If there are any versions without a corresponding plugin, a message is issued and the command exits early.
> >    Note: Higher level directories with .tool-versions are not searched.
> 
> If this is the case I think this might be a bug in the old process (probably out of scope for this PR).

Quite possibly. We can always follow this PR up with another one which hardens this behaviour.

> 
> > Recursively scan .tool-versions files starting with $PWD/.tool-versions, going up a directory each time. For each .tool-versions file:
> > Install versions in the order they appear in the file.
> > Store the plugins we have installed in a variable to keep track of what we have already covered off - this allows us to skip version installs for same plugin versions as we traverse higher directories.
> 
> For the new process we might actually want to start at the uppermost .tool-versions file, installing versions in order, and work our way down to the current directory. I feel like the home/root dir is typically used for system wide stuff that the user just happens to be using asdf for (ie version doesn't vary between dirs). This is definitely a nitpick. No need to implement right now.
> 
> > Currently in both the previous and new install process, poetry will be installed before python because asdf install installs tools in the deepest directory and the recurses upwards to shallower directories. However, there is an argument that it might be more useful to do this the opposite way. In a monorepo for example a user may specify a general tool in the root directory, and then sub directories could be free to declare additional tools. It may be possible that those additional tools have a dependency on tools declared in the root directory. Therefore it could make sense to actually modify asdf install's behaviour to recurse from the shallower directory downwards.
> 
> Yep, exactly!

I would be happy to chase this up in a follow up PR.

> 
> > versions without a corresponding plugin will not contribute to the installation process. This is the same as the previous process because, the previous installation process is driven by the list of plugins that do exist and hence will also ignore versions that have no corresponding plugin.
> 
> Interesting, I wonder if that is best. I don't think we should be doing anything until all required plugins (ie all those referenced in all .tool-version files at and above the current dir) are installed. Again, definitely out of scope for this PR but I think it may have ended up this way by accident rather than by design. Any idea why we might want to silently ignore missing plugins at this step?

Ah I think when I said that "a version without a corresponding plugin will not contribute to the installation process" what I meant was, ignoring the sanity check that runs before installation kicks off, the main installation loop ignores unknown plugins. However there is a sanity check that runs before the main installation which lets the user know if it detects an unknown plugin. This is only in place for the `.tool-versions` file in the current directory. Again suggests to me a follow up PR to harden this behavior.

lib/functions/installs.bash
Stratus3D: Typically if the value produced by the expression contains spaces and you want it to stay as a string you should quote - `string_with_spaces="$(echo "string with spaces")"`. I think that is the case here, but I may be wrong. Generally, quoting is safer in Bash (unless you are dealing with an array, in which case it won't work).

lib/functions/installs.bash
Stratus3D: Thanks!

lib/functions/installs.bash
```diff
@@ -117,30 +117,189 @@ install_local_tool_versions() {
...
-        some_tools_installed='yes'
-        plugin_versions=$(cut -d '|' -f 1 <<<"$plugin_version_and_path")
-        for plugin_version in $plugin_versions; do
-          install_tool_version "$plugin_name" "$plugin_version"
-        done
-      fi
-    done
+    plugins_installed=$(printf "%s" "$plugins_installed" | tr "\n" " " | awk '{$1=$1};1')
+    display_debug "install_local_tool_versions: plugins_installed='$plugins_installed'"
+    tools_installed=$(install_directory_tools_recursive "$search_path" "$plugins_installed")
```
Stratus3D, line 124: I think the `$()` on lines 122 and 124 should be quoted.

lib/functions/installs.bash
```diff
@@ -117,30 +117,189 @@ install_local_tool_versions() {
...
+  local search_path=$1
+  local plugins_installed=$2
+  local tools_installed=""
+
+  display_debug "install_directory_tools_recursive '$search_path': entered with plugins_installed='$plugins_installed'"
+
+  while [ "$search_path" != "/" ]; do
+    # install tools from files in current directory
+    display_debug_hr
+    tools_installed=$(install_directory_tools "$search_path" "$plugins_installed" "$tools_installed")
```
Stratus3D, line 145: quote this one too I think.

lib/functions/installs.bash
```diff
@@ -117,30 +117,189 @@ install_local_tool_versions() {
...
+
+    # terminate if all tools are installed
+    if [[ -n $(stringlist_a_subset_of_b "$plugins_installed" "$tools_installed") ]]; then
+      display_debug "install_directory_tools_recursive '$search_path': exiting because all known tools ($plugins_installed) are installed ($tools_installed)"
+      printf "%s\n" "$tools_installed"
+      return 0
+    fi
+
+    # go up a directory
+    search_path=$(dirname "$search_path")
```
Stratus3D, line 156: This one as well. paths can contain spaces but they should be treated like a single string.

lib/functions/installs.bash
```diff
@@ -117,30 +117,189 @@ install_local_tool_versions() {
...
+  # install order is the order listed in .tool-versions
+  file_name=$(asdf_tool_versions_filename)
+  tools_installed=$(_install_directory_tools "$search_path" "$file_name" "$plugins_installed" "$tools_installed")
+
+  # install tools from legacy version files
+  # install order is plugin order which is alphabetical
+  local legacy_config
+  legacy_config=$(get_asdf_config_value "legacy_version_file")
+  if [ "$legacy_config" = "yes" ]; then
+    tools_installed=$(_install_directory_tools_legacy "$search_path" "$plugins_installed" "$tools_installed")
```
Stratus3D, line 178: Quote `$()` here too I think.

lib/functions/installs.bash
```diff
@@ -117,30 +117,189 @@ install_local_tool_versions() {
...
+      display_debug "_install_directory_tools '$search_path': '$plugin_name' is already installed... skipping"
+      continue
+    fi
+
+    # install the versions
+    local plugin_version
+    for plugin_version in "${parts[@]:1}"; do
+      # install the version
+      display_none "$(install_tool_version "$plugin_name" "$plugin_version")"
+      tools_installed=$(printf "%s %s" "$tools_installed" "$plugin_name" | awk '{$1=$1};1')
```
Stratus3D, line 222: Quote `$()` here too I think.

lib/functions/installs.bash
```diff
@@ -117,30 +117,189 @@ install_local_tool_versions() {
...
+    # if so also install the environment version
+    local env_version
+    env_version=$(get_version_from_env "$plugin_name")
+    if [ -n "$env_version" ]; then
+      display_debug "_install_directory_tools '$search_path': $plugin_name: using environment override $env_version"
+      plugin_version=$env_version
+
+      # install the version
+      display_none "$(install_tool_version "$plugin_name" "$plugin_version")"
+      tools_installed=$(printf "%s %s" "$tools_installed" "$plugin_name" | awk '{$1=$1};1')
```
Stratus3D, line 235: Quote here.

lib/functions/installs.bash
```diff
@@ -117,30 +117,189 @@ install_local_tool_versions() {
...
+  for plugin_name in $plugins_installed; do
+    # skip if plugin is installed already
+    if [[ -n $(stringlist_contains "$tools_installed" "$plugin_name") ]]; then
+      display_debug "_install_directory_tools_legacy '$search_path': legacy_install $plugin_name: skipping as tool was already installed"
+      continue
+    fi
+
+    # extract plugin legacy information
+    local plugin_path
+    plugin_path=$(get_plugin_path "$plugin_name")
```
Stratus3D, line 260: Quote here. Path may contain whitespace.

lib/functions/installs.bash
```diff
@@ -117,30 +117,189 @@ install_local_tool_versions() {
...
+
+    # skip if no legacy_list_filenames_script available
+    if ! [[ -f "$legacy_list_filenames_script" ]]; then
+      display_debug "_install_directory_tools_legacy '$search_path': legacy_install $plugin_name: skipping as legacy files are not supported"
+      continue
+    fi
+
+    # extract plugin legacy filenames
+    local legacy_filenames=""
+    legacy_filenames=$("$legacy_list_filenames_script")
```
Stratus3D, line 272: For good measure quote here. I can't imagine any legacy file names would contain whitespace since these are typically filenames chosen for software developers, but it won't hurt.

lib/utils.bash
```diff
@@ -156,6 +156,20 @@ display_error() {
...
+
+display_debug() {
+  if [[ $DEBUG = "true" ]]; then
+    printf "debug: %s\n" "$1" >&2
+  fi
+}
+
+display_none() {
+  printf ""
+}
```
Stratus3D, line 171: What is `display_none` for?

lib/utils.bash
```diff
@@ -156,6 +156,20 @@ display_error() {
   printf "%s\n" "$1" >&2
 }
 
+display_debug_hr() {
+  display_debug "--------------------------------------------------------------------------------------------------------------"
+}
```
Stratus3D, line 161: It looks like we only use this function once. Are we certain we need it?

Stratus3D:
@chrisjpalmer just scanned through the code and left some comments.

I see our build is broken here, but it isn't due to your changes. I'll see about fixing it on `master`.

lib/functions/installs.bash
slewsys: Not necessary. Word splitting is not done on values in assignments (see [Bash Shell Parameters](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameters)), i.e., quotes are not needed when assigning command substitutions (`$(...)`) or variables (`$var`).  E.g.:

```
foo='a space-filled value'
bar=$foo
echo $bar
```

> a space-filled value

```
foo=$(echo this is a command substitution with spaces)
echo $foo
```
> this is a command substitution with spaces

stoutput-icon:
Just checking in on this as it's been a while @Stratus3D. This would be an extremely useful feature to have implemented especially for python + poetry. Seems like the remaining asks are easy cleanups that could be knocked out in 2 min

n-rodriguez:
Hi there! Any news? I'm not sure if this PR is still relevant with `asdf 0.16.x` written in Go but I like the idea.

Thank you!

Stratus3D:
@chrisjpalmer as you know I recently re-rewrote asdf in Go, so this PR cannot be merged. I'm going to close it and write up what I think will be a more flexible solution and post it on https://github.com/asdf-vm/asdf/issues/929. Please follow that issue for future discussion.

PR #1724, opened 2024-02-24
fix: shim and preset plugin mismatch message

bootleq:
# Summary

Detect the case "missing preset plugin" is in fact installed, print different message instead of the misleading one.

Fixes: #928 

## Other Information

Example output:
sample-output

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -803,6 +803,7 @@ with_shim_executable() {
   (
     local preset_plugin_versions
     preset_plugin_versions=()
+    local preset_plugin_installed=
     local closest_tool_version
     closest_tool_version=$(find_tool_versions)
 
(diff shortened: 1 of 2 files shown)


Stratus3D:
Hi @bootleq , I merged #1805 this week which rewrites asdf in Golang. Can you test out the latest dev version? Download a dev binary from https://github.com/asdf-vm/asdf/releases/tag/54d15e3, or run ` go install github.com/asdf-vm/asdf@54d15e3`.

bootleq:
Yes, @Stratus3D, the issue still exists with latest dev.
You can install some new `node` version, set it as current node without installing `yarn`, to reproduce.

pr-1724-v0 15

Stratus3D:
Thanks @bootleq , this is an important fix. I'm going to translate this solution to the new Go. It may take me a couple weeks to get to it.

PR #1741, opened 2024-04-13
Update .gitignore

jeffersonsimaogoncalves:
Adding idea folders to gitignore.

/.fleet
/.idea
/.vscode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ repository
 .vagrant
 keyrings
 /tmp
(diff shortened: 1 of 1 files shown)


jthegedus:
Adding IDE specific ignore to projects is a contentious topic. Since we do not currently do this for other IDEs I will fall to the standard suggestion that you ignore these on your machine level, or add them but do not track the ignore file in Git.

PR #1742, opened 2024-04-14
fix(nushell): Use correct env var for shims dir

ludams:
# Summary

Changed the `asdf.nu` setup script to actually use the `ASDF_DATA_DIR` instead of the `ASDF_DIR` environment variable as base for the shims directory, which is added to the `PATH` env var.

Currently the script only checks, if the `ASDF_DATA_DIR` env var is set, but then uses `ASDF_DIR` env var instead to create the path to the shims. The scripts of the other shells do it correctly e.g. [asdf.fish](https://github.com/asdf-vm/asdf/blob/ccdd47df9b73d0a22235eb06ad4c48eb57360832/asdf.fish#L10).

## Other Information

[Relevant documentation part about ASDF_DATA_DIR](https://asdf-vm.com/manage/configuration.html#asdf-data-dir)


Stratus3D:
Thanks for the PR @ludams !

Stratus3D:
Thanks for the PR @deas ! Merging now as I've already got a fix for the failing build on #1806.

PR #1746, opened 2024-04-23
fix: when download path got removed, it should use -f to force delete…

kamontat:
… the download files

# Summary


Stratus3D (approved):

Stratus3D:
Thanks for the PR @kamontat !

PR #1747, opened 2024-04-24
docs: propose edits for tiny typographical/grammatical errors

zipperer:
### Summary

1. replace a use of "forcily" with a use of "forcibly". I believe "forcibly" is the word intended.
2. replace "`shell` set the version" with "`shell` sets the version". The second one matches the surrounding patterns of "`global` writes the version" and "`local` writes the version".


Stratus3D (approved):
Thanks for the PR @zipperer ! And sorry for the late review here. Changes look good.

PR #1755, opened 2024-05-11
fix: Avoid false warning on clean up.

slewsys:
# Summary

This is an alternative to #1721, and may be closer to the original intent: if `asdf` is configured to preserve downloads, but the download_path directory is missing, then give warning.

Fixes: Removes false warning on clean up.


BraDaga:
H

Stratus3D:
Hi @slewsys, sorry for the late reply here. Thanks for the PR! It looks like https://github.com/asdf-vm/asdf/pull/1756 was already merged, and solves the same issue so I'm going to close this PR. If that PR doesn't solve the full issue please comment here and we can re-evaluate this.

PR #1756, opened 2024-05-12
fix: Only display the "can't keep downloads" warning when asked to keep downloads

benblank:
# Summary

Currently, the warning which is intended to alert users that the download directory can't be kept only appears if they haven't asked for it to be. This fix just rearranges the logic a bit so that the warning instead only appears if the download directory *should* be kept, but *can't* be (because it doesn't exist).

Fixes: #1711

### DIFF
--- a/lib/functions/installs.bash
+++ b/lib/functions/installs.bash
@@ -234,15 +234,16 @@ install_tool_version() {
 
     local install_exit_code=$?
     if [ $install_exit_code -eq 0 ] && [ $download_exit_code -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


rubiin (approved):

leobastiani (approved):

CodeReaper (approved):

fharper:
Any ETA on merging this PR? Not critical, but it's there :)

Stratus3D (approved):
Sorry for the late review here. Looks good to me.

Issue #352, opened 2018-09-11
Use best-matching version


mwean:
### Steps to reproduce

```
$ asdf list ruby
> 2.3.1
  2.3.7
$ asdf local ruby 2.3
```

### Expected behavior

Use version 2.3.7, since it is the highest version that matches the given value.

### Actual behavior

`version 2.3 is not installed for ruby`

### Environment

**OS**: macOS 10.13.6

**asdf version**: 0.5.1

Similar to #226, I'd love to be able to be less specific about versions and get the highest matching installed version. For something like postgres, I might just say `9` or `10`, for ruby, I might use major and minor. I think this would make the tool a lot more useable.

If this sounds good, I could probably put together a PR, although I'm not sure how much overlap there will be in the logic for installing a highest-matching version.

vic:
Hm, currently asdf is very dumb about versions, they can actually be anything. I mean, not only semantic versions, but things like "system", "ref:master", even "foobar" is valid (I have things like "java intellij" or "android-tools brew" used with my asdf-link plugin)

My primary use for asdf is keeping a single `.tool-versions` in my projects and let any developer have the same *exact* versions for tools (compilers, etc). I'm using my `.tool-versions` file just like a Gem.lock or any-other-tool-locks-file to share the exact version to use on development, testing and production environments. 

So I'd say implementing this would cause more problems (not allowing you to not be *really* sure what was the exact version you are using just by looking at the source code of your project).

I'd consider this issue as something not desirable. But if you really like to have the behaviour you want, you could easily do: 

```shell
ln -s ~/.asdf/installs/ruby/2.3.7 ~/.asdf/installs/ruby/2.3
```

Yes, you'd have to update that symlink as you see fit. But again it's only on your development environment and not for all people in the world ^^.

mwean:
> My primary use for asdf is keeping a single .tool-versions in my projects and let any developer have the same exact versions for tools

That's definitely one of my use cases, but I also use it for side projects, scripts, etc. where I don't care about the exact version.

This change wouldn't impact any existing users. If you specify 2.3.7, it will use that exact version. If you just specify 2.3, it will find the best versions available. In my mind, it just adds more flexibility and doesn't take anything away.

vic:
Perhaps something better would be allowing some custom hooks for pre-post tool installation (so you can create the symlinks I mentioned, IIRC other people have asked for executing some commands like reshiming, or installing essential packages after tool installation. Can't remember right now if there's an issue for that). 

[...]

PR #1767, opened 2024-07-20
feat: allowing find the closest version if the exact version is not found

bsalesc:
# Summary
This is a way to allow you find the closest version in a non semantic way, if plugins are using semantic version is makes even easier.

I was going to open a feature request but there seem to be one opened here https://github.com/asdf-vm/asdf/issues/352.

This makes easier for users that have the below cases:

```
$ asdf list python
  2.7.18
  3.1.5
> 3.2.0
  3.2.1
```

In which I could `asdf global python 2` and as I've `2.7.18` installed and it's the only one that starts with 2, then that's the version that will be used, if I want 3, I need to be more specific, `asdf global python 3.1`,  `asdf global python 3.2.0` or `asdf global python 3.2.1`. Otherwise it'll fallback and complain saying `version {VERSION} is not installed for {PLUGIN}`.

Fixes: https://github.com/asdf-vm/asdf/issues/352

## Other Information
I've added tests for the "when is not found a single version" and "when it's found the closest version", let me know if there is need to have more tests.


test/version_commands.bats
```diff
@@ -204,6 +204,18 @@ teardown() {
   [ "$(cat "$HOME/.tool-versions")" = "dummy 1.1.0" ]
 }
 
+@test "global should create a global .tool-versions file when a close version is found" {
+  run asdf global "dummy" "1.1"
+  [ "$status" -eq 0 ]
+  [ "$(cat "$HOME/.tool-versions")" = "dummy 1.1.0" ]
+}
+
+@test "global should emit an error when a single close plugin version is found" {
```
bsalesc, line 213:
```suggestion
@test "global should emit an error when a single close plugin version is not found" {
```

RobLoach:
This is very handy. Thanks.

RobLoach:
@Stratus3D Thanks for the updates... Haven't been in the loop lately with the Go rewrite. There is some infrastructure that I have that won't be able to run Go executables. Is the intent to completely move to Go, and port everything away from bash/zsh/fsh?

I'm sure this question has been asked before. Where's the best place I can help out? Just the issue queue?

Stratus3D:
Hi @RobLoach , yes, this issue tracker is the best place to raise issues about both the current version (old Bash implementation) and the new Go version. We'll still have some completion code in bash/zsh/fish for good UX, but the core will be 100% Go.

RobLoach:
> What infra do you have that cannot run Go binaries?

Super old legacy hardware that will never receive any software updates. No problem though, I'll just keep that to use an older version of asdf.

Stratus3D:
> > What infra do you have that cannot run Go binaries?
> 
> Super old legacy hardware that will never receive any software updates. No problem though, I'll just keep that to use an older version of asdf.

I'm sorry to hear that we won't be able to support your older hardware. I was hoping Go would cover 99% of our users. You can certainly continue to use the current version.

Stratus3D:
Thanks for the PR @bsalesc. Unfortunately, asdf has been rewritten in Go (https://github.com/asdf-vm/asdf/releases/tag/v0.16.0) so this PR cannot be applied. I'm going to close it now.

Please continue discussion under #352.

PR #1768, opened 2024-07-23
Update getting-started.md - branch v0.14.0 does not exist

fcbry:
Update getting-started.md - branch v0.14.0 does not exist

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -36,7 +36,7 @@ asdf primarily requires `git` & `curl`. Here is a _non-exhaustive_ list of comma
 
 
 ```shell
-git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.14.0
+git clone https://github.com/asdf-vm/asdf.git ~/.asdf
 
 ```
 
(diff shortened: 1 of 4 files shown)


Stratus3D:
Hi @fcbry, the instructions are correct, at least for recent versions of Git.

> -b <name>
> --branch <name>
>
>    Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to <name> branch instead. In a non-bare repository, this is the branch that will be checked out. --branch can also take **tags and detaches the HEAD at that commit in the resulting repository.**

-- from https://git-scm.com/docs/git-clone

The Git flag is poorly named, but we instruct the user to pass in the tag via the `--branch` option, which is valid.

fcbry:
@Stratus3D - I mean the branch doesn't exist, and there is a different branch referenced in the japanese docs

Stratus3D:
@fcbry it's not a branch it's a tag.

fcbry:
My appologies, I skimmed your response and failed to pick up on the rather obvious and correct point:
>> The Git flag is poorly named, but we instruct the user to pass in the tag via the --branch option, which is valid.
😊

PR #1770, opened 2024-07-28
Fix Bash completions

benblank:
## Summary

This PR fixes two bugs I discovered in the Bash completions for version 0.14.0. I briefly tested the Fish completions as well, but did not see either issue there.

Fixes: (I did not find existing issues for these bugs.)

### Problem distinguishing between plugin names and commands

The first issue occurs when attempting to complete a command which is a substring of an installed plugin; for example, trying to complete `asdf shell ` while the `shellcheck` plugin is installed.

Currently, the Bash completions perform a simple substring match to determine whether the previous token is the name of a plugin: `[[ "$plugins" == *"$prev"* ]]`. The string `"shell"` is present in `"shellcheck"` and therefore the list of plugins, so it is interpreted as a plugin name and the completion tries to get the list of installed versions. The resulting list is empty, but because `" system"` is added, pressing <kbd>Tab</kbd> after `asdf shell ` completes with `system `.

The change I've made is to wrap both the list of plugins and the previous word in spaces (the delimiters used in the plugin list) — `[[ " $plugins " == *" $prev "* ]]`. This effectively forces a full-token match, as `" shell "` is *not* present in `" shellcheck "`.

### Asterisk present in installed-version completions

The other issue is that the version currently in use retains its asterisk in Bash completions of installed versions. This comes from the use of `asdf list $plugin` to get the list of versions without removing the asterisk which indicates the current version. The fix simply uses `colrm` to remove the first two columns of that command's output, which are always a space followed by either a space or an asterisk.

## Other Information

I didn't see any existing testing for command completions (I'm not sure how that would work), so no new tests cover these changes.

### DIFF
--- a/completions/asdf.bash
+++ b/completions/asdf.bash
@@ -39,7 +39,7 @@ _asdf() {
     COMPREPLY=($(compgen -W "$available_plugins" -- "$cur"))
     ;;
   install | list | list-all | help)
-    if [[ "$plugins" == *"$prev"* ]]; then
+    if [[ " $plugins " == *" $prev "* ]]; then
       local versions
       versions=$(asdf list-all "$prev" 2>/dev/null)
       # shellcheck disable=SC2207
@@ -54,7 +54,7 @@ _asdf() {
     COMPREPLY=($(compgen -W "--head" -- "$cur"))
     ;;
   uninstall | where | reshim)
-    if [[ "$plugins" == *"$prev"* ]]; then
+    if [[ " $plugins " == *" $prev "* ]]; then
       local versions
       versions=$(asdf list "$prev" 2>/dev/null)
       # shellcheck disable=SC2207
@@ -65,7 +65,7 @@ _asdf() {
     fi
     ;;
   local | global | shell)
-    if [[ "$plugins" == *"$prev"* ]]; then
+    if [[ " $plugins " == *" $prev "* ]]; then
       local versions
       versions=$(asdf list "$prev" 2>/dev/null)
       versions+=" system"
--- a/completions/asdf.bash
+++ b/completions/asdf.bash
@@ -56,7 +56,8 @@ _asdf() {
   uninstall | where | reshim)
     if [[ " $plugins " == *" $prev "* ]]; then
       local versions
-      versions=$(asdf list "$prev" 2>/dev/null)
+      # The first two columns are either blank or contain the "current" marker.
+      versions=$(asdf list "$prev" 2>/dev/null | colrm 1 2)
       # shellcheck disable=SC2207
       COMPREPLY=($(compgen -W "$versions" -- "$cur"))
     else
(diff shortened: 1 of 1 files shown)


completions/asdf.bash
```diff
@@ -54,9 +54,10 @@ _asdf() {
     COMPREPLY=($(compgen -W "--head" -- "$cur"))
     ;;
   uninstall | where | reshim)
-    if [[ "$plugins" == *"$prev"* ]]; then
+    if [[ " $plugins " == *" $prev "* ]]; then
       local versions
-      versions=$(asdf list "$prev" 2>/dev/null)
+      # The first two columns are either blank or contain the "current" marker.
+      versions=$(asdf list "$prev" 2>/dev/null | colrm 1 2)
```
Stratus3D, line 60: Is `colrm` available on most systems? Curious if it's available on all Linux distros and MacOS.

Stratus3D:
Thanks for the PR @benblank ! This is great!

I left one comment on the changes but otherwise I think this is good to merge.

> I didn't see any existing testing for command completions (I'm not sure how that would work), so no new tests cover these changes.

If you know anything about automated testing of shell completions let us know. This is a pain point for this project.

completions/asdf.bash
Stratus3D: It's available on my MacOS machine (Sonoma 14.5) and my Ubuntu madchine (Linux 5.4.0).

Stratus3D (approved):
Approved pending one comment.

completions/asdf.bash
benblank: Good catch, I hadn't even thought about that. 😅

Fortunately, it looks to be pretty universally available. It's part of util-linux, so should be part of every Linux distro since forever. Both the Linux and FreeBSD man pages identify `colrm` as having been part of ”3.0 BSD", which I think predates the Net/Free/Open split. That would put it in every version of OS X / macOS, as well. Oracle's man page even describes it as being part of Solaris (via util-linux again).

It's not clear to me whether the Linux and BSD variants are separate code bases (util-linux comes from <kernel.org> and the BSD man page indicates `colrm` is GNU software), but I'm not seeing anything about compatibility issues. `colrm` is a pretty ancient tool and very simple as well, after all.

benblank:
> If you know anything about automated testing of shell completions let us know. This is a pain point for this project.

I'm vaguely aware of automated shell testing via e.g. BATS, which sounds like it could be used in some manner, but don't know enough to do anything useful without starting from scratch. 🙃

completions/asdf.bash
Stratus3D: Thanks, I've not used `colrm` before but it seems like a good tool for situations like this. I'm going to merge this now, I'm sure if it breaks someone will complain :laughing: .

Stratus3D:
Yes we use BATS for general testing, but completions feel like a special case. I'll give it some thought and maybe add something later.

PR #1771, opened 2024-08-07
docs: Add a GitLab pipeline example

bheesham:
# Summary

Adds a GitLab pipeline example.

## Other Information

Tested here: https://gitlab.com/bheesham/asdf-exercism/-/blob/main/.gitlab-ci.yml?ref_type=heads

### DIFF
--- a/docs/plugins/create.md
+++ b/docs/plugins/create.md
@@ -874,6 +874,25 @@ os:
   - osx
 ```
 
+### GitLab Config
+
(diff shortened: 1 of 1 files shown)


bheesham:
Ah! I see there's actually a way better thing here: https://github.com/asdf-vm/asdf-plugin-template/blob/main/template/.gitlab-ci.yml

PR #1778, opened 2024-09-11
fix: make plugin-test work on alpine linux

vic1707:
# Summary

the alpine (busbox) version of `mktemp` expects templates to end with exactly 6 Xs whereas GNU want at least 3.

Bumping to 6 Xs allows alpine users to run the plugin-test command without installing the `coreutils` package first

## Other Information

On alpine anything with less than 6 Xs is an error
image

On ubuntu it doesn't matter if its 4 or 6
image

It will also fix running the `asdf-vm/actions/plugin-test` actions on an alpine container (for now an `apk add coreutils` is required)
image

### DIFF
--- a/lib/commands/command-plugin-test.bash
+++ b/lib/commands/command-plugin-test.bash
@@ -67,7 +67,7 @@ plugin_test_command() {
     set -- "${SHELL:-sh}" -c "$1"
   fi
 
-  TEST_DIR=$(mktemp -dt asdf.XXXX)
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR @vic1707 ! I think this is worth merging.

Stratus3D:
Merging now as I've already got a fix for the failing build on #1806.

Stratus3D (approved):

PR #1787, opened 2024-10-06
docs: Update core.md with clear explanation on reshim

HwangTaehyun:
# Summary

Hi! Recently, I was looking into asdf's reshim and thought supplementing the docs a bit more would be a good idea.

`npm` is usually installed by default when installing a node plugin, and according to the current example, when we do `npm install -g yarn`, npm itself is included in the asdf plugin lifecycle, so asdf will automatically create a shim for `yarn`.

For a clear explanation, I changed the docs for reshim to use yarn and pnpm instead of node's package manager npm.

Can you review it?

### DIFF
--- a/docs/manage/core.md
+++ b/docs/manage/core.md
@@ -38,7 +38,21 @@ A helper command to print the OS, Shell and `asdf` debug information. Share this
 asdf reshim <name> <version>
 ```
 
-This recreates the shims for the current version of a package. By default, shims are created by plugins during installation of a tool. Some tools like the [npm CLI](https://docs.npmjs.com/cli/) allow global installation of executables, for example, installing [Yarn](https://yarnpkg.com/) via `npm install -g yarn`. Since this executable was not installed via the plugin lifecycle, no shim exists for it yet. `asdf reshim nodejs <version>` will force recalculation of shims for any new executables, like `yarn`, for `<version>` of `nodejs` .
+This recreates the shims for the current version of a package. By default, shims are created by `asdf` plugins during the installation of a tool. However, when using package managers like `yarn` or `pnpm` to globally install executables (for example, installing [serve](https://github.com/vercel/serve) via `yarn add global serve`), these packages are not installed during the asdf plugin lifecycle. As a result, shims for these executables are not automatically created.
+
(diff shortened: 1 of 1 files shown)


hyperupcall:
Hey! This LGTM, but i'm not a maintainer and I'm no longer contributing to this project

HwangTaehyun:
@hyperupcall Oh, I see. Thank you!!

HwangTaehyun:
@Stratus3D Hi! Could you review this PR?

docs/manage/core.md
```diff
@@ -38,7 +38,21 @@ A helper command to print the OS, Shell and `asdf` debug information. Share this
 asdf reshim <name> <version>
 ```
 
-This recreates the shims for the current version of a package. By default, shims are created by plugins during installation of a tool. Some tools like the [npm CLI](https://docs.npmjs.com/cli/) allow global installation of executables, for example, installing [Yarn](https://yarnpkg.com/) via `npm install -g yarn`. Since this executable was not installed via the plugin lifecycle, no shim exists for it yet. `asdf reshim nodejs <version>` will force recalculation of shims for any new executables, like `yarn`, for `<version>` of `nodejs` .
+This recreates the shims for the current version of a package. By default, shims are created by `asdf` plugins during the installation of a tool. However, when using package managers like `yarn` or `pnpm` to globally install executables (for example, installing [serve](https://github.com/vercel/serve) via `yarn add global serve`), these packages are not installed during the asdf plugin lifecycle. As a result, shims for these executables are not automatically created.
```
Stratus3D, line 41: The problem with this (and the existing text) is that asdf is a generic version manager. Not everyone using asdf will know what `yarn` and `npm` are, as they don't use Node.JS. I understand the behavior may be confusing to newcomers, but I'm think the references to Node.JS package managers may make this more confusing for most new users.

HwangTaehyun: Thank you for the review! Yes, I agree! But as you know, the original example explanation for reshim was about node and npm, and it was not accurate, so I edited this part to make it clearer.

Also, in order to explain reshim, there should be a case where the asdf plugin lifecycle and other package managers exist, but I can't think of any. Do you have any other good examples?

Issue #1736, opened 2024-04-05
Support version requirements in tool-versions


KieranP:
### Is your feature request related to a problem? Please describe

Must currently specify exact versions, can't specify version in style of package.json (e.g. "^1.1.0")

### Describe the proposed solution

Support same version requirements system as nodejs, example:
* `>1.0.0`
* `>=1.1.0`
* `^1.1.0`
* etc

### Describe similar `asdf` features and why they are not sufficient

No similar feature currently, at least, not one that is documented

### Describe other workarounds you've considered

No known workarounds available

Stratus3D:
Hi @KieranP , this is a common request I've received over the years. I've documented this on the FAQ page in #2091

Issue #1342, opened 2022-10-06
Switch to the latest release of a version


turboninh:
### Is your feature request related to a problem? Please describe

Is there a way to automatically switch to another runtime version using only the major version. Using NodeJs plugin, for example, running `asdf local nodejs 16` doesn't work but `asdf local nodejs 16.17.0`.

### Describe the proposed solution

Implement `asdf local nodejs 16.x.x` or `asdf local nodejs 16.x` as a command, then allow `16.x` or `16.3.x` in **.tool-versions**

### Describe similar `asdf` features and why they are not sufficient

The command `asdf local nodejs 16.17.0` will create **.tool-versions** file in the project repo with `nodejs 16.17.0` in it, which does not take effect if the other team member clones the repo and have `16.17.1` installed locally.

### Describe other workarounds you've considered

Force other team member to install the same version

jthegedus:
You can get the very latest with `asdf local nodejs latest`.

Unfortunately, there seems to be a bug where the filter expression `asdf local nodejs latest 16` does not resolve as expected. `16` does not get passed to the `latest` function to filter properly.

jthegedus:
The feature request for:

>Implement asdf local nodejs 16.x.x or asdf local nodejs 16.x as a command, then allow 16.x or 16.3.x in .tool-versions

has already been discussed and declined numerous times.

If `asdf local nodejs latest` is not sufficient, then this issue can be closed as a duplicate.

turboninh:
The issue is more about better suppport for auto-switching between versions using **.tool-versions** file. With `nvm`, for example, I can create an `nvmrc` and put in `16` and it will automatically select 16.x on computers of my team members, `asdf` doesn't seem to support that.

jthegedus:
> The issue is more about better suppport for auto-switching between versions using **.tool-versions** file. With `nvm`, for example, I can create an `nvmrc` and put in `16` and it will automatically select 16.x on computers of my team members, `asdf` doesn't seem to support that.

Yes, that scenario is specifically not supported. `asdf` is designed to ensure everyone on a team is using the **exact** same version. If I am using `nodejs@16.0.0` and you are using `nodejs@16.19.0` those two versions are not anywhere close to being the same. They have different V8 engines, many different features and numerous bug fixes etc. How can we determine the root cause with this variability? Solution, remove the variability. Yes, this makes `asdf` slightly less convenient to use, but the purpose of locking a team to a version is to lock them to a version. Version ranges introduce variability.

[...]

Issue #1235, opened 2022-05-24
Interactive upgrades


timdp:
### Is your feature request related to a problem? Please describe

If you like to use latest-and-greatest, updating `.tool-versions` files with the right versions of individual tools can get pretty tedious.

### Describe the proposed solution

The Yarn package manager for JavaScript has an [`upgrade-interactive` command](https://yarnpkg.com/cli/upgrade-interactive) that provides a list of radio buttons for available upgrades. Here's an (outdated) example from a random blog post about it:

![](https://user-images.githubusercontent.com/201034/169978025-e2fc8a69-e2aa-43ee-8275-1bad5b1beb16.png)

I think this would be a good fit for asdf as well. Ultimately, I feel like it's a phase in the life of every package manager out there.

To get a feel for it, I wrote [`asdf-bump`](https://github.com/timdp/asdf-bump) in JS. It's just a proof of concept but it does work for me. (I initially wanted to create a PR but as comfortable as I am with Bash, it was a lot easier to do in JS.)

### Describe similar `asdf` features and why they are not sufficient

* `asdf update` updates asdf itself.
* `asdf list all <tool>` lists upgrades but doesn't apply them.
* `asdf latest <tool>` shows the latest version. You may not want the very latest one though. For example, with Node.js, you may want the latest version under the 16.x branch.

### Describe other workarounds you've considered

I think the only workaround is the manual approach, which is labor-intensive.

jthegedus:
As a note: `asdf latest <tool> <version>` exists to filter specific versions. Eg: `asdf latest nodejs 16` will give latest of the `16` releases.

As you state, this is easier to implement in tools that aren't Shell scripts.

I do not think interactive update is something we would consider adding to the core. This is something that would be ideal to implement as a user-land script. We are intending on adding a recipes/cookbook section to the documentation site. Feel free to share your Shell implementations here for sharing.

timdp:
> As a note: `asdf latest <tool> <version>` exists to filter specific versions. Eg: `asdf latest nodejs 16` will give latest of the `16` releases.

I know, but part of interactive upgrades is having the option to upgrade to a higher major version.

> As you state, this is easier to implement in tools that aren't Shell scripts.

I think that goes for most features of asdf, and most things in general, but I don't think that's a valid reason not to include it in asdf. 🙂 

[...]

Issue #1012, opened 2021-07-26
Support "latest" in .tool-versions


cj81499:
### Is your feature request related to a problem? Please describe

I would like the default version of various tools on my system to always be the latest available version.

### Describe the proposed solution

I propose adding support for the special value "latest" within `.tool-versions` files. It would evaluate to the latest available version for that particular tool.

One concern I see with this is the need to frequently check for updates to a given tool. To avoid this, available versions of a tool could be cached for some duration. Ideally, this duration would be configurable by the user.

### Describe similar `asdf` features and why they are not sufficient

Currently, I can run `asdf global my-tool latest`, and the version in my global settings (`~/.tool-versions`) will be set to the latest available version. However, when a new version of the tool is released, the default remains set to whatever value latest evaluated to when I last ran the command.

### Describe other workarounds you've considered

I've considered changing my `.zshrc` file to check for updates at shell launch, but this obviously has a significant negative impact on shell startup time.

Another possible workaround is a `cron` job that would run a script to check for updates. This would work, but I ultimately feel this functionality may be useful enough to justify being built-in.

jthegedus:
Updating your global deps is something you can script and run in user space, similar to most system package managers where installation of updates is opt-in on an update by update basis.

We do not want to support this type of feature in `asdf`. Here is an explanation from another discussion:

>Special versions like latest are resolved at the time the asdf command specifying them is run and are not permitted to in .tool-versions because of the nondeterminism they would introduce. We don't want to allow any variability in versions specified in any of the version files that asdf supports because we want asdf to be **deterministic**. We don't want a new version release to accidentally pull the rug out from under someone because they had specified a shorthand version or version range. **This constraint allows us to provide stronger guarantees about asdf behavior**.
> -@Stratus3D in https://github.com/asdf-vm/asdf-nodejs/issues/235

@Stratus3D I will PR an FAQ item for this issue so we can link to it in future. And and also propose an update to the legacy versions file support docs so we can link to it in discussions like https://github.com/asdf-vm/asdf-nodejs/issues/235.

Stratus3D:
> @Stratus3D I will PR an FAQ item for this issue so we can link to it in future. And and also propose an update to the legacy versions file support docs so we can link to it in discussions like asdf-vm/asdf-nodejs#235.

Thanks, sounds good!

rtyley:
> We don't want to allow any variability in versions specified in any of the version files that asdf supports because we want asdf to be **deterministic**.

[...]

PR #1793, opened 2024-10-17
feat: Enable use of `latest:` in `.tool-versions` files

rtyley:
# Summary

_Implements https://github.com/asdf-vm/asdf/issues/1012 , and addresses some of the underlying needs in https://github.com/asdf-vm/asdf/issues/1235, https://github.com/asdf-vm/asdf/issues/1342, https://github.com/asdf-vm/asdf/issues/1736, etc_

This change enables `asdf`'s existing latest-version-resolution functionality within the `.tool-versions` file itself. Rather than having to specify a full version number in the file:

```
java corretto-21.0.5.11.1
```

...you can now use the same `latest:` syntax that is already available in the `local` & `global` commands, ie:

```
java latest:corretto-21
```

### Use case

For many tool/runtime ecosystems (eg Java), if a program runs correctly under a specific version of that runtime, it can generally be relied on to run correctly under any _later_ version of that runtime with the same major version number (eg if a project runs under Corretto Java 21.0.5.11.1, it will run on any _later_ version of Corretto Java 21).

This means that for projects in those ecosystems, there is little incentive to pin to fully-specified versions like `21.0.5.11.1`, and in fact there are downsides - over time, developers will default to using older, unpatched versions of Java, unless they are assiduous in continually updating the contents of the `.tool-versions` file, or have tooling devoted to doing so.

At the Guardian we have [several hundred projects](https://github.com/orgs/guardian/repositories?q=visibility%3Apublic+language%3AScala+) that run on the Java platform, and due to our security obligations we generally want to be running under the _latest_ security-patched version of the Java runtime that matches our major-version requirement. We love `asdf` as a tool, and like that the `.tool-versions` file can become a source-of-truth documenting which version of Java a project uses, but we don't want to have to commit fully-specified version numbers like `21.0.5.11.1` to source control, or set up tooling to increment those version numbers across those hundreds of repositories.

Allowing the use of `latest:` in the `.tool-versions` file means that we don't need to continually update those `.tool-versions` files. It also partially addresses some of the needs raised by https://github.com/asdf-vm/asdf/issues/1736, though this solution uses the existing `asdf` version-resolution functionality, rather than adopting nodejs version syntax (`^1.1.0`, etc).

### Determinism & variability of version - could we let the user decide?

In the past, the maintainers of `asdf` have argued [against](https://github.com/asdf-vm/asdf-nodejs/issues/235#issuecomment-885809776) this kind of change:

> Special versions like `latest` are resolved at the time the asdf command specifying them is run and are not permitted to in `.tool-versions` because of the nondeterminism they would introduce. We don't want to allow any variability in versions specified in any of the version files that asdf supports because we want asdf to be **deterministic**. We don't want a new version release to accidentally pull the rug out from under someone because they had specified a shorthand version or version range.

...and [also](https://github.com/asdf-vm/asdf/issues/1342#issuecomment-1368572137):

> `asdf` is designed to ensure everyone on a team is using the **exact** same version. If I am using `nodejs@16.0.0` and you are using `nodejs@16.19.0` those two versions are not anywhere close to being the same. They have different V8 engines, many different features and numerous bug fixes etc. How can we determine the root cause with this variability? Solution, remove the variability. Yes, this makes `asdf` slightly less convenient to use, but the purpose of locking a team to a version is to lock them to a version. Version ranges introduce variability.

I think there are at least _some_ users who would accept some variability in exchange for staying up-to-date, and it's worth considering whether it's reasonable to allow them to select that. In terms of the existing `latest:` syntax already used by `asdf`, the constraint can be made quite tight (eg. `java latest:corretto-21.0.5`) or looser (eg. `java latest:corretto-21`) - so the choice of how much variation they want to accept can be expressed by the user.

In some ecosystems - like Java - if the version number is constrained to particular LTS major version, the only expected updates are security updates. At the Guardian, we _do_ want to be using those latest security updates, and so would use a constraint like `latest:corretto-21`.

### Implementation

A new `resolve_version_spec()` function has been extracted from the existing `version_command()` function. This takes a version-spec string, like `latest:corretto-11` or `corretto-21.0.5.11.1`, and resolves it to a precise version number.

To support `latest:` syntax in `.tool-versions`, the new `resolve_version_spec()` function is now called in `select_version()`, used by `with_shim_executable()`, meaning that any execution of the `asdf` shim (eg, executing `java`) will now resolve any version specifications found in the `.tool-versions` file - if `.tool-versions` contains `java latest:corretto-21`, this will be resolved and the latest version of Java 21 used.

## Testing

Tests have been added for these `asdf` commands to ensure they can handle version resolution in the `.tool-versions` file:

* `exec`
* `install`
* `current` 

## Other Information

Previous `asdf` PRs relating to `latest`:

* https://github.com/asdf-vm/asdf/pull/575 in November 2019: added the `latest` command, eg `asdf latest python 3.6` reports the latest version of Python 3.6.
* https://github.com/asdf-vm/asdf/pull/633 in July 2021: made it possible to specify `latest` when using the `local` & `global` commands, eg: `asdf local python latest:3.7` - this would save a precise version number to `.tools-versions`, which is undesired behaviour for us at the Guardian.

Previous `asdf` issues relating to keeping up with latest-and-greatest versions:

* https://github.com/asdf-vm/asdf-nodejs/issues/235
* https://github.com/asdf-vm/asdf/issues/1012 - _"I propose adding support for the special value "latest" within `.tool-versions` files"_
* https://github.com/asdf-vm/asdf/issues/1342 - _"Is there a way to automatically switch to another runtime version using only the major version"_
* https://github.com/asdf-vm/asdf/issues/1736 - _"specify version in style of package.json (e.g. "^1.1.0")"_
* https://github.com/asdf-vm/asdf/issues/1235 - _"to use latest-and-greatest, updating .tool-versions files with the right versions of individual tools can get pretty tedious"_

A couple of Guardian systems attempting to standardise on using `.tool-versions` as a source of truth:

* https://github.com/guardian/gha-scala-library-release-workflow/pull/36
* https://github.com/guardian/setup-scala

cc @adamnfish @akash1810 @tjsilver

### DIFF
--- a/lib/utils.bash
+++ b/lib/utils.bash
@@ -736,7 +736,8 @@ select_version() {
     version_and_path=$(find_versions "$plugin_name" "$search_path")
     IFS='|' read -r version_string _path <<<"$version_and_path"
     IFS=' ' read -r -a usable_plugin_versions <<<"$version_string"
-    for plugin_version in "${usable_plugin_versions[@]}"; do
+    for version_spec in "${usable_plugin_versions[@]}"; do
+      plugin_version="$(resolve_version_spec "$version_spec")"
       for plugin_and_version in "${shim_versions[@]}"; do
         local plugin_shim_name
         local plugin_shim_version
--- a/lib/commands/command-current.bash
+++ b/lib/commands/command-current.bash
@@ -1,6 +1,8 @@
 # -*- sh -*-
 # shellcheck source=lib/functions/plugins.bash
 . "$(dirname "$(dirname "$0")")/lib/functions/plugins.bash"
+# shellcheck source=lib/functions/versions.bash
+. "$(dirname "$(dirname "$0")")/lib/functions/versions.bash"
 
 # shellcheck disable=SC2059
 plugin_current_command() {
@@ -21,7 +23,8 @@ plugin_current_command() {
   local description=""
 
   IFS=' ' read -r -a versions <<<"$full_version"
-  for version in "${versions[@]}"; do
+  for version_spec in "${versions[@]}"; do
+    version="$(resolve_version_spec "$version_spec")"
     if ! (check_if_version_exists "$plugin_name" "$version"); then
       version_not_installed="$version"
     fi
--- a/lib/commands/command-env.bash
+++ b/lib/commands/command-env.bash
@@ -1,4 +1,6 @@
 # -*- sh -*-
+# shellcheck source=lib/functions/versions.bash
+. "$(dirname "$(dirname "$0")")/lib/functions/versions.bash"
 
 shim_env_command() {
   local shim_name="$1"
--- a/lib/commands/command-exec.bash
+++ b/lib/commands/command-exec.bash
@@ -1,4 +1,6 @@
 # -*- sh -*-
+# shellcheck source=lib/functions/versions.bash
+. "$(dirname "$(dirname "$0")")/lib/functions/versions.bash"
 
 shim_exec_command() {
   local shim_name
--- a/lib/commands/command-which.bash
+++ b/lib/commands/command-which.bash
@@ -1,4 +1,6 @@
 # -*- sh -*-
+# shellcheck source=lib/functions/versions.bash
+. "$(dirname "$(dirname "$0")")/lib/functions/versions.bash"
 
 which_command() {
   local shim_name
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -37,15 +37,7 @@ version_command() {
   declare -a resolved_versions
   local item
   for item in "${!versions[@]}"; do
-    IFS=':' read -r -a version_info <<<"${versions[$item]}"
-    if [ "${version_info[0]}" = "latest" ] && [ -n "${version_info[1]}" ]; then
-      version=$(latest_command "$plugin_name" "${version_info[1]}")
-    elif [ "${version_info[0]}" = "latest" ] && [ -z "${version_info[1]}" ]; then
-      version=$(latest_command "$plugin_name")
-    else
-      # if branch handles ref: || path: || normal versions
-      version="${versions[$item]}"
-    fi
+    version="$(resolve_version_spec "${versions[$item]}")"
 
     # check_if_version_exists should probably handle if either param is empty string
     if [ -z "$version" ]; then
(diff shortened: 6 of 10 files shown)


lib/utils.bash
```diff
@@ -736,7 +736,8 @@ select_version() {
     version_and_path=$(find_versions "$plugin_name" "$search_path")
     IFS='|' read -r version_string _path <<<"$version_and_path"
     IFS=' ' read -r -a usable_plugin_versions <<<"$version_string"
-    for plugin_version in "${usable_plugin_versions[@]}"; do
+    for version_spec in "${usable_plugin_versions[@]}"; do
+      plugin_version="$(resolve_version_spec "$version_spec")"
```
rtyley, line 740: This is the key change to enable the desired feature of this PR - now _resolving_ each version spec in the `.tool-versions` file to a precise version number, rather than it having to _already_ be a precise version number.

Stratus3D:
Hi @rtyley , sorry for the late response here. Unfortunately asdf has been rewritten in Go (https://github.com/asdf-vm/asdf/releases/tag/v0.16.0) so this PR cannot be applied. I'm going to close it now.

> Implements #1012 , and addresses some of the underlying needs in #1235, #1342, #1736, etc

Please continue discussion under these issues. My opposition to this feature is listed in #1012 - https://github.com/asdf-vm/asdf/issues/1012#issuecomment-886350845. I understand this isn't ideal for your workflow. But without exact versions asdf would no longer be deterministic.

claytonolley:
Not sure where best to put this so here it goes. 

We use asdf and Terraform Cloud. In Terraform Cloud, we can set the terraform workspace version to a floating patch version like `~>1.11.0` which will automatically use the latest patch version in `1.11`. We'd really like the ability to have asdf support this, although I do understand the reluctance. Ideally it would also install the latest `1.11.x` patch version as well but that I understand might be out of bounds! 😀

rtyley:
Hi @claytonolley - as confirmed here, `asdf` isn't going adopt any non-deterministic behaviour, and that's unfortunate for those of us who would like features that require it.

At my organisation, we're now adopting https://mise.jdx.dev/ as an alternative to `asdf` - `mise` reads the `asdf` `tool-versions`  file format, but also allows the floating/latest behaviour that we're looking for- it is very good, fast, and capable 👍

claytonolley:
Thanks @rtyley, I've come across that tool a couple times now, maybe it's finally time to take a closer look!

PR #1798, opened 2024-11-14
fix: Removed $HOME/ prefix from global filename path

alyvusal:
# Summary

Fixes:

1. Fixes issue for  ASDF_DEFAULT_TOOL_VERSIONS_FILENAME when location defined outside /home/user folder

### DIFF
--- a/lib/functions/versions.bash
+++ b/lib/functions/versions.bash
@@ -20,7 +20,7 @@ version_command() {
   file_name="$(asdf_tool_versions_filename)"
 
   if [ "$cmd" = "global" ]; then
-    file="$HOME/$file_name"
+    file="$file_name"
   elif [ "$cmd" = "local-tree" ]; then
     file=$(find_tool_versions)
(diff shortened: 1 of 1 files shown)


Stratus3D:
Hi @alyvusal thanks for the PR! However, I believe the existing code was correct. 


```
 file_name="$(asdf_tool_versions_filename)"
```

The function `asdf_tool_versions_filename` returns the `.tool-versions` **filename**. The filename is just the name of the file, not the full path to the file on the file system. A user might customize it to a different **filename**, but it must always be a **filename** because asdf climbs up directories on the file system and looks for files matching this **filename** in every directory on it's way to `/`. So when we need to come up with the "global" version file we do this:

```
file="$HOME/$file_name"
```

Which is correct if the user has set the default tool versions filename setting to a **filename**.

alyvusal:
My intention here was toremove harcoded location of `.tool-versions`, my change allows to use custom location.

PR #1801, opened 2024-11-25
fix: Support for Windows with Git-Bash or MSYS2 Bash

daramkun:
# Summary
Support Windows through Git-Bash or MSYS2 Bash (Not WSL2) for Use vscode-elixir-ls on Windows

## Other Information
* On *NIX
   * It works as it used to.
* On Windows
   * When executing the `asdf reshim` command, create a `*.bat` file, and when executing the file, execute the command through bash.
   * `asdf reshim` remove `.exe` extension. (ex: erl.exe -> erl)

### DIFF
--- a/lib/commands/reshim.bash
+++ b/lib/commands/reshim.bash
@@ -78,7 +78,7 @@ write_shim_script() {
   local version=$2
   local executable_path=$3
 
-  if ! is_executable "$executable_path"; then
+  if ! is_executable "$executable_path" && ! is_executable "$executable_path.exe"; then
     return 0
   fi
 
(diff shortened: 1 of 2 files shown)


Stratus3D:
Hi @daramkun I'm sorry for the late reply here. Thanks for taking the time to create this PR. However, I've been re-writing asdf in Golang (PR #1805) and recently merged the Go code. It no longer makes sense to merge these changes for the Bash code since the next release will only contain the Go code. If support for Windows with Bash is still lacking in the Go code another PR targeting the Go code is welcome (latest dev builds here - https://github.com/asdf-vm/asdf/releases/tag/54d15e3).

PR #1805, opened 2024-12-16
feat(golang-rewrite): Rewrite application in Go

Stratus3D:


DeedleFake:
> Figure out how to handle shell completions. Currently completions are separate files, but I think it may make sense to bake them into the Go binary for easier distribution. Thoughts?

I took a look at a couple of other projects that have completions and it seems that the common thing to do is to have a `completion` subcommand. Hugo, for example, lets you specify the name of the shell and then [that subcommand just outputs the necessary script to set it up](https://github.com/gohugoio/hugo/blob/55ecd3a90e37b8d14651789a582d3b60c7dc4bd1/docs/content/en/commands/hugo_completion_bash.md?plain=1#L19). That way you can just put `source <(hugo completion bash)` in your `.bashrc`, for example, and completion works. Considering the setup process for asdf, something similar seems like a good route. It also fits well with `go install`-based installations.

Stratus3D:
Thanks for investigating @DeedleFake. That does seem like a good solution. I'll look through the Hugo codebase tomorrow and see how they do that. I'm using urfave/cli v2 for the CLI interface and it looks like does completions a different way. I may see if I can adapt urfave/cli to provide a subcommand that outputs a self-contained completion script.

Stratus3D:
Changes have been rebased onto latest `master`. Diff is now correct.

Stratus3D:
@DeedleFake completions added in https://github.com/asdf-vm/asdf-core-go/pull/97

Stratus3D:
Waiting for @asdf-vm/core team to review. I intend to merge this in a couple days.

Issue #1810, opened 2024-12-17
bug: lib/commands/command-exec.bash improperly attempts to reference `shim_args[@]`


dossy:
### Describe the Bug

Fix unbound variable error when executing shims with no arguments. When executing certain shims (like `gke-gcloud-auth-plugin`) with no arguments, the following error occurs:

```bash
/usr/local/opt/asdf/libexec/lib/commands/command-exec.bash: line 23: shim_args[@]: unbound variable
```

The issue occurs because the script attempts to reference `shim_args[@]` without properly checking if the array has been initialized or is empty.

Previously:

* https://github.com/asdf-vm/asdf/issues/522
* https://github.com/jthegedus/asdf-gcloud/issues/68
* https://github.com/Jakski/asdf-janet/pull/3
* https://github.com/code-lever/asdf-rust/issues/17

### Steps to Reproduce

1. Install asdf v0.14.1
2. Install gcloud plugin: `asdf plugin add gcloud`
3. Install gcloud: `asdf install gcloud 503.0.0`
4. Install kubectl plugin: `asdf plugin add kubectl`
5. Install kubectl: `asdf install kubectl 1.32.0`
6. Set up a GKE cluster
7. Try to access cluster: `kubectl get pods`
8. Observe error messages about unbound variable

### Expected Behaviour

`kubectl` outputs information about the pods in the cluster

### Actual Behaviour

The following error is displayed:

```
/usr/local/opt/asdf/libexec/lib/commands/command-exec.bash: line 23: shim_args[@]: unbound variable
E1216 19:50:38.362751   24048 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://[redacted]/api?timeout=32s\": getting credentials: exec: executable gke-gcloud-auth-plugin failed with exit code 1"
/usr/local/opt/asdf/libexec/lib/commands/command-exec.bash: line 23: shim_args[@]: unbound variable
E1216 19:50:38.480657   24048 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://[redacted]/api?timeout=32s\": getting credentials: exec: executable gke-gcloud-auth-plugin failed with exit code 1"
/usr/local/opt/asdf/libexec/lib/commands/command-exec.bash: line 23: shim_args[@]: unbound variable
[...]

PR #1811, opened 2024-12-17
fix: handle case correctly when `shim_args` is empty

dossy:
# Summary

Ensures that `shim_args` is properly initialized and handles the case correctly where it is empty.

Fixes: #1810

### DIFF
--- a/lib/commands/command-exec.bash
+++ b/lib/commands/command-exec.bash
@@ -3,7 +3,11 @@
 shim_exec_command() {
   local shim_name
   shim_name=$(basename "$1")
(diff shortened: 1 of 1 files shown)


Stratus3D:
Closing as explained in https://github.com/asdf-vm/asdf/issues/1810#issuecomment-2555657520

fxalgrain:
Thank your @dossy  for the PR, I have the same issue, and I was able to fix it locally. 

@Stratus3D if you can release a minor version with this fix, it will be amazing

PR #1849, opened 2025-01-23
fix(golang-rewrite): allow missing post-plugin-update callback during plugin update

Stratus3D:
* Remove incorrect breaking change from docs
* Handle missing post-plugin-update callback


docs/guide/upgrading-from-v0-15-to-v0-16.md
```diff
@@ -8,11 +8,6 @@ rather than a set of scripts.
 
 ## Breaking Changes
 
-### `download` is now a required callback for plugins
```
Stratus3D, line 11: Turns out I'd implemented things such that `download` is an optional callback. So this wasn't actually a change at all.

PR #1858, opened 2025-01-30
feat: adds install script for asdf v0.16.0 and above

jpm-halfspace:
# Summary

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,79 @@
+#!/bin/sh
+
+# Note: we cannot assume we're running bash and use the set -euo pipefail approach.
+set -e
+
+VERSION="0.16.0"
+FILE="asdf-v$VERSION"
+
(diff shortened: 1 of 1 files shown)


Stratus3D:
Hi @jpm-halfspace , thanks for the PR! However, I'm not sure we want to add yet another method for installing asdf. We currently provide 4 methods: OS package manager, manual binary download, `go install`, and build from source. It's not clear that adding a variant of the binary download will improve the user experience. More options often means more confusion as to which method is best.

@asdf-vm/core got any thoughts on this?

jpm-halfspace:
Hi @Stratus3D,

That's totally fair. I made this to help with installing the binary and just wanted to share in case other people found it useful too. 

If it's not in scope, then I'll just close the PR no worries :)

PR #1872, opened 2025-02-02
docs: update path from `~/.bashprofile` -> `~/.zshrc` for zsh

nurulhudaapon:
# Summary

The ZSH Configuration section was saying to add path in bashprofile file instead of zshrc


Stratus3D (approved):
Thanks for the PR @nurulhudaapon ! It looks like line 188 is wrong as well. I will fix that.

Issue #1876, opened 2025-02-03
bug: 0.16 doesn't pass current shell environment variables to command


garry-harthill-cko:
### Describe the Bug

Shells environment variables aren't set in the context of the command that `asdf` execs

### Steps to Reproduce

Set an environment variable that the command requires

Run the command

Application displays an error about missing environment variable

### Expected Behaviour

I would expect the parent shell environment to be passed to the exec'd binary.

### Actual Behaviour

The command doesn't recognise environment variables.

However manually running `asdf env terraform echo $SOME_ENVAR` does actually show the contents of the variable.

### Environment

```shell
OS:
Darwin MAC-G6WN34MH7H 23.6.0 Darwin Kernel Version 23.6.0: Fri Nov 15 15:13:15 PST 2024; root:xnu-10063.141.1.702.7~1/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.9 (x86_64-apple-darwin23.0)

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
v0.16.0

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/user/.asdf
ASDF_CONFIG_FILE=/Users/user/.asdfrc

ASDF INSTALLED PLUGINS:
terraform https://github.com/asdf-community/asdf-hashicorp.git c44c314b220f36790bbee4f6869e16a95ac36bfc
```

garry-harthill-cko:
After some digging it appeared that is was just when the environment variable value has additional `=`. eg:
`MYVAR=some things = with = in it`

The culprit is likely in `execenv.SliceToMap`:
```
	for _, envVar := range env {
		varValue := strings.Split(envVar, "=")
		if len(varValue) == 2 { // <<<< silently ignoring it when there's more `=` than 1
			envMap[varValue[0]] = varValue[1]
		}
	}
```

Stratus3D:
Hi @garry-harthill-cko , thanks for the additional info! I was testing this myself and confirmed again that env vars in general do work:

```
$ export MYVAR="some string val"

$ asdf exec irb
irb(main):001> ENV['MYVAR']
=> "some string val"
irb(main):002>
```

But as you've identified the env var will not be set if the value of it contains `=`. 

```
$ export MYVAR="some string=val"

asdf exec irb
irb(main):001> ENV['MYVAR']
=> nil
[...]

PR #1879, opened 2025-02-04
fix: correct SliceToMap environment variable parsing function

Stratus3D:
The old code did not allow the equals sign in the environment variable value. But the equals sign is a valid character in value.

Fixes https://github.com/asdf-vm/asdf/issues/1876

### DIFF
--- a/internal/execenv/execenv.go
+++ b/internal/execenv/execenv.go
@@ -69,8 +69,8 @@ func SliceToMap(env []string) map[string]string {
 
 	for _, envVar := range env {
 		varValue := strings.Split(envVar, "=")
-		if len(varValue) == 2 {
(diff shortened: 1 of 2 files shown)


tomciopp:
This seems to still be broken in asdf I'm running version 0.16.3 and am running into this issue.

Stratus3D:
@tomciopp can you provide steps to reproduce?

tomciopp:
using asdf v0.16.3 with elixir 0.18.2 and erlang 27.2.3 on most recent Mac OS.

export TEST="foo=abc"

```bash
$: iex

1: System.get_env("TEST") => nil
```

Stratus3D:
@tomciopp I'm seeing it as well. Definitely an issue with 0.16.3.

PR #1885, opened 2025-02-04
docs: Remove old hyphenated command from help

jn-jairo:
Remove old hyphenated command from help

* shim-versions -> shimversions

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -56,7 +56,7 @@ asdf env <command> [util]               Runs util (default: `env`) inside the
 asdf info                               Print OS, Shell and ASDF debug information.
 asdf version                            Print the currently installed version of ASDF
 asdf reshim <name> <version>            Recreate shims for version of a package
-asdf shim-versions <command>            List the plugins and versions that
+asdf shimversions <command>             List the plugins and versions that
                                         provide a command
 asdf update                             Update asdf to the latest stable release
 asdf update --head                      Update asdf to the latest on the master branch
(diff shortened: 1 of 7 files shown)


Stratus3D (approved):

help.txt
```diff
@@ -56,7 +56,7 @@ asdf env <command> [util]               Runs util (default: `env`) inside the
 asdf info                               Print OS, Shell and ASDF debug information.
 asdf version                            Print the currently installed version of ASDF
 asdf reshim <name> <version>            Recreate shims for version of a package
-asdf shim-versions <command>            List the plugins and versions that
+asdf shimversions <command>             List the plugins and versions that
```
Stratus3D, line 59: This file isn't actually read by the Go code, it only exists because the Bash code which still exists in the repo reads it. Can you revert this change from just this file?

Stratus3D (approved):
One comment otherwise looks good.

Stratus3D:
Thanks for the PR @jn-jairo ! I made one change and then merged your branch. Not sure why GitHub isn't recognizing that I've already applied these changes.

help.txt
jn-jairo: It is used by the docs at https://github.com/asdf-vm/asdf/blob/9502b564b0a986a9edf9b9dd37565f2220c5d901/docs/manage/commands.md?plain=1#L5

Stratus3D: This is fixed in https://github.com/asdf-vm/asdf/pull/1890

Issue #1867, opened 2025-02-01
bug: [0.16.0] Completion bash using old hyphenated commands


jn-jairo:
### Describe the Bug

The completion bash is using the old hyphenated commands which produces wrong completions.

### Steps to Reproduce

I'm using this to set up shell completions.
```bash
. <(asdf completion bash)
```

The wrong commands in the completion script produce the wrong completions like in this line of the completion script:

```bash
plugins=$(asdf plugin-list 2>/dev/null | tr '\n' ' ')
```

The `asdf plugin-list` outputs the help instead of the list of plugins, so it lists the words from the help text instead of the plugins.

The other commands also need to be fixed.

### Expected Behaviour

Correct completions.

### Actual Behaviour

Wrong completions.

### Environment

```shell
OS:
Linux notebook 6.8.0-52-generic #53-Ubuntu SMP PREEMPT_DYNAMIC Sat Jan 11 00:06:25 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

BASH VERSION:
5.2.21(1)-release

ASDF VERSION:
v0.16.0
```

dchenbecker:
This also impacts zsh

dchenbecker:
Also, a related issue is #1748

ilyachch:
Also impacts Fish.
And, as I can see, if type `asdf list <tab>`, it will output part of help:
```
ilyachch@Thinkpad ~> asdf list all
                                        add a Git repo as a plugin by
                                        begins with the given string
                                        default branch
                                        default branch or a particular git-ref
                                        directory.
                                        environment used for command shim execution.
                                        git urls and git-ref
                                        in the .tool-versions file
                                        install the latest stable version that
```

![Image](https://github.com/user-attachments/assets/ac066c1d-e59b-4646-b989-aa5ae9c41e55)

Stratus3D:
I won't be able to get to this this week. Contributions are welcome here.

xeron:
[...]

PR #1886, opened 2025-02-04
fix: completion bash

jn-jairo:
# Summary

Rewrite of the completion bash removing the hyphenated commands and adapting the other changes in command line.

Fixes: #1867

### DIFF
--- a/internal/completions/asdf.bash
+++ b/internal/completions/asdf.bash
@@ -12,10 +12,12 @@ _asdf() {
   cur=${COMP_WORDS[COMP_CWORD]}
   local cmd
   cmd=${COMP_WORDS[1]}
+  cmd2=${COMP_WORDS[2]}
+  cmd3=${COMP_WORDS[3]}
   local prev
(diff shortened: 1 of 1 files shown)


jn-jairo:
Updated with the correct code style.

Stratus3D (approved):

Stratus3D:
Thanks @jn-jairo ! This fix should ship sometime this week.

Stratus3D:
Want to take a stab at doing this for zsh @jn-jairo ?

jn-jairo:
> Want to take a stab at doing this for zsh @jn-jairo ?

Sorry, I don't use zsh.

I looked at the code but it's not as simple as the bash one.

I'm not a specialist at completion script, I managed to edit the bash one because that one is pretty simple.

Stratus3D:
Thanks @jn-jairo , I'm no expert at completion scripts in any shell 😆 . Going to try to fix the zsh and fish ones tonight.

Issue #1896, opened 2025-02-06
bug: lib/utils.sh: line 27: HOME: unbound variable


n-rodriguez:
### Describe the Bug

Hi there! I got this error when running `asdf plugin update --all` or `asdf plugin update nodejs` :

```sh
/Users/nicolas/.asdf/plugins/nodejs/bin/../lib/utils.sh: line 27: HOME: unbound variable
/Users/nicolas/.asdf/plugins/nodejs/bin/../lib/utils.sh: line 27: HOME: unbound variable
```

From https://github.com/asdf-vm/asdf-nodejs/issues/411#issuecomment-2638326045 :

> It seems the `asdf_data_dir()` function in `.asdf/plugins/nodejs/lib/utils.sh` file isn't reading the environment variables correctly. (eg. `ASDF_DATA_DIR`, `ASDF_DIR`, and `HOME`)
>
> Maybe the latest version of asdf-vm is not passing the environment variables?

Thank you!

### Steps to Reproduce

1. install asdf
2. install asdf-nodejs plugin
3. run `asdf plugin update --all` or `asdf plugin update nodejs`

### Expected Behaviour

It should update plugins without errors

### Actual Behaviour

It shows an error :

```sh
/Users/nicolas/.asdf/plugins/nodejs/bin/../lib/utils.sh: line 27: HOME: unbound variable
```

### Environment

```shell
OS:
Darwin MacBook-Pro-de-Nicolas.local 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6020 arm64 arm Darwin

SHELL:
GNU bash, version 5.2.37(1)-release (aarch64-apple-darwin24.0.0)
Copyright (C) 2022 Free Software Foundation, Inc.
Licence GPLv3+ : GNU GPL version 3 ou ultérieure <http://gnu.org/licenses/gpl.html>

Ceci est un logiciel libre ; vous être libre de le modifier et de le redistribuer.
AUCUNE GARANTIE n'est fournie, dans les limites permises par la loi.

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.16.1

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/nicolas/.asdf
[...]

PR #1899, opened 2025-02-06
fix: properly pass missing env vars in Plugin.Update

n-rodriguez:
# Summary

Properly pass missing env vars in Plugin.Update

Fixes:

* https://github.com/asdf-vm/asdf/issues/1896

Thank you!


internal/plugins/plugins.go
```diff
@@ -255,6 +255,9 @@ func (p Plugin) Update(conf config.Config, ref string, out, errout io.Writer) (s
 	}
 
 	env := map[string]string{
+		"HOME":                 os.Getenv("HOME"),
+		"ASDF_DIR":             os.Getenv("ASDF_DIR"),
```
Stratus3D, line 259: `ASDF_DIR` isn't something that is used in asdf 0.16.0+. Are there still plugins relying on this?

n-rodriguez: > Are there still plugins relying on this?

Yes : https://github.com/asdf-vm/asdf-nodejs/blob/93bd217ba74f05080eebb7d8e1c146a385b01d7f/lib/utils.sh#L24

andrecloutier: Is there a reason why we shouldn't propagate all env variables of the current process by default here (similar to https://github.com/asdf-vm/asdf/pull/1958 )?

n-rodriguez: ping @Stratus3D thank you!

Stratus3D (commented):
I think if plugin callbacks need arbitrary environment variables I think we should pass-through any environment variable that is already set. Like we do for shims when executing the real executable.

Stratus3D:
Replaced by https://github.com/asdf-vm/asdf/pull/1982

PR #1901, opened 2025-02-06
fix: update make utils script to set correct version

mkungla:
# Summary

This update modifies `./scripts/asdf-version` to correctly determine the version for `Makefile` Go build `ldflags` when building from source.  It improves accuracy by aligning the versioning with Git tags and eliminates the need for manual version management via `version.txt`.

#### Changes  
- Replaces the usage of `version.txt`, which is now redundant.  
- Retrieves the latest semantic version tag (`vX.Y.Z`) using `git describe --tags`.  in asdf_version


Stratus3D:
Thanks for the PR @mkungla ! I actually made a few more changes so that the `asdf-version` script could be removed completely. It was invoked in the GitHub workflow but the output from it was never actually used. So the only place it was used was in the Makefile for dev builds. PR + additional changes has been merged.

mkungla:
@Stratus3D with you changes make created binary prints now 

```
$ asdf --version
asdf version "fdcd8af-dev"
```

I would expect also dev builds to produce valid asdf version, this is essential to ensure that tools which may work differently based on asdf version would work correctly when using dev build.

I suggest using in Makefile somethin like this

```makefile
# Set base version for dev build from previous tag
BASE_VERSION = $(shell git describe --tags --abbrev=0 2>/dev/null || echo "0.1.0")

# Get number of commits since the last tag so that it can be used to generate valid higher semver
COMMITS_SINCE_TAG = $(shell git rev-list $(BASE_VERSION).. --count 2>/dev/null || echo "0")

# Get the current commit hash (short); add hash as semver build label which is not used for version sorting precedence
GIT_HASH = $(shell git rev-parse --short HEAD)

# Determine final version:
ifeq ($(COMMITS_SINCE_TAG),0)
    # If exactly on a tag, use BASE_VERSION as-is
    FULL_VERSION = "$(BASE_VERSION)"
else
    # Extract major, minor, patch components needed to generate final dev build version
    MAJOR = $(shell echo $(BASE_VERSION) | awk -F. '{print $$1}')
    MINOR = $(shell echo $(BASE_VERSION) | awk -F. '{print $$2}')
    PATCH = $(shell echo $(BASE_VERSION) | awk -F. '{print $$3}')

    # Increment patch version for dev builds so that dev build version is always > than last tag version
    DEV_PATCH = $(shell echo $$(($(PATCH) + 1)))

    # Construct full version for dev build
    FULL_VERSION = "$(MAJOR).$(MINOR).$(DEV_PATCH)-dev.$(COMMITS_SINCE_TAG)+$(GIT_HASH)"
endif

# Linker flags
LINKER_FLAGS = '-s -X main.version=${FULL_VERSION}'

# Not sure what the default location should be for builds
build: # test lint
	go build -ldflags=${LINKER_FLAGS} -o=${TARGET_DIR}/${TARGET} ${MAIN_PACKAGE_PATH}
```

that would produce always valid [semver](https://semver.org/) for dev builds e.g. output would be

```
$ asdf --version
asdf version "v0.16.2-dev.9+856cc1d"
```

Issue #1908, opened 2025-02-06
bug: Completions in 0.16.1 still show removed hyphenated commands.


mike-mike-mike-mike-mike:
### Describe the Bug

[This section](https://asdf-vm.com/guide/upgrading-to-v0-16#hyphenated-commands-have-been-removed) of the migration documentation notes that hyphenated commands were replaced with non-hyphenated versions. Additionally, [this commit](https://github.com/asdf-vm/asdf/commit/1f0296a3d11a9df0d300ca61f59d097750ab6f1c) updates the help text accordingly. 

But the completions that are generated (at least in zsh) still show the old, hyphenated versions (see image). I've even uninstalled and reinstalled asdf completely.

![Image](https://github.com/user-attachments/assets/4f6dba45-dfbc-47f1-b4d9-81402b1fcf2f)

![Image](https://github.com/user-attachments/assets/6b71da9d-3044-4f7a-b665-6f3a605d09b7)

### Steps to Reproduce

1. Install or upgrade to asdf 0.16.1
2. type asdf into zsh and hit `TAB`

### Expected Behaviour

Completion options should match the actual commands

### Actual Behaviour

Completions show commands that no longer exist

### Environment

```shell
OS:
Darwin Mac 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:22 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6041 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
3.2.57(1)-release

ASDF VERSION:
0.16.1

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/michael.lambert/.asdf
ASDF_CONFIG_FILE=/Users/michael.lambert/.asdfrc

ASDF INSTALLED PLUGINS:
nodejs    https://github.com/asdf-vm/asdf-nodejs.git 93bd217ba74f05080eebb7d8e1c146a385b01d7f
ruby      https://github.com/asdf-vm/asdf-ruby.git   98e32b9c38e146b164d651c96ccb6d5fa4c3af90
yarn      https://github.com/twuni/asdf-yarn.git     376c5407ef22c82fa56e4bd9f6a9108354063acb
```

maple3142:
Same issue on zsh, and newer sub commands like `set` is not available too.

[...]

PR #1912, opened 2025-02-07
fix: completions for zsh and fish

manuraj17:
# Summary

Updates completions for `fish` and `shell` to use the new commands format without hyphens.

Fixes: #1908


Stratus3D (approved):
Thanks for the PR @manuraj17 ! I'm going to merge this as it is an improvement. But suggestions for things like `asdf plug*` don't suggest just `asdf plugin` but rather all plugin subcommands, which I think is not in line with what most people expect. I'm going to tweak it myself tomorrow and then try to publish a new patch version of asdf.

keldin-coding:
```
complete -f -c asdf -n __fish_asdf_needs_command -a 'plugin list all' -d "List all existing plugins"
```
I was able to "fix" this by wrapping the words passed to the `-a` argument here in quotes. However, the autocomplete still does not work as I would expect. For example `asdf plugin` completes, but none of the subcommands of that afterwards. 

I would imagine some working with [`__fish_seen_subcommand_from`](https://fishshell.com/docs/current/completions.html) will have to be used since all the commands have been converted to nested subcommands

manuraj17:
I was travelling when this got merged and I see the issue. It was not tested completely, I was still exploring fish completions and there are some more work. Apologies for the trouble caused.

manuraj17:
Folks, I have made some changes here https://github.com/asdf-vm/asdf/pull/1936. I tested it by copying the completions into my fish shell config directory and running it. Let me know if there are still any edge cases. @keldin-coding @kemko 

cc @Stratus3D

Stratus3D:
Thanks @manuraj17 ! Will test that PR more thoroughly and merge if it looks good 👍

PR #1916, opened 2025-02-07
feat: update makefile to generate valid semver for dev builds

mkungla:
# Summary

### **Ensure Dev Builds produce valid [SemVer](https://semver.org/) versions**  

This PR updates the `Makefile` to ensure that development builds generate **valid SemVer versions**, making it easier for tools to correctly identify the asdf version in different environments.  

Currently, dev builds produce versions like:

```sh
$ asdf --version
asdf version "fdcd8af-dev"
```

This format is **not** a valid SemVer version and can cause issues for tools that rely on version comparisons.

With this PR, dev builds will now follow **SemVer-compatible** formatting, ensuring they always sort **higher** than the latest tagged release. For example, after **9 commits** past v0.16.1, the output would be:

```sh
$ asdf --version
asdf version "v0.16.2-dev.9+856cc1d"
```

This ensures a consistent versioning scheme while keeping the commit hash for traceability.

## Other Information

This addresses the concern I raised in this comment: https://github.com/asdf-vm/asdf/pull/1901#issuecomment-2642643206

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,37 @@ MAIN_PACKAGE_PATH := ./cmd/asdf
 TARGET_DIR := .
 TARGET := asdf
 
-# Because this Makefile isn't used as part of the actual release binary build,
-# It sets FULL_VERSION to a dev version containing the SHA of the current
-# commit. If we ever use this Makefile to generate release binaries this code
-# will need to change.
-FULL_VERSION = "$(shell git rev-parse --short HEAD)-dev"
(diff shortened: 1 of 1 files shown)


Stratus3D:
Sorry for the late reply here. This PR does fix the issue for those running `make` but it does not fix it for those running `go install`. See https://github.com/asdf-vm/asdf/issues/2046

Stratus3D:
Closing in favor of https://github.com/asdf-vm/asdf/pull/2077

PR #1921, opened 2025-02-08
fix: correct help for asdf set command completion bash

jn-jairo:
# Summary

Update the completion bash with the fix from 554b4ea


Stratus3D (approved):
Thanks for catching this @jn-jairo !

PR #1927, opened 2025-02-09
Update version in version.txt

akkadaya:
# Summary

Update the version in version.txt to reflect on the https://asdf-vm.com website navbar

Fixes: Version on `asdf` website


Stratus3D:
Thanks for the PR @akkadaya ! Ended up re-structuring things a little so these changes weren't used. Version on the website should be correct again in just a minute!

Issue #1915, opened 2025-02-07
bug: asdf version displays the full help output - confusing scripts checking for what asdf is installed


jaclu:
### Describe the Bug

After the rewrite `asdf version` displays `asdf help` thus easily confusing scripts detecting what version of asdf is installed.

The version number is still listed on line 1 of the output, so in that sense it could be argued to still be valid output.

However this is not usable in reality, since virtually all version detections are done lazily, assuming the second word is the version, not filtering additional lines of output. 

End result: this causes essentially all scripts checking what asdf is installed to fail

### Steps to Reproduce

asdf version

### Expected Behaviour

version: v0.16.1

### Actual Behaviour

version: 0.16.1

MANAGE PLUGINS
asdf plugin add <name> [<git-url>]      Add a plugin from the plugin repo OR,
                                        add a Git repo as a plugin by
                                        specifying the name and repo url
asdf plugin list [--urls] [--refs]      List installed plugins. Optionally show
                                        git urls and git-ref
asdf plugin list all                    List plugins registered on asdf-plugins
                                        repository with URLs
asdf plugin remove <name>               Remove plugin and package versions
asdf plugin update <name> [<git-ref>]   Update a plugin to latest commit on
                                        default branch or a particular git-ref
asdf plugin update --all                Update all plugins to latest commit on
                                        default branch

MANAGE TOOLS
asdf current                            Display current version set or being
                                        used for all packages
asdf current <name>                     Display current version set or being
                                        used for package
asdf help <name> [<version>]            Output documentation for plugin and tool
asdf install                            Install all the package versions listed
                                        in the .tool-versions file
asdf install <name>                     Install one tool at the version
                                        specified in the .tool-versions file
asdf install <name> <version>           Install a specific version of a package
asdf install <name> latest[:<version>]  Install the latest stable version of a
                                        package, or with optional version,
                                        install the latest stable version that
                                        begins with the given string
asdf latest <name> [<version>]          Show latest stable version of a package
asdf latest --all                       Show latest stable version of all the
                                        packages and if they are installed
asdf list <name> [version]              List installed versions of a package and
                                        optionally filter the versions
asdf list all <name> [<version>]        List all versions of a package and
                                        optionally filter the returned versions
asdf set [-h] [-p] <name> <versions...> Set a tool version in a .tool-version in
                                        the current directory, or a parent
                                        directory.
asdf uninstall <name> <version>         Remove a specific version of a package
[...]

PR #1931, opened 2025-02-09
fix: add missing version command

jn-jairo:
# Summary

Adds the missing version command.

Fixes: #1915


Stratus3D (approved):
Thanks @jn-jairo !

Issue #1934, opened 2025-02-10
bug: asdf fish completion error


Syntist:
### Describe the Bug

getting these completion errors in fish terminal.

```╰─❯ asdf complete: too many arguments                                                                                                                                            ─╯

/usr/share/fish/vendor_completions.d/asdf.fish (line 58): 
complete -f -c asdf -n __fish_asdf_needs_command -a plugin add -d "Add git repo as plugin"
^
from sourcing file /usr/share/fish/vendor_completions.d/asdf.fish
        called on line 1 of file ~/.config/fish/functions/_autopair_tab.fish
in function '_autopair_tab'

(Type 'help complete' for related documentation)
complete: too many arguments

/usr/share/fish/vendor_completions.d/asdf.fish (line 64): 
complete -f -c asdf -n __fish_asdf_needs_command -a plugin list -d "List installed plugins"
^
from sourcing file /usr/share/fish/vendor_completions.d/asdf.fish
        called on line 1 of file ~/.config/fish/functions/_autopair_tab.fish
in function '_autopair_tab'

(Type 'help complete' for related documentation)
complete: too many arguments

/usr/share/fish/vendor_completions.d/asdf.fish (line 67): 
complete -f -c asdf -n __fish_asdf_needs_command -a plugin list all -d "List all existing plugins"
^
from sourcing file /usr/share/fish/vendor_completions.d/asdf.fish
        called on line 1 of file ~/.config/fish/functions/_autopair_tab.fish
in function '_autopair_tab'

(Type 'help complete' for related documentation)
complete: too many arguments

/usr/share/fish/vendor_completions.d/asdf.fish (line 70): 
complete -f -c asdf -n __fish_asdf_needs_command -a plugin remove -d "Remove plugin and package versions"
^
from sourcing file /usr/share/fish/vendor_completions.d/asdf.fish
        called on line 1 of file ~/.config/fish/functions/_autopair_tab.fish
in function '_autopair_tab'

(Type 'help complete' for related documentation)
complete: too many arguments

/usr/share/fish/vendor_completions.d/asdf.fish (line 74): 
complete -f -c asdf -n __fish_asdf_needs_command -a plugin update -d "Update plugin"
^
from sourcing file /usr/share/fish/vendor_completions.d/asdf.fish
        called on line 1 of file ~/.config/fish/functions/_autopair_tab.fish
in function '_autopair_tab'

(Type 'help complete' for related documentation)
complete: too many arguments

[...]

PR #1936, opened 2025-02-10
fix: repair invalid fish shell completion code

manuraj17:
Fixes related to #1934 & #1908

Earlier changes introduced in #1912 broke the fish shell completions. The changes were for both `zsh` and `fish`. Though, here we are making changes only for fixing `fish`.

There is room for improvement but the fix right now is focused to make the broken stuff work. 

Below commands are removed
- `global`
- `local`
- `update`

### DIFF
--- a/internal/completions/asdf.fish
+++ b/internal/completions/asdf.fish
@@ -50,30 +50,30 @@ function __fish_asdf_list_shims
     path basename $asdf_data_dir/shims/*
 end
 
-# update
-complete -f -c asdf -n __fish_asdf_needs_command -a update -d "Update asdf"
-complete -f -c asdf -n '__fish_asdf_using_command update; and __fish_asdf_arg_number 2' -l head -d "Updates to master HEAD"
-
 # plugin add completion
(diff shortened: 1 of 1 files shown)


Stratus3D:
I don't see a suggestion for `asdf plugin`:

```
~> asdf 
current  (Display version set or being used for package)  set                                       (Set version for a plugin)
info        (Print OS, Shell and ASDF debug information)  shimversions  (List the plugins and versions that provide a command)
install        (Install a specific version of a package)  uninstall                   (Remove a specific version of a package)
latest         (Show latest stable version of a package)  where                (Display install path for an installed version)
list              (List installed versions of a package)  which                        (Display executable path for a command)
reshim         (Recreate shims for version of a package) 
```

Plugin should be presented here as it is an option.

manuraj17:
@Stratus3D Added `plugin` option

I feel there could be more refinements for ex: currently `asdf plugin list <tab>` gives an all out list and then `all`. I can check on them, but since the current release has this broken maybe I can work on those incrementally since this works at the moment.

Stratus3D:
Thanks @manuraj17 !

Issue #523, opened 2019-05-17
Alias support


danhper:
> @bitwalker the goal of asdf is to make version management simple, that usually results in it being fairly easy to use too, but being easy to use is a side affect of being simple and well designed.

I'm definitely aware of the correspondence between simple/easy, my point was that as a user, those qualities mean different things than they do as a maintainer. Specifically, with my user hat on the "ease" in which I can avoid using multiple tools to do one thing (version management), and the "simplicity" of something that doesn't use any magic, just simple OS primtives, are the reasons why I chose it.

> It is fundamental.

> Part of being simple is only doing one thing, and for asdf that one thing is extendable version management. Adding support for an --alias flag would add a lot of complexity to the code for a feature isn't central to the purpose of asdf. 

I'm not convinced that it _is_ as complex as you expect. Supporting options with `getopt` would be a simple solution, and would vibe with using OS primitives as done so far, and certainly supports the basic functionality needed for the `--alias` flag.

> There are plenty of other tools for creating and managing aliases, these can be used in combination with asdf if users have the desire. asdf provides a good API, there is nothing preventing you from writing a shell script to make it do what you want.

I will of course have to investigate to see if there are limitations here, but if that's the case, then if I need to maintain my own shell script to add this capability, I'm not opposed to that idea, but it does seem rather fundamental to `asdf` itself. I'm a firm supporter of providing necessary primitives and letting the community go from there though, so as long as everything is present that would be required to support my proposal, then I'm good.

Do you have suggestions for tools for alias management? The shell is plainly inadequate for this out of the box, as creating arbitrarily-long lived aliases and destroying them on the fly is not something you get for free.

> Adding an --alias flag would fundamental change what a version in asdf actual is. Currently a version is exactly what it says it is. If you specify elixir 1.7.1, you can be sure you'll get the 1.7.1 release of Elixir, if you use elixir ref:abcdef1, you can be assured you will get that ref of Elixir. Using aliases would mean .tool-versions files would not contain versions, they'd contain identifiers which may be literal versions or may be aliases. 

[...]

PR #1937, opened 2025-02-10
fix: restore symlink support for tool versions

### LINKED ISSUE #1873: regresssion / feature: [0.16.0] asdf no longer recognizes symlinks as versions

(Note: this might actually be a regression for a unintended/unsupported feature, so I'm reluctant to classify this as a bug)

Before the change to use Go, it was possible to symlink inside the installs folder to create versions with a stable name. The main rationale was to use them for creating updateable patch versions.
These are useful e.g. for Java, where there is a plethora of patch versions that refer to essentially the same version (e.g. for java -> "temurin-21.0.0+35.0.LTS", "temurin-21.0.1+12.0.LTS" etc) - mostly including non-API-breaking security updates.

The issue is that having to reference the exact version in .tool-versions means that if you update, you need to go through the various files to update the number there. This either means you have to keep an insecure version or update the version files individually.

There are at some older reports related to this:
https://github.com/asdf-vm/asdf/issues/1622
https://github.com/asdf-vm/asdf/issues/523
https://github.com/andrewthauer/asdf-alias/issues/13

While I do not necessarily want to open a discussion about an alias core command specifically, I would like to have the symlink functionality restored so that you can write a plugin that does this or create those yourself.

You can obviously just copy the whole installed version after download - but this means that you need to implement a wrapper around the install/uninstall functionality.

The only workaround is to copy the installed version as far as I can see.

khmm12:
# Summary

Fixes a regression where symlink support for tool versions was inadvertently dropped. Versions prior to 0.16.0 supported symlinks for tool versions, which are used, for example, by [alias plugin](https://github.com/andrewthauer/asdf-alias).

- Fixes #1873

### DIFF
--- a/internal/installs/installs_test.go
+++ b/internal/installs/installs_test.go
@@ -48,34 +48,66 @@ func TestInstallPath(t *testing.T) {
 }
 
 func TestInstalled(t *testing.T) {
-	conf, plugin := generateConfig(t)
-
 	t.Run("returns empty slice for newly installed plugin", func(t *testing.T) {
+		conf, plugin := generateConfig(t)
 		installedVersions, err := Installed(conf, plugin)
 		assert.Nil(t, err)
 		assert.Empty(t, installedVersions)
 	})
 
 	t.Run("returns slice of all installed versions for a tool", func(t *testing.T) {
+		conf, plugin := generateConfig(t)
 		mockInstall(t, conf, plugin, "1.0.0")
 
 		installedVersions, err := Installed(conf, plugin)
 		assert.Nil(t, err)
 		assert.Equal(t, installedVersions, []string{"1.0.0"})
 	})
+
+	t.Run("returns installed versions including symlinks", func(t *testing.T) {
+		conf, plugin := generateConfig(t)
+		mockInstall(t, conf, plugin, "1.0.0")
+		symlinkVersion(t, conf, plugin, "1.0.0", "1")
+
+		installedVersions, err := Installed(conf, plugin)
+		assert.Nil(t, err)
+		assert.Equal(t, installedVersions, []string{"1", "1.0.0"})
+	})
 }
 
 func TestIsInstalled(t *testing.T) {
-	conf, plugin := generateConfig(t)
-	installVersion(t, conf, plugin, "1.0.0")
-
 	t.Run("returns false when not installed", func(t *testing.T) {
+		conf, plugin := generateConfig(t)
+		installVersion(t, conf, plugin, "1.0.0")
+
 		version := toolversions.Version{Type: "version", Value: "4.0.0"}
 		assert.False(t, IsInstalled(conf, plugin, version))
 	})
+
 	t.Run("returns true when installed", func(t *testing.T) {
+		conf, plugin := generateConfig(t)
+		installVersion(t, conf, plugin, "1.0.0")
+
 		version := toolversions.Version{Type: "version", Value: "1.0.0"}
 		assert.True(t, IsInstalled(conf, plugin, version))
+	})
+
+	t.Run("returns true when using symlinks", func(t *testing.T) {
+		conf, plugin := generateConfig(t)
+		installVersion(t, conf, plugin, "1.0.0")
+		symlinkVersion(t, conf, plugin, "1.0.0", "1")
+
+		version := toolversions.Version{Type: "version", Value: "1"}
+		assert.True(t, IsInstalled(conf, plugin, version))
+	})
+
+	t.Run("returns false when symlink broken", func(t *testing.T) {
+		conf, plugin := generateConfig(t)
+		installVersion(t, conf, plugin, "1.0.0")
+		symlinkVersion(t, conf, plugin, "2.0.0", "2")
+
+		version := toolversions.Version{Type: "version", Value: "2"}
+		assert.False(t, IsInstalled(conf, plugin, version))
 	})
 }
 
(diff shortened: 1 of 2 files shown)


internal/installs/installs_test.go
```diff
@@ -101,6 +133,14 @@ func mockInstall(t *testing.T, conf config.Config, plugin plugins.Plugin, versio
 	assert.Nil(t, err)
 }
 
+func aliasVersion(t *testing.T, conf config.Config, plugin plugins.Plugin, versionStr, aliasStr string) {
+	t.Helper()
+	originPath := InstallPath(conf, plugin, toolversions.Version{Type: "version", Value: versionStr})
+	aliasPath := InstallPath(conf, plugin, toolversions.Version{Type: "version", Value: aliasStr})
+	err := os.Symlink(originPath, aliasPath)
+	assert.Nil(t, err)
```
Stratus3D, line 141: These last two lines can be combined into one.

internal/installs/installs_test.go
```diff
@@ -48,35 +48,67 @@ func TestInstallPath(t *testing.T) {
...
 
 		installedVersions, err := Installed(conf, plugin)
 		assert.Nil(t, err)
 		assert.Equal(t, installedVersions, []string{"1.0.0"})
 	})
+
+	t.Run("returns installed versions including symlinks", func(t *testing.T) {
+		conf, plugin := generateConfig(t)
+		mockInstall(t, conf, plugin, "1.0.0")
+		aliasVersion(t, conf, plugin, "1.0.0", "latest")
```
Stratus3D, line 70: We do not support aliases in asdf core. Please rename this function to something that indicates it only generates a symlink, and please remove the `latest` version as `latest` is never permitted in `.tool-versions` files.

Stratus3D (commented):
Thanks for the PR @khmm12! I'm in favor of restoring symlink support for `.tool-versions` files if it is indeed missing, but it does not appear that these changes actually do that. These changes are affecting how asdf determines if a version is installed, which is completely unrelated to the process that locates and reads `.tool-versions` files. I think it'd be best to close this PR as I don't think any of these changes can be accepted.

khmm12:
Greetings @Stratus3D :wave:

> These changes are affecting how asdf determines if a version is installed, which is completely unrelated to the process that locates and reads .tool-versions files

Exactly. The previous Bash-based versions correctly detected installed tool versions that were symlinks. However, the current implementation no longer recognizes them, which is a clear regression.

I agree that aliasing should be handled by reworking how asdf selects tool versions from .tool-versions files. However, this PR specifically addresses the regression where symlinked versions, which were previously supported, are no longer detected. Recent issue reports show that many users rely on symlinks for managing tool versions, and restoring this functionality is important for maintaining backward compatibility.

Example use case of symlinks:

```
asdf/installs/nodejs/22.10.0
asdf/installs/nodejs/22.10      -> ./22.10.
asdf/installs/nodejs/22         -> ./22
```

.tool-versions
```
nodejs 22
```

This setup worked in previous versions but is no longer working. Since users depend on this behavior, it seems reasonable to restore it. 

Could you clarify do you see an alternative way to address the issue?

yacchi:
My plugin ([asdf-go-sdk](https://github.com/yacchi/asdf-go-sdk)) has also stopped working due to this issue. However, I have confirmed that this pull request resolves the problem.

From what I can see, symbolic link detection is handled carefully, and the errors module is used for error handling, making it a well-implemented fix that properly restores the previous behavior.

I hope this gets merged and released soon so that my plugin can function again.

Stratus3D:
Hi @khmm12. The asdf data directory (`~/.asdf/` by default) is where asdf is allowed to store plugins, versions, shims, and anything else it needs to store on disk on the computer. It's very important that these directories are not tampered with. The only directories in `~/.asdf` that plugins should ever write to are those given to the plugin's callbacks in the `*_PATH` environment variables - `ASDF_INSTALL_PATH`, `ASDF_DOWNLOAD_PATH`, and `ASDF_PLUGIN_PATH`. The only other place plugins should read/write files is in an OS-provided temp directory.

I'm sorry @yacchi you'll need to alter your plugin so that it doesn't need to write symlinks.

yacchi:
@Stratus3D 

Thank you for your response.

I understand the importance of asdf's data directory structure and the need to limit where plugins read and write files. However, the `asdf-go-sdk` plugin follows the behavior of the Go language itself by utilizing the default installation directory ($HOME/sdk).
This approach ensures compatibility with various tools and IDEs that expect the Go SDK to be in this location.

Changing this directory would break compatibility with these tools, making the development experience more cumbersome. Symbolic links were a solution to bridge this compatibility gap while integrating with asdf.

Given the current restrictions, do you have any suggestions or guidance on how to maintain compatibility with IDEs while adhering to asdf's data directory policy? We would greatly appreciate any ideas or best practices for addressing this challenge.

Thank you for your support and consideration.

ofalvai:
@Stratus3D I'd like to echo the concerns of @khmm12 and @yacchi about the current situation. There has been a lot of feedback over the years in #523 that alias versions as a concept is useful. I understand why you don't want it to become a core feature of asdf itself, but I'd like you to consider the practical reality of how asdf is used by many of its current users (plus the people who switched to Mise because of missing features like this one). The `asdf-alias` plugin was born to fill this gap, and while I agree that it's an imperfect solution, it is a practical fix for the current situation where the version alias concept is not a core feature of asdf itself.

Thank you for considering our use case, and I really hope we can find a way forward that fits everyone.

mrmeszaros:
Hi @Stratus3D, I am very saddened that this PR was closed.

I have a MAC, with `arm64` processor architecture, but can use Rosetta to run `x86_64` java versions which is necessary for some libraries that use JNI  with linked binaries only available for `x86` (like LMDB, RocksDB or LevelDB).

So since I cannot install `x86` java versions with `asdf`, I wanted to use an externally installed java version (downloaded with IntejjIJ IDEA), but I wasn't able to do it, because of the #1873 regression.

I am now very frustrated, and considering dropping the usage of this tool.
While I could revert to an older version, but if I need to change my setup, why not use something that does not pose such limits.

I hope this helps You understand why the support for symlinks is useful: it provides an extension point where the tool users can adapt `asdf` to their personal use-case. 
While a more native support for version aliases would be cool, the regression is not about that.
It is about an extension point that was previously supported, and was lost.
Letting symlinks be handled like other tools that work with file-systems do, could help us in the real world.

Yours truly,
a currently frustrated user of `asdf`

Issue #1893, opened 2025-02-05
bug: [0.16.0] asdf install when successful exits with 1


kapral18:
### Describe the Bug

When running asdf install against .tool-versions successfully, exits with 1

### Steps to Reproduce

1. Add .tool-versions
2. asdf install 
3. See status code 1

### Expected Behaviour

exit with status code 0 like before 0.16.0

### Actual Behaviour

<img width="531" alt="Image" src="https://github.com/user-attachments/assets/de7b5a4c-05ae-45dd-a470-4fb1107076f3" />

### Environment

```shell
🐟  ❯ asdf info
OS:
Darwin MacBook-Pro.local 24.2.0 Darwin Kernel Version 24.2.0: Fri Dec  6 19:01:59 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6000 arm64 arm Darwin

SHELL:
fish, version 3.7.1

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.16.1

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/kapral18/.asdf
ASDF_CONFIG_FILE=/Users/kapral18/.asdfrc

ASDF INSTALLED PLUGINS:
bun       https://github.com/cometkim/asdf-bun.git          019058c3a9f408a06bab63a4a3782c555a8dde3c
golang    https://github.com/asdf-community/asdf-golang.git e2527a31714da7bc671a684308579f4ef8863281
java      https://github.com/halcyon/asdf-java.git          a86e504d20eb3fa07dc6c8a9795247ad592056a3
lua       https://github.com/Stratus3D/asdf-lua.git         e0c15ea8bef0d4375353360e67c5c0e8403c7da2
nodejs    https://github.com/asdf-vm/asdf-nodejs.git        93bd217ba74f05080eebb7d8e1c146a385b01d7f
python    https://github.com/asdf-community/asdf-python.git a3a01856098d6d2b9642e382f5b38e70275726d1
ruby      https://github.com/asdf-vm/asdf-ruby.git          98e32b9c38e146b164d651c96ccb6d5fa4c3af90
rust      https://github.com/code-lever/asdf-rust.git       95acf4fe65df1de74fca502482b8f3ac5af73c05
[...]

PR #1951, opened 2025-02-13
Install should not return non-zero exit code on repeated invocations

### LINKED ISSUE #1929: Introduce flag for "asdf install" to ignore already installed version

I want to be able to write idempotent script that will install version if it's not installed, and skip it if it is already present, while also being able to detect actual installation errors.

Introduce new flag/option for `asdf install` to not fail in case the version is already installed, .e.g. `--ignore-installed`.

Currently none, but this was the default behaviour prior to 0.16.0 go rewrite

I can completely ignore exit code of "asdf install", but then I will also ignore actual installation errors.

reegnz:
# Summary

This is a bugfix that restores the idempotency of the install command.

In the bash version this invocation was idempotent and repeated calls
didn't return non-zero exit codes if first invocation succeeded to
install the tool.

This PR restores the idempotency contract that exists in the bash
version.

Fixes #1929


Stratus3D:
Thanks for the PR @reegnz ! I did not realize this was the same issue as was reported on #1893. #1952 has already been merged and applies changes similar to what you had here. Closing this PR.

PR #1954, opened 2025-02-14
docs: fix command examples in upgrading-to-v0-16 guide

deiga:
# Summary

Based on the description I believe the `asdf cmd foo bat man` has a typo in the name of the command thats bring run.
This PR changes the example


Stratus3D (approved):
Both the old and new behavior confuses me 😆 . Took me a minute but I believe you are correct.

PR #1956, opened 2025-02-14
fix: correct version resolution order to restore legacy file fallback behavior

nek023:
# Summary

Fixed the issue where the order of processing during version resolution was reversed between v0.15 and v0.16.

In v0.15, the lookup was performed in the order of `.tool-versions → legacy file`.

https://github.com/asdf-vm/asdf/blob/31e8c93004abd76253d186b8896785895069749b/lib/utils.bash#L166-L182

In v0.16, the lookup order was changed to `legacy file → .tool-versions`.

https://github.com/asdf-vm/asdf/blob/01461646483da1d35995bf990cca56d14490d5cc/internal/resolve/resolve.go#L59-L79

The documentation states that setting `legacy_version_file = yes` enables the behavior: "Use plugin fallback to legacy version files if available."
To ensure fallback works as expected, the lookup order needs to be reverted to the v0.15 behavior.

https://asdf-vm.com/manage/configuration.html#legacy-version-file


Stratus3D:
Thanks for the PR @nek023 ! I will test this today and hopefully merge tonight.

Issue #1924, opened 2025-02-08
bug: Missing environment variables in the exec-env script for plugins


darthShadow:
### Describe the Bug

Following from https://github.com/asdf-vm/asdf/issues/1876, would it be possible to also pass the environment variables to the exec-env script too?

I know it's not documented or supported, but there are quite a few plugins relying on the passthrough of environment variables.

Plugins relying on such behaviour:

https://github.com/search?q=org%3Aasdf-community+path%3Aexec-env++%22%24%7B%22+OR+%22-z%22+OR+%22unset%22&type=code

https://github.com/search?q=org%3Aasdf-vm+path%3Aexec-env+%22%24%7B%22+OR+%22-z%22+OR+%22unset%22&type=code

The follow is an example for the Golang plugin.

### Steps to Reproduce

1. Install the Golang Plugin
2. Set the GOPATH
3. Install the latest version of Golang
4. Run `go env`

### Expected Behaviour

GOPATH should be preserved.

### Actual Behaviour

GOPATH is overridden (because of https://github.com/asdf-community/asdf-golang/blob/e2527a31714da7bc671a684308579f4ef8863281/bin/exec-env#L8-L10)

### Environment

```shell
OS:
Darwin Darth-MBP.local 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000 arm64

SHELL:
fish, version 3.7.1

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.16.1

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/darthshadow/.asdf
ASDF_CONFIG_FILE=/Users/darthshadow/Personal/dotfiles/asdf/.asdfrc

ASDF INSTALLED PLUGINS:
golang    https://github.com/asdf-community/asdf-golang.git e2527a31714da7bc671a684308579f4ef8863281
python    https://github.com/asdf-community/asdf-python.git a3a01856098d6d2b9642e382f5b38e70275726d1
ruby      https://github.com/asdf-vm/asdf-ruby.git          98e32b9c38e146b164d651c96ccb6d5fa4c3af90
```

### asdf plugins affected (if relevant)

golang

ylansegal:
`asdf-postgres` plugin is also affected:

[...]

PR #1958, opened 2025-02-15
fix: pass environment to exec-env

ubuntuegor:
# Summary

Plugins want to work with the existing environment in exec-env, so we should pass it.

Fixes: #1924

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -521,6 +521,8 @@ func envCommand(logger *log.Logger, shimmedCommand string, args []string) error 
 		"PATH":                 setPath(execPaths),
 	}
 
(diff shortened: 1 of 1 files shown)


Stratus3D (approved):
Thanks for the PR @ubuntuegor ! Any suggestions on how we would better test this code? I'd like to get to a point where fixes aren't accepted without accompanying tests.

Stratus3D:
@ubuntuegor changes have been merged, along with an added BATS test.

ubuntuegor:
@Stratus3D Thanks! Good idea, adding tests for the fix didn't cross my mind, might be worth it to add a point about it to the PR template.

Stratus3D:
Yes that is a good idea! Will update the PR template!

PR #1964, opened 2025-02-16
fix: install should report missing plugins

andrecloutier:
# Summary

Indicates in the "install" output if a requested tool/version could not be installed due to the plugin not being installed. This will cause the application to exit with a non-zero exit code. Prior to this change, install would complete successfully without any indication that some required tool(s) were not successfully configured.

## Other Information

Sample output:
```
andre@Andres-MacBook-Pro asdf % ./asdf install
version {version 1.8.2} of bats is already installed
version {version 1.18.2-otp-27} of elixir is already installed
version {version 27.2.2} of erlang is already installed
version {version 1.23.4} of golang is already installed
version {version 8.12.1} of gradle is already installed
version {version zulu-21.40.17} of java is already installed
version {version 3.6.0} of shfmt is already installed
missing plugin for shellcheck
andre@Andres-MacBook-Pro asdf % echo $?
1
```

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


andrecloutier:
Closed this PR in favour of: https://github.com/asdf-vm/asdf/pull/1970

Issue #1923, opened 2025-02-08
Remove archived dependency: github.com/mitchellh/go-homedir


mihaitodor:
### Is your feature request related to a problem? Please describe

First of all, thank you for taking the time to rewrite this project in Go!

I noticed that [`github.com/mitchellh/go-homedir`](https://github.com/mitchellh/go-homedir) is used [here](https://github.com/asdf-vm/asdf/blob/63e7dcaeae46d6d43ae63db9bf635e227a6ba944/internal/config/config.go#L103) and [here](https://github.com/asdf-vm/asdf/blob/e7df5ff3253b6caba8ceef6b0754d17fcb765a9b/internal/config/config.go#L207-L215) to load the config and the env config, but this project has been archived by the maintainer and it might be worth migrating away from it.

### Describe the proposed solution

Deprecate and remove the usage of [`github.com/mitchellh/go-homedir`](https://github.com/mitchellh/go-homedir).

### Describe similar `asdf` features and why they are not sufficient

N/A

### Describe other workarounds you've considered

N/A

Stratus3D:
Contributions welcome here!

endorama:
What would be the accepted path forward here? I see some options: 
1. Copy the code in an internal package in this repo
2. Find an alternative third party library
3. Keep using it

The library itself is pretty simple and can be reviewed for an effective risk of compromise (dependencies and review for current code). As is archived as long as it compiles with the new Go versions there may not be a need to remove it.

Alternatively I think moving the code and test in an internal package would be a simple, effective and practical solution: that code is not going to require much maintenance given the reduced scope.

What do you think? Happy to provide a PR once we agree on a path forward ☺️

Stratus3D:
@endorama sorry for the late reply here. I think any of those options are acceptable, but I guess I'd prefer options 1 or 2.

PR #1970, opened 2025-02-18
Report missing plugins recursive

andrecloutier:
# Summary


internal/config/config.go
```diff
@@ -100,7 +101,7 @@ func LoadConfig() (Config, error) {
 		return config, err
 	}
 
-	homeDir, err := homedir.Dir()
+	homeDir, err := os.UserHomeDir()
```
manuraj17, line 104: Since you are replacing the library - it is being used here as well 
https://github.com/asdf-vm/asdf/pull/1970/files#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdL207

There is a related issue 
https://github.com/asdf-vm/asdf/issues/1923

andrecloutier:
@Stratus3D - Ready for some 👀 🙏

PR #1972, opened 2025-02-18
Fix: Add a newline delimiter when suggesting versions to install

andrecloutier:
# Summary

Adds a new line delimiter when suggesting versions for a user to set.

Sample output without this change:
```
andre@Andres-MBP ws % go version
No version is set for command go
Consider adding one of the following versions in your config file at /Users/andre/ws/.tool-versions
golang 1.24.0golang 1.23.4%
```


Stratus3D (approved):
Thanks for the fix @andrecloutier !

PR #1975, opened 2025-02-19
fix: envMap to work with env var values containing an equal sign.

jechol:
# Summary

Fixes: envMap to work with env var values containing an equal sign.

## Other Information

SliceToMap has been modified to support = in values, but envMap was not updated accordingly.


jechol:
Anyway, I'm glad it's been fixed. In our internal project, some developers are switching from asdf to vfox due to this issue. When is the release schedule for version 0.16.4?

Issue #1986, opened 2025-02-25
bug: Multiline string ENV var handling issue in shim logic (asdf v0.16.4)


eizengan:
### Describe the Bug

Related to https://github.com/asdf-vm/asdf/issues/1979

https://github.com/asdf-vm/asdf/pull/1982 fixes the problem which truncated most multiline ENV, however there are still issues handling values containing particular content. For example, we currently drop all lines beginning-with-and-including lines ending in =, which breaks e.g. x509 certificates. Other characters may also trigger this behavior; I am not sure.

### Steps to Reproduce

1. create a multiline env var like the following:
  ```
  export CERT="-----BEGIN CERTIFICATE-----
  MANY\LINES\THE
  LAST\ONE\ENDS\IN
  AN=
  -----END CERTIFICATE-----"
  ```
2. `ruby -e "puts ENV['CERT']"`

### Expected Behaviour

Prints
> -----BEGIN CERTIFICATE-----
  MANY\LINES\THE
  LAST\ONE\ENDS\IN
  AN=
  -----END CERTIFICATE-----

### Actual Behaviour

Prints
> -----BEGIN CERTIFICATE-----
  MANY\LINES\THE
  LAST\ONE\ENDS\IN

### Environment

```shell
OS:
Darwin C02FC0QGMD6V 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:22:00 PST 2025; root:xnu-11215.81.4~3/RELEASE_X86_64 x86_64

SHELL:
zsh 5.9 (x86_64-apple-darwin24.0)

BASH VERSION:
3.2.57(1)-release

ASDF VERSION:
0.16.4

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/nate.eizenga/.asdf
ASDF_CONFIG_FILE=/Users/nate.eizenga/.asdfrc

ASDF INSTALLED PLUGINS:
python    https://github.com/danhper/asdf-python.git           a3a01856098d6d2b9642e382f5b38e70275726d1
ruby      https://github.com/asdf-vm/asdf-ruby.git             2885ab7d311715ec4f6dfe2fc5dfed329fa1f263
rust      https://github.com/code-lever/asdf-rust.git          95acf4fe65df1de74fca502482b8f3ac5af73c05
[...]

PR #1988, opened 2025-02-25
fix: simplify env vars parsing

andrecloutier:
# Summary

Simplifies the parsing of env vars reported by bash by parsing NULL delimited values rather than new lines.

Fixes: https://github.com/asdf-vm/asdf/issues/1986

### DIFF
--- a/internal/execenv/execenv_test.go
+++ b/internal/execenv/execenv_test.go
@@ -74,9 +74,8 @@ func TestGenerate(t *testing.T) {
 		assert.Nil(t, err)
 		plugin := plugins.New(conf, testPluginName)
 		assert.Nil(t, repotest.WritePluginCallback(plugin.Dir, "exec-env", "#!/usr/bin/env bash\nexport BAZ=\""+value+"\""))
-		env, err := Generate(plugin, map[string]string{"EQUALSTEST": "abc\n123"})
+		env, err := Generate(plugin, map[string]string{})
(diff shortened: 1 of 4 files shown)


internal/execenv/execenv_test.go
```diff
@@ -65,4 +65,18 @@ func TestGenerate(t *testing.T) {
...
+		testDataDir := t.TempDir()
+		conf := config.Config{DataDir: testDataDir}
+		_, err := repotest.InstallPlugin("dummy_plugin", testDataDir, testPluginName)
+		assert.Nil(t, err)
+		plugin := plugins.New(conf, testPluginName)
+		assert.Nil(t, repotest.WritePluginCallback(plugin.Dir, "exec-env", "#!/usr/bin/env bash\nexport BAZ=\""+value+"\""))
+		env, err := Generate(plugin, map[string]string{"EQUALSTEST": "abc\n123"})
+		assert.Nil(t, err)
+		assert.Equal(t, value, env["BAZ"])
+		assert.Equal(t, "abc\n123", env["EQUALSTEST"])
```
Stratus3D, line 80: I think the `EQUALSTEST` code could be removed from this test as it's not really relevant and already converted by another `t.Run`.

Stratus3D (approved):
Thanks for the PR @andrecloutier ! I left one comment, otherwise looks good to me.

PR #1989, opened 2025-02-26
fix: set correct env vars on recursive calls

andrecloutier:
# Summary

Addresses the feedback mentioned here: https://github.com/asdf-vm/asdf/pull/1982#issuecomment-2682248660

The solution outlined here strips values which should not be passed through asdf on recursive calls. It's done before any other logic is invoked in order to minimize the number of places that need to consider this case. Open to feedback on this approach.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,9 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[*.go]
+indent_style = tab
+
 [*.py]
 indent_size = 4
 
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -340,7 +340,12 @@ func Execute(version string) {
 		},
 	}
 
-	err := app.Run(os.Args)
+	err := unsetAsdfReservedEnvVars()
+	if err != nil {
+		os.Exit(1)
+	}
+
+	err = app.Run(os.Args)
 	if err != nil {
 		os.Exit(1)
 	}
@@ -1565,4 +1570,17 @@ func installedStatus(installed bool) string {
 		return "installed"
 	}
 	return "missing"
+}
+
+func unsetAsdfReservedEnvVars() error {
+	// These are environment variables which are passed via env or exec.
+	// We strip these out to avoid any potential issues with recursive calls to asdf.
+	asdfManagedVars := []string{"ASDF_INSTALL_TYPE", "ASDF_INSTALL_VERSION", "ASDF_INSTALL_PATH"}
+	for _, v := range asdfManagedVars {
+		err := os.Unsetenv(v)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/internal/help/help.go
+++ b/internal/help/help.go
@@ -93,9 +93,7 @@ func WriteToolVersionHelp(conf config.Config, toolName, toolVersion string, writ
 func writePluginHelp(conf config.Config, toolName, toolVersion string, writer io.Writer, errWriter io.Writer) error {
 	plugin := plugins.New(conf, toolName)
 	env := map[string]string{
-		"ASDF_INSTALL_PATH":    plugin.Dir,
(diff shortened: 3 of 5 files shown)


Stratus3D:
Thanks for the PR @andrecloutier ! Testing this now.

internal/help/help.go
```diff
@@ -93,7 +93,9 @@ func WriteToolVersionHelp(conf config.Config, toolName, toolVersion string, writ
 func writePluginHelp(conf config.Config, toolName, toolVersion string, writer io.Writer, errWriter io.Writer) error {
 	plugin := plugins.New(conf, toolName)
 	env := map[string]string{
-		"ASDF_INSTALL_PATH": plugin.Dir,
+		"ASDF_INSTALL_PATH":    plugin.Dir,
+		"ASDF_INSTALL_VERSION": "",
+		"ASDF_INSTALL_TYPE":    "",
```
Stratus3D, line 98: Why do the changes above not unset these two variables?

internal/cli/cli.go
```diff
@@ -1566,3 +1571,16 @@ func installedStatus(installed bool) string {
...
+	// We strip these out to avoid any potential issues with recursive calls to asdf.
+	asdfManagedVars := []string{"ASDF_INSTALL_TYPE", "ASDF_INSTALL_VERSION", "ASDF_INSTALL_PATH"}
+	for _, v := range asdfManagedVars {
+		err := os.Unsetenv(v)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
Stratus3D, line 1586: It might be good to write tests that validate this is doing what we expect it to. The only way I can think to test this would be by testing the `asdf env` command, which normally passes the env through unchanged.

Stratus3D:
Here is a test I wrote that I thought would exercise this fix. It fails on latest master but also on this branch.

```diff
diff --git a/test/shim_env_command.bats b/test/shim_env_command.bats
index 2926f46..261df89 100644
--- a/test/shim_env_command.bats
+++ b/test/shim_env_command.bats
@@ -121,3 +121,23 @@ teardown() {
   run grep -q '::' <<<"$path_line"
   [ "$status" -ne 0 ]
 }
+
+@test "asdf env unsets ASDF_INSTALL_TYPE ASDF_INSTALL_VERSION and ASDF_INSTALL_PATH env variables" {
+  export ASDF_INSTALL_VERSION=1.2.3
+  export ASDF_INSTALL_TYPE=version
+  export ASDF_INSTALL_PATH=/somewhere
+
+  echo "dummy 1.0" >"$PROJECT_DIR/.tool-versions"
+  run asdf install
+
+  run asdf env dummy
+  [ "$status" -eq 0 ]
+  envs="$output"
+
+  echo "$envs" | grep -q "ASDF_INSTALL_TYPE" 2>/dev/null
+  [ "$status" -ne 0 ]
+  echo "$envs" | grep -q "ASDF_INSTALL_VERSION" 2>/dev/null
+  [ "$status" -ne 0 ]
+  echo "$envs" | grep -q "ASDF_INSTALL_PATH" 2>/dev/null
+  [ "$status" -ne 0 ]
+}
```

internal/help/help.go
andrecloutier: D'oh. I thought I added these to a test file. Let me see if we can do without.

andrecloutier: Could you re-test with your setup to confirm whether any of the help tests fail?

.editorconfig
```diff
@@ -8,6 +8,9 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[*.go]
+indent_style = tab
```
andrecloutier, line 12: Change here is to align with what `go fmt` emits.

internal/cli/cli.go
andrecloutier: Added such a test. Confirmed it fails without the change. Thanks for the sample to work off of!

PR #1991, opened 2025-02-27
fix: correct exit status when sub-command does not exist

### LINKED ISSUE #1928: bug: running asdf with incorrect command returns 0 exit code
### Describe the Bug

If asdf is called with incorrect (non-existent) command, I expect exit code to be non-zero.

### Steps to Reproduce

```bash
asdf foo-bar
echo $?
```

### Expected Behaviour

Exit code is non-zero

### Actual Behaviour

Exit code is 0

### Environment

```shell
➜ asdf info    
OS:
Darwin DAR-CF933GCFKQ 24.2.0 Darwin Kernel Version 24.2.0: Fri Dec  6 19:01:59 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.16.2
```

_No response_

AlfredoRamos:
Change exit status code when command does not exists.

Fixes #1928

### DIFF
--- a/test/non_existent_command.bats
+++ b/test/non_existent_command.bats
@@ -1,0 +1,32 @@
+#!/usr/bin/env bats
+
+load test_helpers
+
+setup() {
+  setup_asdf_dir
+  install_dummy_plugin
+  install_dummy_legacy_plugin
+  run asdf install dummy 1.0
+  run asdf install dummy 1.1
+
+  PROJECT_DIR="$HOME/project"
+  mkdir -p "$PROJECT_DIR"
+}
+
+teardown() {
+  clean_asdf_dir
+}
+
+@test "should show help when no valid command is provided" {
+  cd "$PROJECT_DIR"
+
+  run asdf non-existent-command
+
+  [ "$status" -eq 1 ]
+  [[ $output == 'invalid command provided:'* ]]
+  [[ $output == *$'version: v'* ]]
+  [[ $output == *$'MANAGE PLUGINS\n'* ]]
+  [[ $output == *$'MANAGE TOOLS\n'* ]]
+  [[ $output == *$'UTILS\n'* ]]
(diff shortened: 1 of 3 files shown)


test/non_existent_command.bats
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bats
+
+load test_helpers
+
+setup() {
+  setup_asdf_dir
+  install_dummy_plugin
+  install_dummy_legacy_plugin
+  run asdf install dummy 1.0
+  run asdf install dummy 1.1
```
Stratus3D, line 10: I think these 4 lines can be removed. They should not be needed for the code your test exercises.

test/non_existent_command.bats
```diff
@@ -0,0 +1,32 @@
...
+
+  PROJECT_DIR="$HOME/project"
+  mkdir -p "$PROJECT_DIR"
+}
+
+teardown() {
+  clean_asdf_dir
```
Stratus3D, line 17: Is this needed? You don't actually install anything so I'm not sure this clean function needs to run.

Stratus3D (approved):
Thanks for the PR @AlfredoRamos ! I left a couple comments but overall I think this PR is great.

test/non_existent_command.bats
AlfredoRamos: Yep, local tests pass without those lines.

test/non_existent_command.bats
AlfredoRamos: Not sure.

I see it removes some environment variables.

I removed the `teardown()` function.

Stratus3D (approved):

Stratus3D:
Thanks for the contribution @AlfredoRamos !

Issue #1980, opened 2025-02-20
bug: Updating the python plugin breaks it


kief:
### Describe the Bug

After running `asdf plugin update python`, the python plugin is no longer able to list avaiable versions.

### Steps to Reproduce

```
asdf plugin remove python
asdf add python
asdf latest python
asdf plugin update python
adsf latest python
```

### Expected Behaviour

```
$ asdf latest python
stackless-3.7.5
$ asdf latest python 3.10.
3.10.16
$ asdf list all python
(LONG LIST)
```

### Actual Behaviour

```
$ asdf latest python
No compatible versions available (python )
$ asdf latest python 3.10.
No compatible versions available (python 3.10.)
$ asdf list all python

$
```

### Environment

```shell
$ asdf info
OS:
Darwin Kiefs-Macbook-2024.local 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:22 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6041 arm64

SHELL:
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin24)
Copyright (C) 2007 Free Software Foundation, Inc.

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.16.3

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/kief/.asdf
ASDF_CONFIG_FILE=/Users/kief/.asdfrc

ASDF INSTALLED PLUGINS:
java      https://github.com/halcyon/asdf-java.git   97f962b1fc9a359d8c5c2c6abf10ebad51666c1c
python    https://github.com/danhper/asdf-python.git a3a01856098d6d2b9642e382f5b38e70275726d1
ruby      https://github.com/asdf-vm/asdf-ruby.git   3a1f574c51292cc84f4d8893d101078cba89f640
```

### asdf plugins affected (if relevant)

python

daneroo:
Can confirm this behavior as well

Issue #1948, opened 2025-02-12
bug: 0.16.2 `asdf plugin update X` removes support files


markjreed:
### Describe the Bug

Several plugins populate subdirectories with support files - e.g.  nodejs/.node-build, perl/perl-install, python/pyenv. As of version 0.16.2, `asdf plugin update` is _emptying_ these subdirectories without _deleting_ them, breaking the on-demand population logic – and the plugins.

### Steps to Reproduce

1. asdf plugin update nodejs
2. asdf list all nodejs

### Expected Behaviour

asdf should display a list of versions available to install

### Actual Behaviour

```shell
Plugin nodejs's list-all callback script failed with output:
Trying to update node-build... ok
```

### Environment

```shell
Darwin ... 24.2.0 Darwin Kernel Version 24.2.0: Fri Dec  6 19:02:41 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6030 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.16.1

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/mjreed/.asdf
ASDF_CONFIG_FILE=/Users/mjreed/.asdfrc

ASDF INSTALLED PLUGINS:
[...]
nodejs    https://github.com/mkungla/asdf-nodejs               2858c702050785196f20c7f1557165e90a90f761
perl      https://github.com/ouest/asdf-perl.git               2b4609b9bed22a67c5d6a7c12e957fff3bcab434
python    https://github.com/danhper/asdf-python.git           a3a01856098d6d2b9642e382f5b38e70275726d1
[...]
```

### asdf plugins affected (if relevant)

nodejs, perl, python, possibly others.

Stratus3D:
Working on a fix. Should have something merged this week.

Issue #1882, opened 2025-02-04
feature: [0.16.0] support for private plugins `ASDF_GIT_USERNAME` and `ASDF_GIT_PASSWORD`


m4rii0:
### Is your feature request related to a problem? Please describe

Before 0.16.0 (Go migration), the tool supported the installation of private plugins (hosted in a private repository ), and the git client had access to it.

```bash
➜ asdf plugin add myplugin https://github.com/my-org/my-asdf-plugin
unable to clone plugin: authentication required: Repository not found.
```

With the migration to Go, relying on the [go-git](https://github.com/go-git/go-git) library, this feature is no longer working, since the tool is internally doing a PlainClone
https://github.com/asdf-vm/asdf/blob/e63aec61020a907fe5960d74b6d3dbf229214ae0/internal/git/git.go#L51

and before that, with the bash implementation, the git client was relying on the .gitconfiguration file, whether you have the authentication in a [git credential helper](https://git-scm.com/doc/credential-helper) or any other sort of configuration.

### Describe the proposed solution

While the go-git library does have support for authenticated calls to GitHub repositories with SSH and basic authentication (see [examples](https://github.com/go-git/go-git/tree/main/_examples/clone/auth) folder), asdf should adapt to each case and call the go-git API differently.

While having support for loading the user's .gitconfig folder in the go-git library would be the best solution, there is no support for that, and for the git credential helpers too as they are opened to receive contributions now, see https://github.com/go-git/go-git/issues/1420.

So I suggest having `ASDF_GIT_USERNAME` and `ASDF_GIT_PASSWORD` env variables, if they are present, send that credentials during the `asdf plugin add` and `asdf plugin update` commands.

### Describe similar `asdf` features and why they are not sufficient

n/a

### Describe other workarounds you've considered

I have not found any other workaround for making the plugin repository public, which is not an option right now.

Stratus3D:
Hi @m4rii0 , thanks for the detailed feature request! I'm wondering if maybe asdf should abandon go-git and use the installed `git` CLI client. That would allow individual users to configure their git client as needed and asdf would just use the same client. I've not tried this yet, but I think I could make any work. Any reason I should not try this? This would make version 0.16.0 work the same as 0.15.0 in this regard.

m4rii0:
@Stratus3D absolutely, it would work as long as the env variables that are in the shell configuration files are present while invoking the git command.

PR #1995, opened 2025-02-28
fix: preserve files untracked by Git on plugin update

andrecloutier:
# Summary

Related issues which describes the behavior we're seeing:
* https://github.com/go-git/go-git/issues/1031
* https://github.com/go-git/go-git/issues/796

Fixes: https://github.com/asdf-vm/asdf/issues/1980 
Fixes: https://github.com/asdf-vm/asdf/issues/1948

## Other Information

After change:
```
andre@Andres-MacBook-Pro asdf % asdf plugin remove python 
Successfully removed python

andre@Andres-MacBook-Pro asdf % asdf plugin add python
Successfully added python

andre@Andres-MacBook-Pro asdf % asdf latest python 
stackless-3.7.5

andre@Andres-MacBook-Pro asdf % ls -lah ~/.asdf/plugins/python/pyenv | wc -l
      29

andre@Andres-MacBook-Pro asdf % asdf plugin update python
Successfully updated python
updated python to ref refs/heads/master

andre@Andres-MacBook-Pro asdf % ls -lah ~/.asdf/plugins/python/pyenv | wc -l
      29

andre@Andres-MacBook-Pro asdf %  asdf latest python
stackless-3.7.5

```

Before change:
```
andre@Andres-MacBook-Pro asdf % asdf plugin remove python
Successfully removed python

andre@Andres-MacBook-Pro asdf % asdf plugin add python
Successfully added python

andre@Andres-MacBook-Pro asdf % asdf latest python
stackless-3.7.5

andre@Andres-MacBook-Pro asdf % ls -lah ~/.asdf/plugins/python/pyenv | wc -l
      29

andre@Andres-MacBook-Pro asdf % asdf plugin update python
Successfully updated python
updated python to ref refs/heads/master

andre@Andres-MacBook-Pro asdf % ls -lah ~/.asdf/plugins/python/pyenv | wc -l
       4

andre@Andres-MacBook-Pro asdf % asdf latest python 
No compatible versions available (python )
```


andrecloutier:
> Thanks for the PR @andrecloutier ! I'm actually considering switching back to the native CLI Git client (#1998) because it will fix this issue and also some of the authentication issues that have been reported (#1882). Thoughts?

@Stratus3D  Ah I was wondering what lead you down the path of wanting to rip it out. No objections from me with that approach if git-go doesn't support something asdf users rely on. 

As far as the fate of this PR, I guess it depends how far along you are with the move to the git cli PR and when you'd like to get the next release out. This PR is (IMO) safe and addresses a subset of the issues reported. If you want to get a patch release out sooner before the git cli move, I'd advocate for merging this in as a part of it. If not, then we can abandon this one.

augustobmoura:
I vote for merging this PR as it is and then working on the git command migration. We are having quite a few complaints on asdf-nodejs that could be fixed by this. I'm also wondering if we are doing any recursion of submodules on update, syncing and updating versioned git-submodules, do you know anything about it @Stratus3D and @andrecloutier?

I ask this, because I was thinking if pinning a node-build revision as a submodule, instead of cloning/updating it on the demand would also be a fix for asdf-nodejs

andrecloutier:
@augustobmoura I'm not sure off hand, I'd have to test it out. t seems like git-go supports sub-modules, so it should work. However, if you're keeping the compiled results in the plugins directory, my guess is without this change, those will get deleted on update which will also be annoying to users.

@Stratus3D What's your thoughts on merging this in and getting a 0.16.5 release out?

Stratus3D:
Thanks for the context @augustobmoura. I'd say no to submodules. We want the plugin update process to be as simple as possible. A single Git repo with no submodules is what I believe the current code supports. I'd stick with your current approach for node-build.

@andrecloutier I was going to close this but lets go ahead and merge. I expect to get this code replaced later this week but this will help in the meantime. I'm going to prepare a patch release right after merging this. Thanks!

Stratus3D (approved):

Stratus3D:
@raxod502-plaid possibly, https://github.com/asdf-vm/asdf/pull/1998 is my intended fix for these issues. #1998 should make the behavior in this regard the same as it was in version 0.15.0.

aaronjensen:
Am I correct that `plugin update` is completely defective? As far as I can tell, it doesn't actually update. We would have to uninstall the plugin and reinstall it to get the latest version (or use git directly in the plugin directly)

Stratus3D:
@aaronjensen it should be fixed by https://github.com/asdf-vm/asdf/pull/1998 which will go out in the next release.

aaronjensen:
Great, thank you.

Issue #1904, opened 2025-02-06
bug: asdf latest ruby no longer returns the latest stable standard version


farski:
### Describe the Bug

Previously, running something like `asdf latest ruby` would return the latest stable version of the standard flavor or Ruby. Now it seems to return the last item available in `asdf list all ruby `, which is unlikely what most people would expect.

### Steps to Reproduce

Run `asdf latest ruby`

### Expected Behaviour

Should return `3.4.1`

### Actual Behaviour

Returns `truffleruby+graalvm-24.1.2`

### Environment

```shell
OS:
Darwin Canyon 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:06 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T8103 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
3.2.57(1)-release

ASDF VERSION:
0.16.1

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/farski/.asdf
ASDF_CONFIG_FILE=/Users/farski/Repositories/GitHub/farski/zsh.d/asdf/.asdfrc

ASDF INSTALLED PLUGINS:
nodejs    https://github.com/asdf-vm/asdf-nodejs.git 93bd217ba74f05080eebb7d8e1c146a385b01d7f
ruby      https://github.com/asdf-vm/asdf-ruby.git   98e32b9c38e146b164d651c96ccb6d5fa4c3af90
```

### asdf plugins affected (if relevant)

`ruby`

adonig:
Same applies to the `elixir` plugin, see: https://github.com/asdf-vm/asdf-elixir/issues/135

ab:
Similar issue with `python` as well: https://github.com/asdf-community/asdf-python/issues/191

rkh:
Any updates on this? What is a safe asdf release to downgrade to?

Stratus3D:
Thanks for the report @farski. I will trying to find the cause of this sometime this week.

nacengineer:
[...]

Issue #1875, opened 2025-02-03
Add Our Own Regex FilterOut in latest Option or Add a Support for Regex Pattern


nizarakbarm:
### Is your feature request related to a problem? Please describe

I found out that the latest option in version 0.16.0 and 0.15.0 behave differently.

In asdf 0.15.0, the latest option correctly give the latest version, for example:

```
[nizarakb@kronos ~]$ asdf --version
v0.15.0-31e8c93
[nizarakb@kronos ~]$ asdf latest python
3.13.1t
```

But in version 0.16.0 the latest option behave differently and end up not give the real last version:
```
devnull@devnullnothing:~$ asdf --version
asdf version v0.16.0
devnull@devnullnothing:~$ asdf latest python
stackless-3.7.5
```

If i check the 0.15.0, versions.bash script always using:
```
  DEFAULT_QUERY="[0-9]"
```
but at the 0.16.0, it is not using the same DEFAULT_QUERY so it will result in stackless-3.7.5.

If it is intended to not use default query, i think it is needed for a feature to add our own FilterOut regeular expression or add our own regex pattern query.

### Describe the proposed solution

Adding an option for regex filter out or regex pattern query

### Describe similar `asdf` features and why they are not sufficient

asdf latest always return not the real latest version

### Describe other workarounds you've considered

I am not a go programmer, but i can only inform that there is a need for default query regex or an option to add default query regex

PR #1996, opened 2025-02-28
fix: latest version returns latest version

andrecloutier:
# Summary

Adjusting the filter to align with what the bash version of the application does here: https://github.com/andrecloutier/asdf/blob/d7e0740461a26bd3d437c2bfb02619d7e20d2489/lib/functions/versions.bash#L159

Effectively this code misses `grep -E "^\\s*[0-9]"` when filtering the results from `list-all`.

Fixes: https://github.com/asdf-vm/asdf/issues/1904
Fixes: https://github.com/asdf-vm/asdf/issues/1875


internal/versions/versions_test.go
```diff
@@ -338,6 +338,43 @@ func TestLatest(t *testing.T) {
...
+		assert.Nil(t, err)
+		contents := "#!/usr/bin/env bash\ncat \"" + versionsFilePath + "\""
+		listAllPath := filepath.Join(pluginDir, "bin", "list-all")
+		err = os.WriteFile(listAllPath, []byte(contents), 0o777)
+		assert.Nil(t, err)
+
+		plugin := plugins.New(conf, pluginName)
+		version, err := Latest(plugin, "")
+		assert.Nil(t, err)
+		assert.Equal(t, tt.expectedOutput, version)
```
Stratus3D, line 374: Thank you for adding tests for this ❤️

Stratus3D (approved):

augustobmoura:
Where is the filter for numeric numbers comes from? I couldn't find it on the [link provided](https://github.com/andrecloutier/asdf/blob/d7e0740461a26bd3d437c2bfb02619d7e20d2489/lib/functions/versions.bash#L159)

This broke latest version on protonge https://github.com/augustobmoura/asdf-protonge/issues/4, versions in ProtonGE are not numeric, and start with a preffix, because we are testing for versions starting with an number (or optionally spaces before numbers) we cannot find the version for protonge

augustobmoura:
It seeems that the filter should be applied _only_ when `latest-stable` is not being called, meaning it is only used when sorting `list-all`, if the `latest-stable` hook exists we should accept it independently of what is being printed. I will open a PR for this

andrecloutier:
> Where is the filter for numeric numbers comes from? I couldn't find it on the [link provided](https://github.com/andrecloutier/asdf/blob/d7e0740461a26bd3d437c2bfb02619d7e20d2489/lib/functions/versions.bash#L159)
> 
> This broke latest version on protonge [augustobmoura/asdf-protonge#4](https://github.com/augustobmoura/asdf-protonge/issues/4), versions in ProtonGE are not numeric, and start with a preffix, because we are testing for versions starting with an number (or optionally spaces before numbers) we cannot find the version for protonge

It's a bit convoluted to trace.
* Here's the default: https://github.com/andrecloutier/asdf/blob/d7e0740461a26bd3d437c2bfb02619d7e20d2489/lib/functions/versions.bash#L133
* Which gets passed into `list_all_command` here: https://github.com/andrecloutier/asdf/blob/d7e0740461a26bd3d437c2bfb02619d7e20d2489/lib/functions/versions.bash#L159C16-L159C32
* Which is then applied here: https://github.com/andrecloutier/asdf/blob/d7e0740461a26bd3d437c2bfb02619d7e20d2489/lib/functions/versions.bash#L111

Issue #1981, opened 2025-02-21
bug: if build dependencies are not met, asdf still records the version as being installed


blovett:
### Describe the Bug

using asdf 0.16.3 on ubuntu 24.04. i added the ruby plugin (`asdf plugin add ruby`) and attempted to install the latest ruby 3.0 via: `asdf install ruby latest:3.0` and it selected 3.0.7. i did not have all of the necessary build dependencies installed at the time, however (make, gcc, zlib1g-dev, etc), so after i fixed that i attempted the install again and found that version 3.0.7 was recorded as already being installed despite the plugin failing to complete the operation. a quick `asdf uninstall ruby 3.0.7` resolved it, but i'm wondering if this isn't unexpected behavior.

### Steps to Reproduce

- Install asdf 0.16.3 under ubuntu 24.04 (the release probably doesn't matter)
- Add the ruby plugin: `asdf plugin add ruby`
- Ensuring build dependencies are not installed, attempt to install a version of ruby (or perhaps any other plugin that has build/install dependencies): `asdf install ruby latest:3.0`.
- Note that the installation failed, but `asdf list ruby` shows it, and a subsequent install attempt reports an error that it is already installed.

### Expected Behaviour

asdf does not record versions as being installed unless the plugin completes successfully.

### Actual Behaviour

asdf records versions installed no matter the success.

### Environment

```shell
OS:
Linux blovett-linux-vm 6.11.0-17-generic #17~24.04.2-Ubuntu SMP PREEMPT_DYNAMIC Mon Jan 20 22:48:29 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
fish, version 3.7.0

BASH VERSION:
5.2.21(1)-release

ASDF VERSION:
v0.16.3

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/home/blovett/.asdf
ASDF_CONFIG_FILE=/home/blovett/.asdfrc

ASDF INSTALLED PLUGINS:
ruby      https://github.com/asdf-vm/asdf-ruby.git             3a1f574c51292cc84f4d8893d101078cba89f640
terraform https://github.com/asdf-community/asdf-hashicorp.git 22eb1c4a16adcde39aaaf89fbb5d9404a1601fce
```

### asdf plugins affected (if relevant)

ruby

Issue #1940, opened 2025-02-10
bug: Package versions show as installed even after build error (or invalid version specified)


unitof:
### Describe the Bug

Even if a version fails due to build error (possibly even due to an invalid version number being specified), asdf still lists it as installed.

### Steps to Reproduce

```
~ ▶ asdf list
java
  adoptopenjdk-20.0.2+9
 *liberica-8u282+8
nodejs
  18.19.0
  18.20.4
  20.16.0
  20.17.0
  23.6.0
  6.11.3
 *7.10.1
python
 *2.7.18
rust
  No versions installed

~ ▶ asdf install python 2
python-build 2 /Users/jacob/.asdf/installs/python/2
python-build: definition not found: 2
error installing version: failed to run install callback: exit status 2

~ ▶ asdf list
java
  adoptopenjdk-20.0.2+9
 *liberica-8u282+8
nodejs
  18.19.0
  18.20.4
  20.16.0
  20.17.0
  23.6.0
  6.11.3
 *7.10.1
python
  2
 *2.7.18
rust
  No versions installed

~ ▶ asdf uninstall python 2

~ ▶ asdf list
java
  adoptopenjdk-20.0.2+9
 *liberica-8u282+8
nodejs
  18.19.0
  18.20.4
  20.16.0
  20.17.0
  23.6.0
  6.11.3
 *7.10.1
python
 *2.7.18
rust
  No versions installed


### Expected Behaviour

`python 2` would not show as a listed installed version in `asdf list` after build error

### Actual Behaviour

`python 2` shows as installed version, even though it failed due to build error (as it is an invalid version)

### Environment

```shell
OS:
Darwin Kaylee.local 24.2.0 Darwin Kernel Version 24.2.0: Fri Dec  6 19:01:59 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
3.2.57(1)-release

ASDF VERSION:
0.16.2

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/jacob/.asdf
ASDF_CONFIG_FILE=/Users/jacob/.asdfrc

ASDF INSTALLED PLUGINS:
java      https://github.com/halcyon/asdf-java.git    8083bb707299b7f409371780d5855aca3be33d85
[...]

Issue #1932, opened 2025-02-09
bug: unsuccessful installation of NodeJS version


akkadaya:
### Describe the Bug

after the unsuccessful installation of NodeJS version, rerunning the same command will return the already installed version message.

it creates `~/.asdf/installs/nodejs/18.17.0` with empty content.

```sh
[10:03:32]❯ asdf install nodejs 18.17.0
Trying to update node-build...
WARNING: Updating node-build failed with exit code 128. The installation will
try to continue with already installed local defintions. To debug what went
wrong, try to manually update node-build by running: `asdf nodejs update nodebuild`

Binary for node-build not found
Are you sure it was installed? Try running `asdf nodejs update-nodebuild` to do a local update or install
error installing version: failed to run install callback: exit status 1

~
[10:03:44]❯ asdf install nodejs 18.17.0
error installing version: version {version 18.17.0} of nodejs is already installed
```

### Steps to Reproduce

1. run `asdf install nodejs 18.17.0`
2. returns error installing version due to some error in NodeJS plugin
3. run `asdf install nodejs 18.17.0`
4. returns error installing version because the version is already installed

### Expected Behaviour

It should not return an error installing version because the version is already installed

### Actual Behaviour

returns error installing version because the version is already installed

### Environment

```shell
OS:
Darwin xxx 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.16.2

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/xxx/.asdf
ASDF_CONFIG_FILE=/Users/xxx/.asdfrc

ASDF INSTALLED PLUGINS:
gcloud    https://github.com/jthegedus/asdf-gcloud.git 00cdf065aaf10f07ff3da8eba59746526e6a93c4
java      https://github.com/halcyon/asdf-java.git     53c037f335dcb77d013201aa4edae7c57811b781
nodejs    https://github.com/asdf-vm/asdf-nodejs.git   93bd217ba74f05080eebb7d8e1c146a385b01d7f
[...]

PR #2008, opened 2025-03-07
fix: improve handling of unsuccesful tool installations

ifpleti:
# Summary

This PR addresses incorrect displays of incomplete installations when using the `list` command and prevents erroneous ‘version already installed’ errors when attempting to reinstall after a failed or interrupted installation.

The PR refactors the installation verification process by renaming `IsInstalled` to `InstallDirExists`, making it clear that it only checks for the existence of the installation directory. In addition, a new function, `ShimExists`, has been added to verify that a proper shim exists for a given plugin version. The `InstallOneVersion` function now checks both the installation directory and the shim before considering a version already installed, **preventing false positives from partial installations**. The CLI output has also been updated (via `listLocalCommand`) to skip or mark versions that lack a valid shim. Unit tests have been added for `ShimExists` and existing tests have been updated accordingly. Finally, Visual Studio Code settings have been excluded from the repository via `.gitignore`.

Fixes: #1981
Fixes: #1940
Fixes: #1932

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ repository
 keyrings
 /tmp
 .idea
+.vscode
 
 dist/
 
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -437,7 +437,7 @@ func getVersionInfo(conf config.Config, plugin plugins.Plugin, currentDir string
 	if found {
 		firstVersion := toolversion.Versions[0]
 		version := toolversions.Parse(firstVersion)
-		installed = installs.IsInstalled(conf, plugin, version)
+		installed = installs.InstallDirExists(conf, plugin, version)
 	}
 	return toolversion, found, installed
 }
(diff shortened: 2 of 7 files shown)


Stratus3D:
Hi @ifpleti, thanks for the PR with the thorough description! I really appreciate you digging in here to fix this. However, I believe this PR does not address the root of the issue - install directories should not exist if a version install fails. If a version fails to install (that is, the `bin/install` callback exits with non-zero status) then the install directory should be removed by asdf. With the code prior to version 0.16.0 this was how it worked.

I think the code in `IsInstalled` is sufficient, we just need to make sure the install directory is properly removed after an install fails.

Also, not everyone uses Visual Studio Code, so the ignore rule for it should be removed from this repository. If you need to ignore it for your own workflow I'd suggest using a global ignore. I blogged about this here - http://stratus3d.com/blog/2018/06/03/stop-excluding-editor-temp-files-in-gitignore/

Stratus3D:
I believe the directory needs to be removed here - https://github.com/asdf-vm/asdf/blob/master/internal/versions/versions.go#L196-L199

Stratus3D:
https://github.com/asdf-vm/asdf/pull/2024 has fixed this same issue. These changes should not be needed anymore. If find something else that still needs to be fixed please open a new PR @ifpleti.

PR #2011, opened 2025-03-09
fix: correct handling of `ASDF_FORCE_PREPEND` environment variable

machupicchubeta:
# Summary

The error occurs if I set "yes" to `FORCE_PREPEND` env as described in the documentation.

## Error message
```
strconv.ParseBool: parsing \"yes\": invalid syntax
```

### Referenced Documents

- https://github.com/asdf-vm/asdf/blob/7352bf4890143184ba419392ca5e6204167c8306/docs/manage/configuration.md#asdf_force_prepend


internal/config/config_test.go
```diff
@@ -20,6 +21,23 @@ func TestLoadConfigEnv(t *testing.T) {
...
+		{"no"},
+	}
+	for _, c := range cases {
+		t.Setenv("ASDF_FORCE_PREPEND", c.envVar)
+		t.Run(fmt.Sprintf("When ASDF_FORCE_PREPEND env given %s", c.envVar), func(t *testing.T) {
+			config, err := loadConfigEnv()
+
+			assert.Nil(t, err, "Returned error when loading env for config")
+
+			assert.Zero(t, config.Home, "Shouldn't set Home property when loading config")
```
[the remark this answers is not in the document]
machupicchubeta: Thank you for your comment.
I have changed the test code to check that `ForcePrepend` is set the correct/expected value.
And I have check the initial value when `ASDF_FORCE_PREPEND` is not set.

https://github.com/asdf-vm/asdf/pull/2011/commits/942578e9aeac7ead7c8676a4686c71c4b41ae3ce

internal/config/config.go
```diff
@@ -59,8 +61,21 @@ type Settings struct {
...
+	case "":
+		forcePrepend = forcePrependDefault
+		if runtime.GOOS == "darwin" {
+			forcePrepend = true
+		}
+	case "yes":
+		forcePrepend = true
+	default:
+		forcePrepend = false
+	}
```
andrecloutier, line 76: I think the use of switch here adds unnecessary complexity. For instance, there's not longer any case where the value of `forcePrependDefault` is used since its overwritten by the default case.

I suggest using an if instead:

```
	forcePrependEnv := os.Getenv("ASDF_FORCE_PREPEND")
	if forcePrependEnv == "yes" || (forcePrependEnv == "" && runtime.GOOS == "darwin") {
		forcePrepend = true
	}
```

internal/config/config_test.go
```diff
@@ -22,6 +23,26 @@ func TestLoadConfigEnv(t *testing.T) {
...
+		defer os.Unsetenv("ASDF_FORCE_PREPEND")
+
+		config, _ := loadConfigEnv()
+
+		assert.True(t, config.ForcePrepend, "Then ForcePrepend property is true")
+	})
+
+	t.Run("When ASDF_FORCE_PREPEND env given any string other than yes", func(t *testing.T) {
+		os.Setenv("ASDF_FORCE_PREPEND", "no")
+		defer os.Unsetenv("ASDF_FORCE_PREPEND")
```
andrecloutier, line 38: Same comment here.

internal/config/config_test.go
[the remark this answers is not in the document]
andrecloutier: Looking good! Added a few minor comments. 

By the way... From searching the code, it looks like nothing is reading from `ForcePrepend`. So, while this PR will fix the parsing of the value, this won't actually result in any visible change in behavior in the application.

internal/config/config.go
machupicchubeta: Yes. that's exactly it!
Thank you!

https://github.com/asdf-vm/asdf/pull/2011/commits/d76ef9b13754529d203389080465b2007f050110

internal/config/config_test.go
```diff
@@ -22,6 +23,26 @@ func TestLoadConfigEnv(t *testing.T) {
 	assert.Zero(t, config.Home, "Shouldn't set Home property when loading config")
 }
 
+func TestLoadConfigEnv_WithForcePrependEnv(t *testing.T) {
+	t.Run("When ASDF_FORCE_PREPEND env given yes", func(t *testing.T) {
+		os.Setenv("ASDF_FORCE_PREPEND", "yes")
+		defer os.Unsetenv("ASDF_FORCE_PREPEND")
```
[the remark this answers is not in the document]
machupicchubeta: I see.
Thank you very much!
I fix the code.

https://github.com/asdf-vm/asdf/pull/2011/commits/a930763feef93813589e5ee92d291a75a52204f8

internal/config/config_test.go
machupicchubeta: It is the same as https://github.com/asdf-vm/asdf/pull/2011#discussion_r1991408214.

internal/config/config_test.go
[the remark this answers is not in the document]
machupicchubeta: @andrecloutier 
Thank you so much. 
There is nothing is reading `ForcePrepend`  .... Oh, indeed. I noticed that when you mentioned it.
I will consider that as another issue.

The reason I decided to write this code in the first place was because I saw the error on my local terminal.
For some reason I can't reproduce it now (.... hmm), but I hope to be able to prevent similar case.

internal/config/config.go
```diff
@@ -59,8 +61,13 @@ type Settings struct {
 }
 
 func defaultConfig(dataDir, configFile string) *Config {
+	forcePrepend := forcePrependDefault
+	forcePrependEnv := os.Getenv("ASDF_FORCE_PREPEND")
```
Stratus3D, line 65: We might want to be able to handle inconsistent case here. Some folks might set this to `YES` by mistake and I think that should be handled.

Stratus3D (approved):
Thanks for the PR @machupicchubeta ! I really appreciate the added tests!

machupicchubeta:
@andrecloutier @Stratus3D 
Thank you for the review and merge!

internal/config/config.go
machupicchubeta: I think it means case-insensitive. Sounds good to me.
I'll try to write a commit.

PR #2014, opened 2025-03-11
fix: correct concurrency to align with documentation

andrecloutier:
# Summary

Prior to this change:
* Default concurrency was "1" or "" (depending if there was an error). The [documentation suggests](https://asdf-vm.com/manage/configuration.html#full-configuration-example) this should be the number of cores.
* Setting "auto" concurrency would pass "auto" to install script. The [documentation suggests](https://asdf-vm.com/plugins/create.html#bin-install) this should be the number of cores.

Fixes: https://github.com/asdf-vm/asdf-nodejs/pull/409


Stratus3D:
Thanks for the fix @andrecloutier and thanks for the added tests!

Got any thoughts on this feature @andrecloutier ? I've been on the fence for years on this feature. It requires some amount of plumbing and documentation. asdf core has to properly read the config and set the `ASDF_CONCURRENCY` env variable (as you've fixed here) and then plugins are expected to read this env variable and use it somehow to choose the optimal concurrency for their builds. I feel like it'd be less overhead for us and not much more hassle for plugins to remove this feature from asdf corre and leave it up to plugins. Plugins can easily check the number of cores available and apply their own formula to determine their optimal concurrency. Since asdf core already passes all environment variables to plugin callbacks it'd be easy for each plugin to read it's own env variable (say `ASDF_NODEJS_CONCURRENCY`) and use it if set. I've never set `ASDF_CONCURRENCY`, and I doubt many users do.

andrecloutier:
> Thanks for the fix @andrecloutier and thanks for the added tests!
> 
> Got any thoughts on this feature @andrecloutier ? I've been on the fence for years on this feature. It requires some amount of plumbing and documentation. asdf core has to properly read the config and set the `ASDF_CONCURRENCY` env variable (as you've fixed here) and then plugins are expected to read this env variable and use it somehow to choose the optimal concurrency for their builds. I feel like it'd be less overhead for us and not much more hassle for plugins to remove this feature from asdf corre and leave it up to plugins. Plugins can easily check the number of cores available and apply their own formula to determine their optimal concurrency. Since asdf core already passes all environment variables to plugin callbacks it'd be easy for each plugin to read it's own env variable (say `ASDF_NODEJS_CONCURRENCY`) and use it if set. I've never set `ASDF_CONCURRENCY`, and I doubt many users do.

This feature to me reminds me of [MAKEOPTS](https://wiki.gentoo.org/wiki/MAKEOPTS) in Gentoo Linux where its useful to set compile settings globally at the OS/Package Manager level. In some cases, you may want to set it to the number of cores to just get the compile done as fast as possible, in some other cases, you want a more conservative number so that the computer is still usable. So I could see some utility here that folks _may_ want to control how many resources compiling their dependencies will take. The code to support this isn't crazy so I don't feel too bad about keeping it. Especially as we're trying to stabilize the transition from the bash version to go. But, I also haven't been maintaining this project for years. ;)

Stratus3D:
> Especially as we're trying to stabilize the transition from the bash version to go. But, I also haven't been maintaining this project for years. ;)

Yes good point. I've already been pushing the limits of what users will tolerate with the breaking changes introduced in v0.16.x. Probably best to continue supporting this feature. Thanks again for the fix @andrecloutier !

PR #2021, opened 2025-03-15
fix: improve error handling to always return a message

andrecloutier:
# Summary

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -126,8 +126,7 @@ func Execute(version string) {
 				Action: func(_ *cli.Context) error {
 					conf, err := config.LoadConfig()
 					if err != nil {
-						logger.Printf("error loading config: %s", err)
-						return err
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR @andrecloutier, as you probably noticed the `cli` package for this project is the most in need of refactoring. Almost none of the code is tested by Go tests. Some but not all of it is covered by BATS tests. I've love to structure it in such a way as to make testing via Go tests easy. Open to any suggestions you have.

andrecloutier:
> Thanks for the PR @andrecloutier, as you probably noticed the `cli` package for this project is the most in need of refactoring. Almost none of the code is tested by Go tests. Some but not all of it is covered by BATS tests. I've love to structure it in such a way as to make testing via Go tests easy. Open to any suggestions you have.

Yeah I made that observation as well. I'll have to play around with it a bit to see what makes sense to make this testable.

PR #1933, opened 2025-02-10
fix: improve zsh completion suggestions

Valgard:
improve zsh completion suggestions with flags, git refs, versions, etc.

PR #2022, opened 2025-03-17
fix: improve zsh completion suggestions

Stratus3D:
Replaces https://github.com/asdf-vm/asdf/pull/1933


internal/completions/asdf.zsh
```diff
@@ -92,110 +101,246 @@ _asdf__installed_versions_of_plus_system() {
...
+          'plugin:plugin:_asdf__installed_plugins'
+        ;;
+      4)
+        if [[ ${words[3]} == -* ]]; then
+          # After flag, complete with plugin name
+          _asdf__installed_plugins
+        else
+          # Complete with available versions for the plugin
+          local versions
+          if versions=$(asdf list all "${words[3]}" 2>/dev/null); then
```
jonatan-ivanov, line 315: Before the change this listed only the versions that were installed, after this change, it lists everything (same as line 330). See https://github.com/asdf-vm/asdf/issues/2134

PR #2024, opened 2025-03-19
fix: remove install directory for version when install fails

andrecloutier:
# Summary

PR based on feedback in https://github.com/asdf-vm/asdf/pull/2008 . This PR will remove the install directory for the given version if it fails to install (install command returns a non-zero exit code).

Fixes: https://github.com/asdf-vm/asdf/issues/1981
Fixes: https://github.com/asdf-vm/asdf/issues/1940


Stratus3D (approved):
Thank you for fixing this @andrecloutier ! I've been meaning to fix this but I've been in the process of moving to a new laptop and haven't been able to get tests running yet.

PR #2026, opened 2025-03-19
chore: Be tolerant to `ASDF_FORCE_PREPEND` environment variable

machupicchubeta:
# Summary

I agree that it would be valid even if it is set in uppercase as well as lowercase.

@Stratus3D commented:
- https://github.com/asdf-vm/asdf/pull/2011#discussion_r1996208949
> We might want to be able to handle inconsistent case here. Some folks might set this to YES by mistake and I think that should be handled.

## Other Information

I think this code change did not need to be mentioned in the spec, so I did not add it there. What do you think? I would like to hear the committer's opinion.

Current Specification is:
- https://github.com/asdf-vm/asdf/blob/932ac468b7c24c2adef90a293a1f7280a0074cc4/docs/manage/configuration.md#L208-L215


internal/config/config.go
```diff
@@ -63,7 +64,7 @@ type Settings struct {
 func defaultConfig(dataDir, configFile string) *Config {
 	forcePrepend := forcePrependDefault
 	forcePrependEnv := os.Getenv("ASDF_FORCE_PREPEND")
-	if forcePrependEnv == "yes" || (forcePrependEnv == "" && runtime.GOOS == "darwin") {
+	if matched, _ := regexp.MatchString(`yes|YES`, forcePrependEnv); matched || (forcePrependEnv == "" && runtime.GOOS == "darwin") {
```
andrecloutier, line 67: Using regex for this is overkill. Suggest using `strings.ToLower(forcePrependEnv) == "yes"` instead.

machupicchubeta: @andrecloutier @Stratus3D 
Thank you for the comment and emoji reaction.
Yes, I know. I prefer a simple implementation too, so if I can do it that way, that's the way to go.

However, I'm concerned that it implementation might be a bit too loose since it would allow typo strings like “yEs” and “yeS” while “yes” and “YES” and “Yes” would be have consistency, but I'm not sure if that's ok.
Is that too much of a concern?

Stratus3D: I'm not concerned about that. If someone wants to specify odd casing that is fine with me. It will keep the implementation simpler.

machupicchubeta: @Stratus3D 
Alright! I’ll do it that way. Thanks!

machupicchubeta: @andrecloutier @Stratus3D 
Would you please code review again?

Stratus3D (approved):

machupicchubeta:
@andrecloutier @Stratus3D 
Thank you for the review and merge! 🎉

PR #2027, opened 2025-03-20
fix: remove default error action from plugin command

bowd:
# Summary

The plugin command was showing an error message when no subcommand was provided, which was not necessary since the command has valid subcommands (add, list, etc). This PR removes this default error action, allowing the command to properly handle subcommands and show the help message when no subcommand is specified.

Fixes: N/A

## Other Information

This improves UX by making `asdf plugin` less error-y and more useful


bowd:
(It would be so funny if this was some sort of weird legacy thing that can't be changed)

Stratus3D (approved):

PR #2028, opened 2025-03-20
fix: Be aware of `ForcePrepend`

machupicchubeta:
# Summary

Previously, the value of `ForcePrepend` was not used anywhere and was not behaving as specified.
With this commit, the behavior will be as specified.

### Referenced Documents
- https://github.com/asdf-vm/asdf/blob/932ac468b7c24c2adef90a293a1f7280a0074cc4/docs/manage/configuration.md#asdf_force_prepend

### Background

@andrecloutier told about this issue me in a comment on another PullRequest. Thank you!
- https://github.com/asdf-vm/asdf/pull/2011#discussion_r1990316326


Stratus3D:
Hi @machupicchubeta , thanks for this PR and your previous one. I should have given more thought before merging https://github.com/asdf-vm/asdf/pull/2011 and https://github.com/asdf-vm/asdf/pull/2026. The force prepend option was only ever used one place - https://github.com/asdf-vm/asdf/blob/master/asdf.sh#L63. This file was a Bash script used in version 0.15.0 and older. It was sourced by the shell to expose an `asdf` function that wrapped the actual executable and allowed certain commands to alter the user's environment. We've migrated fully to Go, and have decided to eliminate the use of all shell code in asdf itself. All that remains is a couple lines of shell-specific config code that users need to add to their shell's RC file.

The force prepend option was never actually used in the internal code for asdf, even in the older versions. It was only ever used for altering the `PATH` variable in the user's current shell session. Given this information, this PR isn't actually restoring old behavior, it's actually altering behavior that's existed for many version. I don't think there is actually a problem here anymore that can be solved with an option like this. If my understanding of this is flawed or incomplete please let me know, but I think it would be best if I remove all the force prepend option from the Go code completely. It was initially just a workaround for MacOS I believe.

Stratus3D:
Closing for now in favor of #2089

machupicchubeta:
@Stratus3D 
Thank you for the detailed explanation and for taking the time to reflect on this!
Your reasoning makes perfect sense, and I agree with your assessment. I appreciate the context around the legacy shell code and the move to Go — that definitely clarifies things.
I’m happy to see asdf evolving, and I look forward to contributing again in the future.
Thanks again for your thoughtful response!

PR #2030, opened 2025-03-20
fix: remove filtering from latest-stable call

augustobmoura:
# Summary

Partially reverts https://github.com/asdf-vm/asdf/pull/1996, when the `latest-stable` script hook is available, we should use it at face value without any filtering. The hook was made specifically to customize the default `list-all` sortering and filtering. Reference can be found [here](https://github.com/andrecloutier/asdf/blob/d7e0740461a26bd3d437c2bfb02619d7e20d2489/lib/functions/versions.bash#L150-L156).

I also reordered the code to make it clear that list-all is being used as a fallback when `latest-stable` fails. This is **not** the same behavior as the bash version, where if it has an `latest-stable` hook it throws an error when it fails, but I think the new approach is better. 

fixes https://github.com/augustobmoura/asdf-protonge/issues/4

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -240,33 +240,31 @@ func Latest(plugin plugins.Plugin, query string) (version string, err error) {
 	var stdErr strings.Builder
 
 	err = plugin.RunCallback("latest-stable", []string{query}, map[string]string{}, &stdOut, &stdErr)
-	if err != nil {
-		if _, ok := err.(plugins.NoCallbackError); !ok {
-			return version, err
-		}
-
-		allVersions, err := AllVersionsFiltered(plugin, query)
-		if err != nil {
-			return version, err
-		}
-
-		versions := filterOutByRegex(allVersions, numericStartFilterRegex, true)
-		versions = filterOutByRegex(versions, latestFilterRegex, false)
-
+	if err == nil {
+		versions := parseVersions(stdOut.String())
 		if len(versions) < 1 {
 			return version, errors.New(noLatestVersionErrMsg)
 		}
-
 		return versions[len(versions)-1], nil
 	}
 
-	// parse stdOut and return version
-	allVersions := parseVersions(stdOut.String())
+	// Fallback to list-all if latest-stable fails
+	if _, ok := err.(plugins.NoCallbackError); !ok {
+		return version, err
+	}
+
+	allVersions, err := AllVersionsFiltered(plugin, query)
+	if err != nil {
+		return version, err
+	}
+
 	versions := filterOutByRegex(allVersions, numericStartFilterRegex, true)
 	versions = filterOutByRegex(versions, latestFilterRegex, false)
+
 	if len(versions) < 1 {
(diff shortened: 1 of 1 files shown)


internal/versions/versions.go
```diff
@@ -240,33 +240,31 @@ func Latest(plugin plugins.Plugin, query string) (version string, err error) {
...
-
-		allVersions, err := AllVersionsFiltered(plugin, query)
-		if err != nil {
-			return version, err
-		}
-
-		versions := filterOutByRegex(allVersions, numericStartFilterRegex, true)
-		versions = filterOutByRegex(versions, latestFilterRegex, false)
-
+	if err == nil {
```
andrecloutier, line 243: This approach will result in all errors from "latest-stable" falling back to the "all versions" approach. Not just cases where the file/script isn't present. I'd suggest either adding a `HasCallback` method to determine which script/approach to rely on or making this decision on a "file not found" error response from `RunCallback`.

augustobmoura: I mentioned this in the summary of the PR. This is already the case, so I didn't want to tackle it on this change specifically. I just want to close https://github.com/augustobmoura/asdf-protonge/issues/4. I did reversed the if condition to make it more explicit that it fallbacks to `list-all` ordering, the code is otherwise unchanged aside from the removal of filters when latest-stable is called.

Adding `HasCallback` is also out of scope for this PR

andrecloutier: ah sorry I should I read your description more thoroughly! Actually I think the behavior that's here now aligns with what was in the bash script version since it only falls through to the "all versions" implementation when a `plugins.NoCallbackError` error is returned which is only the case if "latest-stable" is missing.

andrecloutier: A test for this case would be helpful to prevent a regression in the future.

Stratus3D (commented):
Thanks for the PR and the detailed description @augustobmoura ! I think this is a good fix we should include in the next patch release, but I do want to add a test for this as the existing tests didn't catch this. If you can add tests that'd be great, otherwise I'll write a couple tomorrow and then merge this.

augustobmoura:
> Thanks for the PR and the detailed description @augustobmoura ! I think this is a good fix we should include in the next patch release, but I do want to add a test for this as the existing tests didn't catch this. If you can add tests that'd be great, otherwise I'll write a couple tomorrow and then merge this.

Yeah, tests makes sense, I will try writing some after I brush up my golang skills

internal/versions/versions.go
augustobmoura: Ah, I didn't notice the if condition for `NoCallbackError` (well, I did notice it but couldn't understand that it is a type check). So yeah, the code should have the same behavior as the bash version now

Stratus3D:
I've got it taken care of @augustobmoura !

Stratus3D:
Replaced by https://github.com/asdf-vm/asdf/pull/2032 which contains all the commits here plus a test that I verified failed on the `master` branch.

PR #2034, opened 2025-03-21
feat: print requested version on fail , closes #2033

### LINKED ISSUE #2033: Please add the requested version to tool selection output

If a version file (e.g. .terraform-version) requests a version, that is not installed, asdf outputs a list of all versions that are installed:

```
No version is set for command terraform
Consider adding one of the following versions in your config file at /home/hfi/git/gitlab.boreus.de/kubernetes/terraform/bcp_clusters_vmware/bcp-prod/bcp-boreus-test-salt/.tool-versions
terraform 1.7.5
terraform 1.2.7
terraform 1.11.2
terraform 1.10.5
```

It would be nice if that output would also contain the version that was actually requested:

```sh
cat .terraform-version 
1.10.4
```

So maybe you could just add an additional line like:

```
No version is set for command terraform
Requested version was: 1.10.4
Consider adding one of the following versions in your config file at /home/hfi/git/gitlab.boreus.de/kubernetes/terraform/bcp_clusters_vmware/bcp-prod/bcp-boreus-test-salt/.tool-versions
terraform 1.7.5
terraform 1.2.7
terraform 1.11.2
terraform 1.10.5
```

-

-

hfi-wiit:
As requested in #2033, this PR just adds a few lines to print the requested version in the event the related binary version is not installed.


carlqt:
Unsure what we need to do to get this merged. It's a very simple fix and the change is appreciated. I hate the current message

internal/shims/shims.go
```diff
@@ -102,7 +107,7 @@ func FindExecutable(conf config.Config, shimName, currentDirectory string) (stri
 	}
 
 	if len(existingPluginToolVersions) == 0 {
-		return "", plugins.Plugin{}, "", false, NoVersionSetError{shim: shimName}
+		return "", plugins.Plugin{}, requestedVersion, false, NoVersionSetError{shim: shimName}
```
Stratus3D, line 116: I think this should only return a version as the 3 value if a matching one is found. Is there a reason this return value was changed?

Stratus3D (commented):
Thanks for the PR @hfi-wiit ! Sorry for the late review, I'm only now catching up on PRs. The changes appear good, but we need tests before we merge to ensure this works as expected in all cases and so to prevent future regressions. I'll circle back to this in about a week and write tests then. In the meantime you are welcome to contribute tests if you want.

Stratus3D:
Also I think we may want to address https://github.com/asdf-vm/asdf/issues/1467 in this as well.

Issue #2035, opened 2025-03-22
bug: ZSH completion is not workin in version 0.16.6


akkadaya:
### Describe the Bug

It was working before the update to v0.16.6

I did follow the instruction again from [here](https://asdf-vm.com/guide/getting-started.html#set-up-shell-completions-optional-3
) but in vain

### Steps to Reproduce

1. type `asdf install` and press Tab key

### Expected Behaviour

it should auto-complete the command

### Actual Behaviour

it shows directories/files in the current directory

### Environment

```shell
OS:
Darwin XXX 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.16.6

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/XXX/.asdf
ASDF_CONFIG_FILE=/Users/XXX/.asdfrc

ASDF INSTALLED PLUGINS:
gcloud    https://github.com/jthegedus/asdf-gcloud.git 00cdf065aaf10f07ff3da8eba59746526e6a93c4
java      https://github.com/halcyon/asdf-java.git     d5c78d613835c62277c48973a55975e3217a6c23
nodejs    https://github.com/asdf-vm/asdf-nodejs.git   5515b8705fbf4d610ede04bb9a3fc4812e3a4168
perl      https://github.com/ouest/asdf-perl.git       2b4609b9bed22a67c5d6a7c12e957fff3bcab434
python    https://github.com/danhper/asdf-python.git   a3a01856098d6d2b9642e382f5b38e70275726d1
ruby      https://github.com/asdf-vm/asdf-ruby.git     c4c9674a7769dfe1cdadfcb40eddb3b15269ca22
zig       https://github.com/cheetah/asdf-zig.git      9d11e5b4a4fee4ddf47caf739b1f5ec8d6862384
```

Stratus3D:
I am able to reproduce this. Will try to fix in the next couple of days. PRs welcome!

roeniss:
we have this now: https://github.com/ohmyzsh/ohmyzsh/pull/13026

jonatan-ivanov:
[...]

PR #2037, opened 2025-03-24
fix: remove comment from first line zsh completion output (#2035)

massahud:
# Summary

The first line of a completion file must be `#compdef`, removed the comment that was added above it.

Fixes: #2035


Stratus3D (approved):

Stratus3D:
Thanks for the PR @massahud !

Issue #2001, opened 2025-03-02
bug: "ref:foo" regression in golang rewrite


elijahr:
### Describe the Bug

Greetings! I am the author of https://github.com/asdf-community/asdf-nim. I make regular use of the `ref:foo` tool identifier syntax to install nightly builds of certain git branches when available, or fallback to building from source when not available.

`ref:foo` seems to install ok, but asdf fails to find or use the installed version:

```shell
export ASDF_NIM_VERSION=ref:version-2-2
asdf install nim # works
asdf which nim # fails with `No nim executable found for nim`
asdf reshim
asf which nim # still fails with `No nim executable found for nim`
```

If I do this, it does show that the the version has been installed (with ":" replaced by "-"):

```shell
$ asdf list nim
  ref-version-2-2
```

So, the `ref:` identifier works okay for installation, but not for shims or calling `which`.  

There is a hacky workaround, which is setting `ASDF_NIM_VERSION=ref-version-2-2` after `ref:version-2-2` has been installed.

### Steps to Reproduce

Here is a minimal Dockerfile to reproduce:

```Dockerfile
FROM debian:bookworm-slim

ENV ASDF_DATA_DIR="/root/.asdf"
ENV PATH="${ASDF_DATA_DIR}/shims:/root/.nimble/bin:/usr/local/go/bin:$PATH"
# 2.2 nightly build
ENV ASDF_NIM_VERSION="ref:version-2-2"

# Install dependencies
RUN apt-get update && apt-get install -y \
    curl \
    git \
    build-essential \
    libssl-dev \
    coreutils \
    && rm -rf /var/lib/apt/lists/*

# Install golang (req of asdf)
RUN curl -fsSL https://go.dev/dl/go1.24.0.linux-arm64.tar.gz | tar -C /usr/local -xz

# Install asdf, asdf-nim, and nim
RUN git clone https://github.com/asdf-vm/asdf.git --depth 1 --branch v0.16.4 && \
    cd asdf && \
    make && \
    ln -s ${PWD}/asdf /usr/local/bin/asdf && \
    asdf plugin add nim && \
    asdf install && \
    asdf reshim

# Fails with:
# No executable nim found for current version. Please select a different version or install nim manually for the current version
RUN nim -v
```

### Expected Behaviour

asdf should find and execute nim using `ref:version-2-2` in ASDF_NIM_VERSION or a .tool-versions file.

### Actual Behaviour

[...]

PR #2049, opened 2025-04-02
fix: support ref installed tool shims

Morgahl:
# Summary

I commonly make use of `ref:` based installs when working in `erlang` and `elixir` repositories, however this appears to be broken in the current builds as the `ref:` prefix is normalized to `ref-` when working with file path rendering the ref based installs unlocatable by the shim.

Fixes: #2001

## Other Information

I do not know if there is a better place to do this, I am happy to make the changes in the correct area if you feel it is needed.

### DIFF
--- a/internal/installtest/installtest.go
+++ b/internal/installtest/installtest.go
@@ -66,18 +66,18 @@ func InstallOneVersion(conf config.Config, plugin plugins.Plugin, versionType, v
 
 // InstallPath returns the path to a tool installation
 func InstallPath(conf config.Config, plugin plugins.Plugin, version string) string {
-	return filepath.Join(pluginInstallPath(conf, plugin), toolVersion(version))
+	return filepath.Join(pluginInstallPath(conf, plugin), formatVersionStringFromFS(version))
 }
 
 // DownloadPath returns the download path for a particular plugin and version
 func DownloadPath(conf config.Config, plugin plugins.Plugin, version string) string {
-	return filepath.Join(conf.DataDir, dataDirDownloads, plugin.Name, toolVersion(version))
+	return filepath.Join(conf.DataDir, dataDirDownloads, plugin.Name, formatVersionStringFromFS(version))
 }
 
 func pluginInstallPath(conf config.Config, plugin plugins.Plugin) string {
 	return filepath.Join(conf.DataDir, dataDirInstalls, plugin.Name)
 }
 
-func toolVersion(version string) string {
+func formatVersionStringFromFS(version string) string {
 	return toolversions.FormatForFS(toolversions.Parse(version))
 }
--- a/internal/installtest/installtest.go
+++ b/internal/installtest/installtest.go
@@ -66,18 +66,18 @@ func InstallOneVersion(conf config.Config, plugin plugins.Plugin, versionType, v
 
 // InstallPath returns the path to a tool installation
 func InstallPath(conf config.Config, plugin plugins.Plugin, version string) string {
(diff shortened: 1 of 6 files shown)


internal/installtest/installtest.go
```diff
@@ -65,14 +66,18 @@ func InstallOneVersion(conf config.Config, plugin plugins.Plugin, versionType, v
...
 func DownloadPath(conf config.Config, plugin plugins.Plugin, version string) string {
-	return filepath.Join(conf.DataDir, dataDirDownloads, plugin.Name, version)
+	return filepath.Join(conf.DataDir, dataDirDownloads, plugin.Name, toolVersion(version))
 }
 
 func pluginInstallPath(conf config.Config, plugin plugins.Plugin) string {
 	return filepath.Join(conf.DataDir, dataDirInstalls, plugin.Name)
 }
+
+func toolVersion(version string) string {
```
Stratus3D, line 81: I think this function name could be improved. Maybe something like `formatVersionStringForFS`?

internal/shims/shims.go
```diff
@@ -411,7 +411,12 @@ func parse(contents string) (versions []toolversions.ToolVersions) {
 			segments := strings.Split(line, " ")
 			// if doesn't have expected number of elements on line skip
 			if len(segments) >= 4 {
-				versions = append(versions, toolversions.ToolVersions{Name: segments[2], Versions: []string{segments[3]}})
+				versions = append(versions, toolversions.ToolVersions{
+					Name: segments[2],
+					Versions: []string{
+						toolversions.VersionFromFSFormat(segments[3]),
```
Stratus3D, line 417: Should shims contains the file system format or the actual version string? If they contained the non-FS version I don't think we'd need this.

internal/shims/shims_test.go
```diff
@@ -100,6 +100,29 @@ func TestFindExecutable(t *testing.T) {
...
+
+		executable, gotPlugin, version, found, err := FindExecutable(conf, "dummy", currentDir)
+		assert.Nil(t, err)
+		assert.True(t, found)
+		assert.Equal(t, "ref:v1.1.0", version)
+		assert.Equal(t, plugin, gotPlugin)
+		assert.Equal(t, "dummy", filepath.Base(executable))
+		assert.Equal(t, "ref-v1.1.0", filepath.Base(filepath.Dir(filepath.Dir(executable))))
+	})
+}
```
Stratus3D, line 124: Thanks for writing a test for this!

Stratus3D (commented):
Thanks for the PR @Morgahl ! I left a couple comments, and the linter is complaining about one line, but I think your PR is pretty close to what we're going to need. I'm going to do some manual testing myself tomorrow and report back on what I see.

I'm thinking shim files should contain the `ref:v1.2.3` syntax to match `.tool-versions` files, but maybe not.

internal/installtest/installtest.go
Morgahl: Good call.

internal/shims/shims.go
Morgahl: I have reverted this in favor of fixing this from the `installs.Installed` directory read to move this to the source of this version string within the system. This did surface a formatting issue with the ref strings `toolversions.Format`. I think this case might have just been missed in all the changes.

Morgahl:
I have made the requested changes and moved the FS format reversion to the source of the FS format string within the system for consistency. This also surfaced a small change to `toolversions.Format`.

PR #2039, opened 2025-03-25
chore: release 0.16.7

Stratus3D:
:robot: I have created a release *beep* *boop*
---

### Bug Fixes

* remove comment from first line zsh completion output ([#2035](https://github.com/asdf-vm/asdf/issues/2035)) ([#2037](https://github.com/asdf-vm/asdf/issues/2037)) ([74d7b17](https://github.com/asdf-vm/asdf/commit/74d7b17a1cc9f640cf0f5134416d1cf7a56fe19f))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #2050, opened 2025-04-03
docs: use release-please-manifest.json for version on documentation website

albertchae:
# Summary

Instead of manually maintaining a separate `docs/version.txt` that can
easily fall out of date, parse the version out of release please's
generated manifest.

Example of recent release PR that updates this https://github.com/asdf-vm/asdf/pull/2039

I'm not sure if it's OK to touch files outside of the docs directory,
but it looks like `version.txt` used to live one level above until this
PR https://github.com/asdf-vm/asdf/pull/1960, so unless something else
changed this seems safe?

## Other Information

I noticed the doc site had an out of date version while checking my asdf version


Stratus3D:
> I'm not sure if it's OK to touch files outside of the docs directory, but it looks like version.txt used to live one level above until this PR #1960, so unless something else changed this seems safe?

I'm not sure either but I'll find out.

PR #2052, opened 2025-04-04
chore(docs): add binary download link to guide page

hyudai28:
## Summary
This pull request addresses an issue in the guide where users are instructed to download the binary, but no direct link is provided. This omission has caused confusion. With this change, a direct link to the binary is added to enhance clarity and user experience.


Stratus3D:
Thanks for the PR @hyudai28 !

PR #2055, opened 2025-04-05
Add flag to current command that auto installs missing versions

mangas:
# Summary

Add a --install-missing flag to the current command so it can immediately install all the missing plugin versions.


Stratus3D:
Closing as we don't want `asdf current` doing anything other than listing set versions.

Issue #2010, opened 2025-03-07
Consider renaming --parent to --parents


jrz:
### Is your feature request related to a problem? Please describe

Just upgraded to 0.16. 
I noticed the --parent option with asdf set.

Now, the documentation/help text is actually clear, but from just the option, I thought it would set in the tool-version of the direct parent (which doesn't really make sense)

### Describe the proposed solution

Change --parent in --parents

### Describe similar `asdf` features and why they are not sufficient

asdf --parent 

### Describe other workarounds you've considered

not applicable

Stratus3D:
I've got no objection to this. PR welcome!

PR #2058, opened 2025-04-07
feat: rename set command's `--parent` flag to `--parents`

jrz:
# Summary

Changes --parent to --parents. Updated docs etc.
It adds a fallback --parent, but will warn about the change. 
Note that the old option still appears in --help (Hence the Hidden: false in cli.go. That can be set to true, or just removed.)

#2010 Consider renaming --parent to --parents

## Other Information
Please check if you want the Hidden: true or not.

I noticed tests fail, but the same tests fail from a fresh master.
I couldn't really think of a test, as the option itself is only used as -p in the test.

Lastly, in the contribute/core.html I noticed the instructions said bats test/list_commands.bash instead of bats test/list_commands.bats


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 * remove comment from first line zsh completion output ([#2035](https://github.com/asdf-vm/asdf/issues/2035)) ([#2037](https://github.com/asdf-vm/asdf/issues/2037)) ([74d7b17](https://github.com/asdf-vm/asdf/commit/74d7b17a1cc9f640cf0f5134416d1cf7a56fe19f))
 
 ## [0.16.6](https://github.com/asdf-vm/asdf/compare/v0.16.5...v0.16.6) (2025-03-21)
-
+* rename --parent argument to --parents ([#2010](https://github.com/asdf-vm/asdf/issues/2010))
```
Stratus3D, line 11: Please revert this change. CHANGELOG.md entries are added automatically.

jrz:
Reverted, and squashed some of the commits.
The last one could be a separate chore.

First commit is for global renaming + a bit or rewording
Second commit is the actual code

PR #2067, opened 2025-04-18
feat(golang-rewrite): add support for shim templates resolution

### LINKED ISSUE #2025: bug: shim creation doesn't honor custom shims
### Describe the Bug

As mentioned here https://asdf-vm.com/plugins/create.html#custom-shim-templates, it should be possible to provide a "shims/<executable>" template inside a plugin.
It seems the go version of asdf does not contain the necessary code yet to honor this.

Is there another way to fix shims that doesn't work with the default code?

### Steps to Reproduce

- create a plugin with a shims directory
- install the executable
- do a asdf reshim
- check the shim, that was actually created against the plugin provided one

### Expected Behaviour

- the provided shim to be used

### Actual Behaviour

- default shim used

### Environment

```shell
OS:
Linux hfi-ukw 6.8.0-55-generic #57-Ubuntu SMP PREEMPT_DYNAMIC Wed Feb 12 23:42:21 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
GNU bash, Version 5.2.21(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2022 Free Software Foundation, Inc.
Lizenz GPLv3+: GNU GPL Version 3 oder jünger <http://gnu.org/licenses/gpl.html>

Dies ist freie Software.  Sie darf verändert und verteilt werden.
Es wird keine Garantie gewährt, soweit das Gesetz es zulässt.

BASH VERSION:
5.2.21(1)-release

ASDF VERSION:
v0.16.5

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/home/hfi/.asdf
ASDF_CONFIG_FILE=/home/hfi/.asdfrc
```

- a self written plugin
- potentially asdf-elixir ?

augustobmoura:
# Summary

In this PR I'm adding support for [custom shim templates](https://asdf-vm.com/plugins/create.html#custom-shim-templates) back, since it is missing from the current go implementation.
As of now, the implementation short-circuits very early if it finds an acceptable shim in the first tool available.

Plugins that I know are using shim templates are: [asdf-elixir](https://github.com/asdf-vm/asdf-elixir/tree/master/shims), [asdf-nodejs](https://github.com/asdf-vm/asdf-nodejs/tree/master/shims), and [asdf-python](https://github.com/asdf-community/asdf-python/tree/master/shims).

fixes #2025 asdf-vm/asdf-nodejs#421

### DIFF
--- a/internal/shims/shims.go
+++ b/internal/shims/shims.go
@@ -56,7 +56,7 @@ func (e NoExecutableForPluginError) Error() string {
 
 // FindExecutable takes a shim name and a current directory and returns the path
 // to the executable that the shim resolves to.
-func FindExecutable(conf config.Config, shimName, currentDirectory string) (string, plugins.Plugin, string, bool, error) {
+func FindExecutable(conf config.Config, shimName, currentDirectory string) (path string, plugin plugins.Plugin, version string, found bool, err error) {
 	shimPath := Path(conf, shimName)
 
 	if _, err := os.Stat(shimPath); err != nil {
--- a/internal/shims/shims.go
+++ b/internal/shims/shims.go
@@ -74,6 +74,11 @@ func FindExecutable(conf config.Config, shimName, currentDirectory string) (stri
 	for _, shimToolVersion := range toolVersions {
 		plugin := plugins.New(conf, shimToolVersion.Name)
 		if plugin.Exists() == nil {
+			// If a shim template is found, we can return it before looping through versions
+			shimTemplate, err := plugin.ShimTemplatePath(shimName)
+			if err == nil {
+				return shimTemplate, plugin, "", true, nil
+			}
 
 			versions, found, err := resolve.Version(conf, plugin, currentDirectory)
 
--- a/internal/shims/shims.go
+++ b/internal/shims/shims.go
@@ -373,18 +373,25 @@ func ToolExecutables(conf config.Config, plugin plugins.Plugin, version toolvers
 
 // ExecutablePaths returns a slice of absolute directory paths that tool
 // executables are contained in.
-func ExecutablePaths(conf config.Config, plugin plugins.Plugin, version toolversions.Version) ([]string, error) {
+func ExecutablePaths(conf config.Config, plugin plugins.Plugin, version toolversions.Version) (paths []string, err error) {
 	dirs, err := ExecutableDirs(plugin)
 	if err != nil {
 		return []string{}, err
 	}
 
+	shimsDir, err := os.Stat(path.Join(plugin.Dir, "shims"))
+	if err == nil && shimsDir.IsDir() {
+		paths = append(paths, path.Join(plugin.Dir, "shims"))
+	}
+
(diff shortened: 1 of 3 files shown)


Stratus3D:
Hi @augustobmoura . I'm not actually sure custom shim templates are needed for Elixir. The plugin seems to work fine for me without the custom shim for `mix`.

Was I wrong in my understanding when I removed this feature? See https://asdf-vm.com/guide/upgrading-to-v0-16.html#custom-shim-templates-are-no-longer-supported

augustobmoura:
> Was I wrong in my understanding when I removed this feature? See https://asdf-vm.com/guide/upgrading-to-v0-16.html#custom-shim-templates-are-no-longer-supported

Shim templates are still being used by asdf-nodejs and asdf-python for auto-reshimming after running global installs.

We could think of a different mechanism for this, instead of shim templates. We need a way of intercepting calls to commands and conditionally marking them for auto-reshiming after the command is finished. Similar to [plugin hooks](https://asdf-vm.com/manage/configuration.html#plugin-hooks), but defined at the plugin level, instead of user level.

Maybe we could implement a way of having plugin hooks at mutiple levels? plugin, user, and maybe system level for example. And all of them are called when the hook is triggered, wdyt?

Stratus3D:
@augustobmoura thanks for the clarification. I guess I'd be inclined to add back in support for custom shim templates. I'm not a fan of how asdf hooks are designed. And hooks are not controlled by the plugin, they are managed by the user. I'm going to work to get this PR merged soon.

internal/shims/shims.go
```diff
@@ -368,18 +373,25 @@ func ToolExecutables(conf config.Config, plugin plugins.Plugin, version toolvers
...
+func ExecutablePaths(conf config.Config, plugin plugins.Plugin, version toolversions.Version) (paths []string, err error) {
 	dirs, err := ExecutableDirs(plugin)
 	if err != nil {
 		return []string{}, err
 	}
 
+	shimsDir, err := os.Stat(path.Join(plugin.Dir, "shims"))
+	if err == nil && shimsDir.IsDir() {
+		paths = append(paths, path.Join(plugin.Dir, "shims"))
+	}
```
Stratus3D, line 385: Are these 4 lines necessary?

Stratus3D (commented):
Overall I think the changes here are great. My only concern is that I don't think shim template code belongs in the `ExecutablePaths` function as that is only suppose to be the executable paths returned by the `list-bin-paths` plugin callback.

internal/shims/shims.go
Stratus3D: I've added this code back into https://github.com/asdf-vm/asdf/pull/2076 which addresses the linter warnings present here. Closing this PR in favor of that one (it still contains all your commits).

PR #2068, opened 2025-04-18
test(golang-rewrite): ignore global git config on repotest

augustobmoura:
# Summary

I noticed that while running tests, a dialog asking for my GPG key password was popping up. This is because I sign all my commits, and because repotests is using the `git` command behind the scenes using my global `~/.gitconfig`.

In this PR I'm setting the 

Fixes: List issue numbers here `GIT_CONFIG_GLOBAL` variable to `/dev/null` for all commands run on repotest, so it ignores the [global user config](https://github.com/git/git/commit/4179b4897f2de28858acaebd6382c06c91532e98), which to me seems logical, since we are using git just for a simple test.

### DIFF
--- a/internal/repotest/repotest.go
+++ b/internal/repotest/repotest.go
@@ -206,6 +206,10 @@ func runCmd(cmdName string, args ...string) error {
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 
+	// Global env vars
+	// GIT_CONFIG_GLOBAL=/dev/null prevents git from looking for user settings like commit.gpgSign and user.name
+	cmd.Env = []string{"GIT_CONFIG_GLOBAL=/dev/null"}
+
 	err := cmd.Run()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


augustobmoura:
> Thanks for the PR @augustobmoura , I'm not sure this makes sense to accept as it might interfere with authentication that is needed for companies installing their own private plugins.

I'm only changing the test helper, it shouldn't impact users

> It's not clear to me why you are being prompted for you GPG key password. Typically the password is only required when you need to use the key for something. Why is your key needed to install a repository via Git?

We are commiting in the tests, I'm being asked my GPG key because of https://github.com/asdf-vm/asdf/blob/92de803ff15f1a887f031d570ae6404f008d829d/internal/repotest/repotest.go#L141

Stratus3D (approved):

Stratus3D:
> I'm only changing the test helper, it shouldn't impact users

I created that file so I should have remembered that. It lacking the `_test.go` suffix always throws me off.

Issue #1983, opened 2025-02-24
bug: plugin version setting validation


kkawula:
### Describe the Bug

asdf should display error message if set version do not exist

```shell
asdf set --home scarb 2.92
```
do not produce any output

```shell
asdf set --home scarb 2.9.2
```
this one should work

### Steps to Reproduce

```shell
asdf set --home scarb 2.92
```

### Expected Behaviour

Any error message

### Actual Behaviour

Blank

### Environment

```shell
macos
```

Stratus3D:
Changing this from a bug report to a feature request as I consider this a UX improvement rather than broken behavior.

PR #2070, opened 2025-04-26
feat: add version validation when using set command

jrodriigues:
# Summary

These changes add version validation when using `asdf set <plugin> <version>` to ensure that only available versions for the plugin are added.
If the version is not part of the available list, the program will fail with a descriptive error message:
```zsh
❯ ./asdf set ruby 3.9.0
version 3.9.0 is not available for plugin ruby
```

Fixes: #1983

## Other Information

I refactored some existing code from the `list` command into a new function on the `set` package since the objective on both commands is the same.


jrodriigues:
I believe that the new function to retrieve a plugin's available versions makes sense to be a `plugin` method rather than a standalone function in the `set` package, so I pushed a change for it.

Stratus3D (commented):
Thanks for the PR @jrodriigues ! I think this is nice improvement. I think introducing that `GetAvailableVersions` method was a good idea.

I'd like to merge this but I see it's failing linter and test checks. Can you take a look and fix them?

jrodriigues:
@Stratus3D are you able to verify if the failure is due to workflow?

Issue #2046, opened 2025-03-31
bug: `go` installed `asdf version` outputs `v-dev`


runephilosof-abtion:
### Describe the Bug

`asdf version` should output the correct version when installed through `go install`

### Steps to Reproduce

1. `go install github.com/asdf-vm/asdf/cmd/asdf@v0.16.7`
2. `~/go/bin/asdf version`

### Expected Behaviour

Should output `0.16.7`

### Actual Behaviour

It outputs `v-dev`

### Environment

```shell
OS:
Linux fedora 6.13.8-200.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Mar 23 05:03:09 UTC 2025 x86_64 GNU/Linux

SHELL:
GNU bash, version 5.2.32(1)-release (x86_64-redhat-linux-gnu)
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

BASH VERSION:
5.2.32(1)-release

ASDF VERSION:
v-dev

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/home/user/.asdf
ASDF_CONFIG_FILE=/home/user/.asdfrc

ASDF INSTALLED PLUGINS:
nodejs    https://github.com/asdf-vm/asdf-nodejs.git 2bd35bcf4bd017dc364fa70a22f8650f14e88ced
ruby      https://github.com/asdf-vm/asdf-ruby.git   87eac3928dac8d4f7a972e32149b933099936584
yarn      https://github.com/twuni/asdf-yarn.git     376c5407ef22c82fa56e4bd9f6a9108354063acb
```

Stratus3D:
Sorry for the late reply here. Working on a fix.

Stratus3D:
I've created https://github.com/asdf-vm/asdf/pull/2077 for this. Not confident it's sufficient but it's much simpler than what we have now. Thoughts?

PR #2077, opened 2025-05-02
fix: set correct version for go install and make builds

Stratus3D:
Intended to address https://github.com/asdf-vm/asdf/issues/2046 and related issues.

With these changes the version will always look something like this:

```
$ ./asdf version
0.16.7 (revision 610d084)
```

### DIFF
--- a/cmd/asdf/main.go
+++ b/cmd/asdf/main.go
@@ -1,12 +1,40 @@
 // Main entrypoint for the CLI app
 package main
 
-import "github.com/asdf-vm/asdf/internal/cli"
+import (
+	"fmt"
+	"runtime/debug"
+	"slices"
 
-// Replaced with the real version during a typical build
-var version = "v-dev"
+	"github.com/asdf-vm/asdf/internal/cli"
+)
+
+// Do not touch this next line
+var version = "0.16.7" // x-release-please-version
 
 // Placeholder for the real code
 func main() {
-	cli.Execute(version)
+	fullVersion := buildFullVersion(version)
+	cli.Execute(fullVersion)
+}
+
+func buildFullVersion(version string) string {
+	var shortRef string
+
+	info, ok := debug.ReadBuildInfo()
+
+	if !ok {
+		panic("can't get build info")
+	}
+	idx := slices.IndexFunc(info.Settings, func(s debug.BuildSetting) bool {
+		return s.Key == "vcs.revision"
+	})
+
+	if idx < 0 {
+		shortRef = "unknown"
+	} else {
+		shortRef = info.Settings[idx].Value[0:7]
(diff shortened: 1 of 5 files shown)


cmd/asdf/main.go
```diff
@@ -3,8 +3,8 @@ package main
 
 import "github.com/asdf-vm/asdf/internal/cli"
 
-// Replaced with the real version during a typical build
-var version = "v-dev"
+// Do not touch this next line
+var version = "0.16.7" // x-release-please-version
```
Stratus3D, line 7: Not confident what I've got here is correct, but the intent is that release-please would update this version upon release.

https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files

cmd/asdf/main.go
```diff
@@ -1,12 +1,31 @@
...
+func buildFullVersion(version string) string {
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		panic("can't get build info")
+	}
+	idx := slices.IndexFunc(info.Settings, func(s debug.BuildSetting) bool {
+		return s.Key == "vcs.revision"
+	})
+
+	return fmt.Sprintf("%s (revision %s)", version, info.Settings[idx].Value[0:7])
```
Stratus3D, line 30: The Git revision is important as it disambiguates dev builds of the master branch. For example, I merged this code and someone built latest master `asdf version` would print something like:

```
0.16.7 (revision 610d084)
```

But the `0.16.7` tag (if this logic were backported) would actually print:

```
0.16.7 (revision cbe8c6b)
```

Most users won't be building asdf on their own. But for those that do and then open up issues we'll be able to uniquely identify the commit they've build based on the revision.

cmd/asdf/main.go
runephilosof-abtion: Why not use `buildInfo.Main.Version` instead (as suggested by @Nouuu in https://github.com/asdf-vm/asdf/issues/2046#issuecomment-2812127919)?
That way you won't need the complexity of having a release phase step.

Stratus3D: That depends on the Git repo being present with the source code at compile time. The archives available on GitHub **do not** contain the Git repo so when someone compiles that source code the version it would print would be "unknown" if I followed what is in the blog post:

```go
if buildInfo.Main.Version != "" {
		fmt.Printf("Version: %s\n", buildInfo.Main.Version)
	} else {
		fmt.Println("Version: unknown")
	}
``` 

It's very important for maintainers for all versions of the code to print a useful version regardless of how the binary was created.

I'm happy to make adjustments here if we find this approach to be insufficient.

Issue #2095, opened 2025-05-17
bug: error messages printed without newline cause shell prompt to appear on the same line


nukhes:
### Describe the Bug

When printing error messages using the printError function, the output does not automatically add a newline (\n) at the end. This causes the shell prompt to appear immediately after the error message on the same line.

### Steps to Reproduce

1. Install asdf.
2. Do something wrong like `asdf set`.

### Expected Behaviour

Error messages should end with a newline so that the shell prompt appears on the next line.

### Actual Behaviour

![Image](https://github.com/user-attachments/assets/969d7955-a3de-48a9-9b6e-964a8bc26fd1)

### Environment

```shell
OS:
Linux alves-A320MH 6.8.0-60-generic #63-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 15 19:04:15 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

SHELL:
GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

BASH VERSION:
5.2.21(1)-release

ASDF VERSION:
0.16.7

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/home/alves/.asdf
ASDF_CONFIG_FILE=/home/alves/.asdfrc

ASDF INSTALLED PLUGINS:
nodejs    https://github.com/asdf-vm/asdf-nodejs.git fd2c7f94c4c71416047cacf2af6e4fa819b364b7
python    https://github.com/danhper/asdf-python.git a3a01856098d6d2b9642e382f5b38e70275726d1
```

Stratus3D:
Thanks for the report. Working on getting these changes merged.

PR #2096, opened 2025-05-17
fix: error message formatting by adding newline in printError

nukhes:
# Summary

- Added a check to append a newline to the error message if it's missing before printing.

- Trimmed the newline when creating the returned error value to keep error messages clean and consistent.

Fixes: #2095 

## Other Information

This prevents error messages from running into the shell prompt, improving readability.


Stratus3D (commented):
Got some failing code here. I will try to fix and merge this on Monday.

Issue #2108, opened 2025-05-30
bug: Unhelpful error message when trying to use tool version that's not installed


ZimbiX:
### Describe the Bug

Context: I started investigating this when I noticed `ruby 3.3.6` ending up in a random file.

Running `ruby` (in `~`, i.e. `/Users/brendan.weibrecht`) logs this error:

```
➜ ruby
No version is set for command ruby
Consider adding one of the following versions in your config file at /Users/brendan.weibrecht/.tool-versions
ruby 3.3.6
```

The last line is to stdout; so when in a script I was running a Ruby one-liner that suppresses stderr and redirects stdout to a file, I end up with that file containing just `ruby 3.3.6`.

```
➜ ruby 2>/dev/null    
ruby 3.3.6
```

I think this line should be to stderr as well.

Regardless, that's probably a separate issue from this message being unhelpful for my situation.

The message is unhelpful because:

- I do have a version set for Ruby, it's just that this version is not yet installed
- `ruby 3.3.6` is already in `/Users/brendan.weibrecht/.tool-versions`

### Steps to Reproduce

```
➜ cat ~/.tool-versions
ruby 3.4.4

➜ asdf list ruby
  3.3.6
 *3.4.4

➜ mv ~/.asdf/installs/ruby/3.4.4 ~/asdf-ruby-3.4.4-disabled

➜ asdf list ruby
  3.3.6

➜ ruby
No executable ruby found for current version. Please select a different version or install ruby manually for the current version

➜ asdf reshim

➜ ruby       
No version is set for command ruby
Consider adding one of the following versions in your config file at /Users/brendan.weibrecht/.tool-versions
ruby 3.3.6
```

### Expected Behaviour

Showing an error saying something accurate and helpful - like that the version of Ruby that's set is not installed and that I should run `asdf install ruby 3.3.6`

### Actual Behaviour

An unhelpful error, as said above

### Environment

```shell
➜ asdf info
OS:
Darwin M25017 24.4.0 Darwin Kernel Version 24.4.0: Wed Mar 19 21:17:35 PDT 2025; root:xnu-11417.101.15~1/RELEASE_ARM64_T6041 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
3.2.57(1)-release

ASDF VERSION:
0.17.0 (revision unknown)

ASDF INTERNAL VARIABLES:
[...]

PR #2109, opened 2025-05-31
fix: print all error output to stderr when shim can't be resolved

Stratus3D:
Fixes https://github.com/asdf-vm/asdf/issues/2108

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -666,16 +666,16 @@ func getExecutable(logger *log.Logger, conf config.Config, command string) (exec
 				logger.Printf("No preset version installed for command %s", command)
 				for _, toolVersion := range toolVersions {
 					for _, version := range toolVersion.Versions {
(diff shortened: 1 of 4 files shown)


internal/cli/cli.go
```diff
@@ -666,16 +666,16 @@ func getExecutable(logger *log.Logger, conf config.Config, command string) (exec
...
 				}
 
-				fmt.Printf("or add one of the following versions in your config file at %s/.tool-versions\n", currentDir)
+				logger.Printf("or add one of the following versions in your config file at %s/.tool-versions\n", currentDir)
 			}
 
 			for _, toolVersion := range toolVersions {
 				for _, version := range toolVersion.Versions {
-					fmt.Printf("%s %s\n", toolVersion.Name, version)
+					logger.Printf("%s %s\n", toolVersion.Name, version)
```
Stratus3D, line 678: This changes cause these lines to be printed to STDERR instead of STDOUT.

Issue #1880, opened 2025-02-04
Upgrading docs should document cleaning the old (0.15) version files


corran2go:
### How can we improve the documentation?

After updating to `0.16` there are still the old files in `.asdf` like `version.txt` etc.

Could you shorty list which files/dirs should be kept to leave the installed plugins and versions in place while cleaning the old asdf files.

PR #2114, opened 2025-06-04
docs: document cleanup of ASDF_DATA_DIR in upgrade guide

Stratus3D:
Fixes: https://github.com/asdf-vm/asdf/issues/1880

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ repository
 .vagrant
 keyrings
 /tmp
-.idea
 
 dist/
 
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -5,7 +5,6 @@ repository
 .vagrant
 keyrings
 /tmp
-.idea
```
Stratus3D, line 8: Editor files should be kept out of project gitignores. See https://stratus3d.com/blog/2018/06/03/stop-excluding-editor-temp-files-in-gitignore/

PR #2118, opened 2025-06-05
chore: enable windows binaries

### LINKED ISSUE #1973: WSL2 Migration Guide v0.16
### How can we improve the documentation?

Cannot upgrade to v0.16. I have tried the following:

* Turn off WSL, delete Ubuntu WSL, reinstall WSL
* Delete Ubuntu WSL and reinstall
* Upgrading Without Losing Data section

I still get this error message with the `asdf info` command:

>NOTICE:
You have tried to upgrade to asdf 0.16.0 or newer. Versions 0.16.0 is a
complete rewrite of asdf in Go. This text is being printed by the older
Bash implementation. If you are seeing this you have not migrated to
asdf 0.16.0. Please follow the instructions on the upgrade guide to
migrate to the new version.

>Aside from this notice, this older Bash implementation works as it did
in asdf version 0.15.0 and older.

>Migration guide: https://asdf-vm.com/guide/upgrading-to-v0-16
asdf website: https://asdf-vm.com
Source code: https://github.com/asdf-vm/asdf

Please add steps/commands for migration. It's vague and provides zero guidance on how to do the migration (ex. tar, cp, ex...).

Stratus3D:
Fixes https://github.com/asdf-vm/asdf/issues/1973


Stratus3D:
More work on CI tests is needed before this can be merged.

CAOShurong:
I tested the current head `7b8266f21f7a944936e34890ff514f8f580d22f3` on native Windows/amd64 with Go 1.26.5 and an isolated `ASDF_DATA_DIR`. To avoid conflating the result with the fixture's Bash executable, I replaced the installed dummy tool with a copy of the native Windows `whoami.exe` named `dummy.exe`, then ran `reshim`.

Observed:

```text
plugin add / install / reshim / set   exit 0
asdf which dummy.exe                  exit 0, resolves the installed dummy.exe
installed dummy.exe directly          exit 0, prints the current Windows identity
asdf exec dummy.exe                   exit 1, empty stdout and stderr
generated dummy.exe shim via Git Bash exit 1, empty stdout and stderr
```

The executable is therefore installed and resolved, but the final execution path fails. At this head, `internal/cli/cli.go:621` calls `internal/exec.Exec`, which calls `syscall.Exec`; Go's Windows implementation returns `syscall.EWINDOWS` unconditionally. Generated shims end in `exec asdf exec ...`, so they cannot complete that final execution path on native Windows even after install, reshim, and which succeed.

I made no changes to the PR; this is verification against the pinned SHA.

PR #2139, opened 2025-06-19
fix: env variable version resolution to match legacy behavior

imranismail:
# Summary

We still use shell-lived tool session in our workflows, however when migrating to manually setting the behavior using env var, we noticed this issue.

This fix ensures the pattern matches the legacy shell behavior

https://github.com/asdf-vm/asdf/blob/2114f1ebc15b3f1653228c8b416f759898a58dc7/lib/commands/command-export-shell-version.bash#L20

### DIFF
--- a/internal/resolve/resolve.go
+++ b/internal/resolve/resolve.go
@@ -131,5 +131,5 @@ func parseVersion(rawVersions string) []string {
 }
 
 func variableVersionName(toolName string) string {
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR @imranismail and sorry for the extremely late review! These same changes were merged from https://github.com/asdf-vm/asdf/pull/2194 and will go out in the next patch release. Closing this PR as it is no longer needed.

PR #2147, opened 2025-07-10
docs: add asdf Dev Container Feature to community projects

Alegrowin:
Adds information about the asdf Dev Container Feature to the community projects documentation. This feature allows installing asdf in any dev container, works with both pre-0.16.0 and 0.16.0+ versions, and can auto-install tools from .tool-versions file.


Stratus3D:
Hi @Alegrowin ! Thanks for creating this asdf dev container! It's great to see more support around asdf being created. However, given the newness of the project, and given the fact that there is currently a more popular asdf dev container available, I'm not going to add a link to the official docs yet. If your project grows in popularity, or becomes the only option going forward I'd be happy to add a link at that time. Thank you for your contributions.

PR #2154, opened 2025-07-22
fix: Print stderr for list-all / latest-stable callbacks

CGA1123:
# Summary

This updates usages of `list-all` and `latest-stable` plugin callbacks to dump their stderr to `asdf`'s stderr, so that plugins can provide more context around why plugin installation may have failed during version resolution.


Stratus3D:
Thanks for the PR @CGA1123 ! I made a small update to these changes and then merged them. They will go out in the next release.

Issue #2156, opened 2025-07-28
bug: Nushell's `get --ignore-errors` flag is deprecated.


rbuchberger:
### Describe the Bug

Currently nushell users on v0.106.0 or later who have set up completions get a warning every time they open a new shell:

```nu
Warning: nu::parser::deprecated

  ⚠ Flag deprecated.
    ╭─[/home/robert/.asdf/completions/nushell.nu:98:33]
 97 │                             where enabled |
 98 │                             get --ignore-errors template |
    ·                                 ───────┬───────
    ·                                        ╰── get --ignore-errors was deprecated in 0.106.0 and will be removed in a future release.
 99 │                             str join '' |
    ╰────
  help: This flag has been renamed to `--optional (-o)` to better reflect its behavior.
```

### Steps to Reproduce

1. Install nushell v0.106.0 or later.
2. Follow setup instructions, including completions
3. Open a new shell prompt.

### Expected Behaviour

No deprecation warning. 

### Actual Behaviour

Deprecation warning.

### Environment

```shell
❯ asdf info
OS:
Linux charybdis 6.15.8-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Thu, 24 Jul 2025 18:17:55 +0000 x86_64 GNU/Linux

SHELL:
0.106.0

BASH VERSION:
5.3.3(1)-release

ASDF VERSION:
0.18.0-arch (revision unknown)

ASDF INTERNAL VARIABLES:
ASDF_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/home/robert/.asdf
ASDF_CONFIG_FILE=/home/robert/.asdfrc

ASDF INSTALLED PLUGINS:
nodejs    https://github.com/asdf-vm/asdf-nodejs.git
          93bd217ba74f05080eebb7d8e1c146a385b01d7f
pnpm      https://github.com/jonathanmorley/asdf-pnpm.git
          032ae1a9c3f59f87af20ed00da0b873e1322173e
python    https://github.com/danhper/asdf-python.git
          d4caa7d4832313b4c141ee0ceca4118747147e89
ruby      https://github.com/asdf-vm/asdf-ruby.git
          3062b6c559206b955b2ff30a37f5a77c7757cd07

~ 
❯
```

rbuchberger:
For anyone looking for a workaround until this is fixed and released, you can update your completion script locally:

[...]

PR #1681, opened 2023-11-20
fix(nushell): Use `def --env` instead of `def-env`

hyperupcall:
# Summary

Nushell recently deprecated use of `def-env`. Migration is clear. [This](https://github.com/nushell/nushell/pull/10715) issue implements the deprecation and [this](https://github.com/nushell/nushell/pull/10566) issue implements `def --env`. It states:

> Because `def-env` is used everywhere, including virtualenv, I propose the following steps:
>
> - 0.86: Add this PR, keep `def-env` and `extern-wrapped` unchanged.
> - 0.87: Add deprecation warning to `def-env` and `extern-wrapped` but keep them working.
> - 0.88 (or later): Remove `def-env` and `extern-wrapped`.

So, the minimum version needs to be bumped to 0.86. Unfortunately, this means older versions will no longer be support, but it is difficult as nushell has many backwards incompatible changes being made

Current version of nushell is [0.87.0](https://github.com/nushell/nushell/releases/tag/0.87.0), which was released last week. To ensure this fix gets merged before the release of 0.88.0, it __should be merged before ~December 13__

Closes #1680
Closes #1700

PR #2157, opened 2025-07-28
fix: nushell completions --ignore-errors deprecation

rbuchberger:
# Summary

Update nushell completion code to use the new flag.

Fixes: #2156

## Other Information

Thanks for the great tool, I've been using it to earn a living for years.

### DIFF
--- a/internal/completions/asdf.nushell
+++ b/internal/completions/asdf.nushell
@@ -95,12 +95,12 @@ module asdf {
         let template = '(?P<name>.+)' + (
                             $params |
                             where enabled |
-                            get --ignore-errors template |
+                            get --optional template |
                             str join '' |
                             str trim
                         )
 
-        let flags = ($params | where enabled | get --ignore-errors flag | default '' )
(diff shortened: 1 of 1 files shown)


rbuchberger:
Wait, I've just realized the fix might need to be a little more complicated. It looks like nushell introduced the new flag and deprecated the old one in the same release. Merging this PR would break completions for anyone on earlier versions. I'm not sure how you want to handle that?

hyperupcall:
@rbuchberger Before, fixing the issue did mean breaking completions for people on older versions: https://github.com/asdf-vm/asdf/pull/1681. The `scripts/install_dependencies.bash` was also modified to reflect the new, expected version.

I think it would be difficult to justify the work for a fix that doesn't break compatability, especially since Nushell is a pretty niche and fast-moving project.

rbuchberger:
Yeah, that sounds pretty reasonable to me. I don't think too many people are hanging out on old versions of nushell and unable to update. When I first made this issue 0.106.0 was brand new, but now they're several releases past that.

PR #2163, opened 2025-08-13
feat: remove multiple plugins at once

joaovictorsl:
# Why?

I wanted to remove multiple plugins on a single command

# How?

Make pluginRemoveCommand take a slice instead of only the first plugin name, then we loop through them and remove one by one. If any error happens when removing a plugin we return before finishing the loop.


joaovictorsl:
Is this ok or is there a reason why removing multiple plugins at once would go wrong?

Stratus3D:
Hi @joaovictorsl, sorry for the late reply here! I've had this PR sitting in a browser tab for months, and I finally got around to looking at this feature this week.

I understand the convenience of uninstalling multiple plugins at once. But, I think adding this feature would cause some problems. This change makes the `plugin remove` command inconsistent with other commands, which still only accept a single tool or version. We could of course update all of the sub-commands to accept multiple last arguments (tools or versions typical), but then this leads to other questions, like how should `asdf current <tool1> <tool2> <tool3>` behave? And other commands, like `asdf plugin add <name> [<git-url>]` can only accept arguments for a single plugin. For this reason I think it makes sense to close this PR.

I'm opening to consider changes to multiple commands that keeps them consistent with each and allows for operation on multiple plugins or versions. But I think this is something that requires more thought and planning to make a good user experience. If you want to give this more thought please open a feature request listing all commands, documenting the ones you think would make sense to accept multiple plugins/versions as arguments.

Issue #2134, opened 2025-06-14
bug: Completion of set command uses all available versions not just the installed ones


jonatan-ivanov:
### Describe the Bug

The (zsh) completion of the `set` subcommand uses the `asdf list all "$plugin"` command to list the available versions to set. The problem is that `asdf list all "$plugin"` lists all the available versions (to install), not only the installed ones. This makes the completion of `set` useless since that subcommand should only need to browse the installed versions.

For example, let's say you use the `java` plugin, and you have three Java versions installed. When you use `asdf set java`, you will get 1948 options as of today not three:
```sh
asdf list all java | wc -l
    1948
```

See
https://github.com/asdf-vm/asdf/blob/2114f1ebc15b3f1653228c8b416f759898a58dc7/internal/completions/asdf.zsh#L314
and
https://github.com/asdf-vm/asdf/blob/2114f1ebc15b3f1653228c8b416f759898a58dc7/internal/completions/asdf.zsh#L329

### Steps to Reproduce

`asdf set <plugin-name><tab><tab>`

### Expected Behaviour

Listing the installed versions, the output of `asdf list "$plugin"` instead of `asdf list all "$plugin"`.

### Actual Behaviour

Listing every version available, not only the versions that are installed.

### Environment

```shell
❯ asdf info
OS:
Darwin <hostname> 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:43 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T8132 arm64

SHELL:
zsh 5.9 (arm64-apple-darwin24.0)

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
0.18.0 (revision unknown)

ASDF INTERNAL VARIABLES:
ASDF_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/<user>/.asdf
ASDF_CONFIG_FILE=/Users/<user>/.asdfrc

ASDF INSTALLED PLUGINS:
java      https://github.com/halcyon/asdf-java.git
          e18c5fcf11d990392abd59a71faa299e8b2e9a21
nodejs    https://github.com/asdf-vm/asdf-nodejs.git
          fd2c7f94c4c71416047cacf2af6e4fa819b364b7
```

### asdf plugins affected (if relevant)

all

PR #2172, opened 2025-08-28
Fix set version completion

### LINKED ISSUE #2047: bug: Regression in Zsh Completion for asdf Java set Command (v0.16.6 & v0.16.7)
### Describe the Bug

I have identified a regression in asdf versions 0.16.6 and 0.16.7 that affects the Zsh completion functionality. Specifically, when using the set command to configure the current Java JDK version, the completion system incorrectly suggests all installable JDK versions instead of only listing the currently installed JDK versions.

This issue was not present in asdf version 0.16.5, where the command completion correctly displayed only the JDK versions installed on the system.

### Steps to Reproduce

1. Install asdf v0.16.6 or v0.16.7.
2. Use Zsh as your shell.
3. Run the following command and attempt auto-completion:
4. asdf set java <TAB>
5. Observe that the completion suggests all available JDKs (installable), rather than only the installed ones.

Image

6. Downgrade to asdf v0.16.5.
7. Run the same command again.
8. Observe that the correct behavior is restored, showing only installed JDK versions.

Image

### Expected Behaviour

When using the set command in Zsh completion, only the JDK versions installed on the system should be suggested.

Image

### Actual Behaviour

When using the set command in Zsh completion, in versions 0.16.6 and 0.16.7, the completion instead lists all installable JDK versions.

Image

### Environment

```shell
Macos 15.3.2
zsh 5.9 (x86_64-apple-darwin24.0)
Homebrew 4.4.27-3-gfaddb73
```

java

zhirzh:
# Summary

Currently, zsh completion of the `set` command show list of all available versions.
This fix will only show the installed (and "system") versions.

Fixes #2047, #2134

### before

Screenshot 2025-08-28 at 3 53 23 PM

### after

Screenshot 2025-08-28 at 3 55 00 PM

### DIFF
--- a/internal/completions/asdf.zsh
+++ b/internal/completions/asdf.zsh
@@ -320,12 +320,7 @@ case "$subcmd" in
         else
           local plugin="${words[3]}"
         fi
-        local versions
-        if versions=$(asdf list all "$plugin" 2>/dev/null); then
-          _wanted "versions-$plugin" \
-            expl "Available versions of $plugin" \
-            compadd -- ${(f)versions}
-        fi
+        _asdf__installed_versions_of_plus_system $plugin
         ;;
     esac
     ;;


jonatan-ivanov:
@Stratus3D Could you please take a look, this is a somewhat annoying bug if you need to set versions frequently.

danhper:
Thanks!!

PR #2177, opened 2025-09-04
fix: restore backward compatibility of ASDF_{TOOL}_VERSION env vars

reegnz:
# Summary

In the bash version it was not only uppercasing, but also replacing `-` with `_` (because `-` is not a valid character for env vars).

See original code for reference:
https://github.com/asdf-vm/asdf/blame/master/lib/commands/command-export-shell-version.bash#L20-L21

### DIFF
--- a/internal/resolve/resolve.go
+++ b/internal/resolve/resolve.go
@@ -131,5 +131,5 @@ func parseVersion(rawVersions string) []string {
 }
 
 func variableVersionName(toolName string) string {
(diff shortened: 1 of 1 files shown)


Stratus3D:
Thanks for the PR @reegnz and sorry for the extremely late review! These same changes were merged from https://github.com/asdf-vm/asdf/pull/2194 and will go out in the next patch release. Closing this PR as it is no longer needed.

PR #2194, opened 2025-10-12
fix: convert hyphens to underscores when converting plugin names to generate valid environment variables

mgrechukh:
# Summary

In pre-0.16.0 shell implementation, tool name consisting of multiple words, separated by dash, were sanitized as environment variable. I.e. for "teleport-community" we have "ASDF_TELEPORT_COMMUNITY_VERSION"

As now toolName passed to variable name unmodified which breaks certain usage scenarios. Although variable containing dash "ASDF_TELEPORT-COMMUNITY_VERSION" sometimes *can* be set in surrounding code, it is clumsy.

This patch restores the old behaviour.


Stratus3D (commented):
Sorry for the extremely late review here @mgrechukh ! This is a good fix, but we need to have tests before I can merge. I'm going to try to write tests for this myself.

Stratus3D:
Thanks for the PR @mgrechukh ! The changes have been merged.

Issue #450, opened 2019-01-31
Windows support? Docs & tests for WSL2


jayvdb:
It would be great to have core support for windows, assuming it has git bash or similar, with Appveyor CI keeping it from breaking.

abhishalya:
I would like to work on this issue :)

jthegedus:
I know this will not help with CI tools, but just so it is documented in this thread for those who are looking for personal Windows use. I've used asdf-vm inside Windows 10 via WSL (Ubuntu) many times of the past two years. Following the Ubuntu Bash setup is compatible with using WSL.

earonesty:
no idea why this is closed.   installation on windows is definitely not supported.   wsl2 is not helpful at all for unit testing or ci/cd of tools that are designed to run on windows clients and servers.   be cool if it was some day.  is there a ticket to track that?

rainecheck:
> no idea why this is closed. installation on windows is definitely not supported. wsl2 is not helpful at all for unit testing or ci/cd of tools that are designed to run on windows clients and servers. be cool if it was some day. is there a ticket to track that?

Agreed. No references to Windows, Powershell, Git-Bash etc in [Getting Started](https://asdf-vm.com/guide/getting-started.html#_1-install-dependencies). Installation instructions list Linux and macOS specifically, but not Windows. It might be that support for Windows is there, but not documented; perhaps a separate "Add installation instructions for Windows" issue needs to be created, if this will not be reopened.

jthegedus:
Hi all,

When people discuss Windows support, I expect they mean "native" Windows support (see #1246), which typically does not refer to installing an entire Shell/Linux ecosystem/toolchain. This is why I always discuss Windows support in the context of WSL, which is the official Windows model to support Linux tooling, which I consider `asdf` to be. I have no idea how much work would be involved in supporting an environment like Powershell. Most other tools in this space have a specific separate repo for their Windows implementation as it differs so much from the Shell implementations.

We can do better with the documentation. Specifically a Windows section pointing out that it is possible to run `asdf` under specific conditions, like WSL, on a Windows machine. How helpful it is calling WSL2 "Windows support" I think is up for debate. Nonetheless, documentation is an area of improvement.

[...]

PR #2195, opened 2025-10-16
fix: use os.PathListSeparator

dolmen:
# Summary

Use [`os.PathListSeparator`](https://pkg.go.dev/os#PathListSeparator) instead of `":"`. This removes one blocker for a Windows port (#450) as the path separator on Windows is `";"`.

Also use [`path/filepath.SplitList`](https://pkg.go.dev/path/filepath#SplitList) instead of [`strings.Split`](https://pkg.go.dev/strings#Split) to correctly handle the (very unexpected) case of an empty `PATH` in `internal/shims.RemoveFromPath`.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -545,7 +545,7 @@ func envCommand(logger *log.Logger, shimmedCommand string, args []string) error 
 }
 
 func setPath(paths []string) string {
-	return strings.Join(paths, ":") + ":" + os.Getenv("PATH")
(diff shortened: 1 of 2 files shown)


carlos2martinize (dismissed):
The plug in gotta fix the pay route still.

Stratus3D (approved):
Thanks for the PR @dolmen ! Looks good to me.

carlos2martinize (changes requested):
I would like to refresh my memory for the next time I review the conclusions or options

Issue #2174, opened 2025-09-04
update documentation


myzxzrus:
### How can we improve the documentation?

The documentation on the site needs to be updated. According to the documentation for version 18.0, it is impossible to install asdf. It needs to be made clearer, or the steps need to be clarified. One option is to record a video with the installation steps on Linux.

johanla0:
How I managed to install it finally on Ubuntu (WSL2 setup):
1. Install go via snap (needs to be >1.23)
2. `go install github.com/asdf-vm/asdf/cmd/asdf@v0.18.0`
3. `mkdir ~/.asdf`
4. `cp /home/user/go/bin/asdf ~/.asdf`
5. Add to ~/.bashrc
```
export ASDF_DATA_DIR="$HOME/.asdf"
export PATH="$ASDF_DATA_DIR:$PATH"
```
---
```
go version go1.25.4 linux/amd64
asdf version 0.18.0 (revision unknown)
```

myzxzrus:
<img width="1864" height="923" alt="Image" src="https://github.com/user-attachments/assets/e3165469-ac64-47bf-b04d-97a07f57eb09" />

Great. It should work! On the website https://asdf-vm.com/guide/getting-started.html there are no steps 3 and 4. Please add this description to the website.  Thank you.

PR #2217, opened 2025-12-04
docs: Update getting started for go install

johanla0:
# Summary

Fixes: Adds correct info of how to get started when going with `go install`

Covers #2174


Stratus3D:
Hi @johanla0 , thanks for the PR! Those details you added are already covered in the step 2 section of the Getting Started page so I'm going to close this PR. Configuring asdf is specific to the shell you are using, so the instructions are listed for each supported shell. Open to any other improvements for the install documentation!

Issue #2002, opened 2025-03-02
doc: please document "asdf global" replacement - how do i access the installed tools?


alexanderankin:
### How can we improve the documentation?

I used to use global a lot. Now i suppose I must litter my aliases file with things like this?

```bash
alias helm=~/.asdf/installs/helm/3.17.1/bin/helm
```

please help!

Stratus3D:
Hi @alexanderankin can you please provide more context here?

No, you should not need to define aliases to use asdf-installed tools like helm. If you've configured asdf properly shims should already be on your `$PATH` and your shell should pick them up automatically. If that's not happening something is not configured right.

Try running this to see what is on your path for helm: `type -a helm`. An asdf shim path should be listed first.

alexanderankin:
```console
$ type -a helm
-bash: type: helm: not found
$ type asdf
asdf is /opt/homebrew/bin/asdf
```

joaquinipar:
It's way more confusing than before. Why changing something that just works? 🤷‍♂️

Stratus3D:
@alexanderankin so the shims directory is not on your path. Check step 2 instructions for your shell and make sure you've got code in your shells RC file to add the shims directory to your path. Once you get the shims directory on your path it should work.

Beyarz:
Try [this](https://github.com/asdf-vm/asdf/issues/1968#issuecomment-2665344943)

Beyarz:
Sounds like you're missing [this](https://github.com/asdf-vm/asdf/issues/1968#issuecomment-2665344943)

alexanderankin:
very tricky migration and i did not understand it but it seems to have finally worked after changing this, building from source and putting asdf in my path elsewhere:

https://github.com/alexanderankin/dotfiles/commit/812c362f9bd82298459282b8692bbf66df26456e

```diff
commit 812c362f9bd82298459282b8692bbf66df26456e

Date:   Sat Mar 8 11:15:30 2025 -0500

    asdf

diff --git a/.bashrc b/.bashrc
index d1cf7d3..e94dbc7 100644
--- a/.bashrc
+++ b/.bashrc
@@ -176,8 +176,18 @@ test -f /usr/bin/terraform && complete -C /usr/bin/terraform terraform
 command -v rbenv > /dev/null && eval "$(rbenv init -)"
 
 
-[[ -f ~/.asdf/asdf.sh ]] && . ~/.asdf/asdf.sh
[...]

PR #2236, opened 2026-01-21
docs: update note on 'asdf global' usage

Kartstig:
# Summary
Clarify the use of 'asdf global' and its replacement.

- Relates to #https://github.com/asdf-vm/asdf/issues/2002

## Other Information

Let me know if you would like to modify the comment/style. I was thinking to say:

> has been superceded by `asdf local` starting in :version:

but I couldn't easily trace it based on the Changelog.md. It looks like there's mention of the command up to version `0.8.1`:

Screenshot from 2026-01-21 16-53-01

### DIFF
--- a/docs/guide/getting-started-legacy.md
+++ b/docs/guide/getting-started-legacy.md
@@ -434,6 +434,8 @@ nodejs 16.5.0
 
 Some OSs already have tools installed that are managed by the system and not `asdf`, `python` is a common example. You need to tell `asdf` to pass the management back to the system. The [Versions reference section](/manage/versions.md) will guide you.
 
+Note: `asdf global` has been superceded by `asdf local` in later version
+
 ### Local
 
 Local versions are defined in the `$PWD/.tool-versions` file (your current working directory). Usually, this will be the Git repository for a project. When in your desired directory execute:


Stratus3D:
Thanks for your PR @Kartstig , but it's not correct. `asdf global` was replaced by `asdf set` with the `--home/-u` flag.

Stratus3D:
Looking at this again:

> but I couldn't easily trace it based on the Changelog.md

The changelog entry for 0.16.0 links to https://asdf-vm.com/guide/upgrading-to-v0-16.html#breaking-changes which details these changes. I do not think adding a single line to the legacy getting started guide is the right thing to do here. 

I'm not sure anything is needed as nobody should really be reading the legacy getting started guide page. Perhaps a banner at the top of the legacy getting started page linking to the upgrade guide (https://asdf-vm.com/guide/upgrading-to-v0-16.html) would be good?

Kartstig:
@Stratus3D It looks like it is now covered in https://asdf-vm.com/guide/upgrading-to-v0-16.html#asdf-global-and-asdf-local-commands-have-been-replaced-with-asdf-set.

I think it looks like I posted this issue prior to your 0.16.0 release, so maybe there wasn't anything at the time: https://github.com/asdf-vm/asdf/commit/c4e55259bb5df08aaf3486dc1cf80457498f551e

I will close this. Seems resolved

Issue #2184, opened 2025-09-17
bug: asdf install is not transactional


ajfabbri:
### Describe the Bug

Simplistic reproduction below... Related: #2036 

### Steps to Reproduce

```
asdf install python 4
```

This fails, as python 4 doesn't exist.

```
 % asdf list python
  4
% asdf install python 4
error installing version: version {version 4} of python is already installed
```

### Expected Behaviour

Installation should fail or succeed, not some partial combination of both.

### Actual Behaviour

```
asdf list python
```
 % asdf list python
  4
% asdf install python 4
error installing version: version {version 4} of python is already installed
```

### Environment

```shell
OS:
Darwin redacted.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6041 x86_64

SHELL:
zsh 5.9 (x86_64-apple-darwin24.0)

BASH VERSION:
3.2.57(1)-release

ASDF VERSION:
v0.16.0

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/Users/foo/.asdf
ASDF_CONFIG_FILE=/Users/foo/.asdfrc

ASDF INSTALLED PLUGINS:
gradle    https://github.com/rfrancis/asdf-gradle.git 3af56332a66b90f2c0743f2968b9f0425c479b94
java      https://github.com/halcyon/asdf-java.git    96f89465a471de99ae9c413d922c25f436537105
nodejs    https://github.com/asdf-vm/asdf-nodejs.git  fd2c7f94c4c71416047cacf2af6e4fa819b364b7
python    https://github.com/danhper/asdf-python.git  d4caa7d4832313b4c141ee0ceca4118747147e89
ruby      https://github.com/asdf-vm/asdf-ruby.git    26be0ebf43b47dba9bf244995700deeaa572b281
```

### asdf plugins affected (if relevant)

python

RishiAhuja:
I'm new to contributing here, but I've started using asdf recently and ran into this exact issue.

today, I was setting up a project that required a specific version of Node.js. I ran asdf install nodejs 22.x.x. partway through the download, my network connection dropped. the process hung, so I terminated it with ctrl+c (SIGINT).

[...]

Issue #2036, opened 2025-03-24
bug: install failed but it tell me already installed


ziyouchutuwenwu:
### Describe the Bug

run asdf install xxx, and it failed or manually interrupted, then run it again, it told me already installed.

### Steps to Reproduce

1. install and interrupt it.
```sh
$ asdf install python 3.10.0
python-build 3.10.0 /home/mmc/.asdf/installs/python/3.10.0
Downloading Python-3.10.0.tar.xz...
-> https://registry.npmmirror.com/-/binary/python/3.10.0/Python-3.10.0.tar.xz
^C
error: failed to download Python-3.10.0.tar.xz
```
2.  run it again
```sh
$ asdf install python 3.10.0
version 3.10.0 of python is already installed
```

### Expected Behaviour

continue to install

### Actual Behaviour

it told me  already installed

### Environment

```shell
$ asdf info                 
OS:
Linux manjaro 6.6.83-1-MANJARO #1 SMP PREEMPT_DYNAMIC Thu Mar 13 15:07:44 UTC 2025 x86_64 GNU/Linux

SHELL:
zsh 5.9 (x86_64-pc-linux-gnu)

BASH VERSION:
5.2.37(1)-release

ASDF VERSION:
v0.16.5

ASDF INTERNAL VARIABLES:
ASDF_DEFAULT_TOOL_VERSIONS_FILENAME=.tool-versions
ASDF_DATA_DIR=/home/mmc/.asdf
ASDF_CONFIG_FILE=/home/mmc/.asdfrc

ASDF INSTALLED PLUGINS:
nodejs    https://github.com/asdf-vm/asdf-nodejs.git 5515b8705fbf4d610ede04bb9a3fc4812e3a4168
python    https://github.com/danhper/asdf-python.git a3a01856098d6d2b9642e382f5b38e70275726d1
```

ziyouchutuwenwu:
hi @Stratus3D , i use version 0.16.6 to try, it seems not to be solved, here are the steps

```sh
mmc at p53 in ~ (dev_env) 
$ asdf list python            
 *3.12.9

mmc at p53 in ~ (dev_env) 
$ asdf install python 3.10.0  
python-build 3.10.0 /home/mmc/.asdf/installs/python/3.10.0
Downloading Python-3.10.0.tar.xz...
-> https://registry.npmmirror.com/-/binary/python/3.10.0/Python-3.10.0.tar.xz
^C
[...]

PR #2245, opened 2026-02-21
fix: use staged installation to prevent partial installs appearing in asdf list

RishiAhuja:
## Problem

When `asdf install <plugin> <version>` is interrupted mid-way — by Ctrl+C
(SIGINT), SIGKILL, a dropped network connection, or a power outage — the
partially installed version directory is left behind on disk. Because
`asdf list` detects installed versions simply by checking which
subdirectories exist under `$ASDF_DATA_DIR/installs/<plugin>/`, the
incomplete installation is reported as if it were valid:

```
$ asdf install nodejs 22.14.0
# ... interrupted halfway through ...

$ asdf list nodejs
nodejs
  22.14.0   ← listed, but broken
```

The user is then forced to manually run `asdf uninstall nodejs 22.14.0`
before they can retry the install. This is confusing because the version
was never successfully installed.

Reported and discussed in: https://github.com/asdf-vm/asdf/issues/2184

---

## Root Cause

`InstallOneVersion` in `internal/versions/versions.go` created the final
install directory (e.g. `installs/nodejs/22.14.0/`) with `os.MkdirAll`
**before** running the plugin's install callback. If the process was
killed at any point after that `MkdirAll` call, the directory remained on
disk. `IsInstalled` and `Installed` both use `os.Stat` / `os.ReadDir` on
the installs directory, so they reported the partial directory as a
complete installation.

---

## Solution: Staged Installation with Atomic Rename

This PR implements the **staged installation** pattern, which is the
standard approach used by package managers (npm, cargo, homebrew, apt)
to solve exactly this class of problem.

**Before:**
```
1. Create ~/.asdf/installs/nodejs/22.14.0/        ← final path, created immediately
2. Set ASDF_INSTALL_PATH=.../22.14.0
3. Run plugin install script (writes into 22.14.0/)
4. If interrupted → partial directory stays at final path, appears in asdf list
```

**After:**
```
1. Create ~/.asdf/installs/nodejs/22.14.0.incomplete/   ← staging path
2. Set ASDF_INSTALL_PATH=.../22.14.0.incomplete
3. Run plugin install script (writes into 22.14.0.incomplete/)
4. On SUCCESS → os.Rename(22.14.0.incomplete/, 22.14.0/)  ← atomic on POSIX
   On FAILURE  → os.RemoveAll(22.14.0.incomplete/)
   On SIGKILL  → 22.14.0.incomplete/ stays on disk, but is ignored by asdf list
                 and cleaned up automatically on the next install attempt
```

The final directory (`22.14.0/`) **never exists while an install is in
progress**. It only appears, atomically, after the install has fully
succeeded. `os.Rename()` within the same filesystem is an atomic POSIX
operation — it either completes fully or not at all.

This approach is strictly more robust than signal handling because it
also covers SIGKILL and power loss, which cannot be caught by any process.

---

### How does it look now?

```bash
# attempt to start a installation and simulate sigkill
(base) rishi@Rishis-MacBook-Pro asdf % ./asdf install nodejs 22.14.0 &
INSTALL_PID=$!
sleep 5
kill -9 $INSTALL_PID
[1] 85292
Trying to update node-build... ok
To follow progress, use 'tail -f /var/folders/5c/lzq441fs7bjdbbg_t7_0zzz40000gn/T/node-build.20260221171748.85326.log' or pass --verbose
Downloading node-v22.14.0-darwin-arm64.tar.gz...
-> https://nodejs.org/dist/v22.14.0/node-v22.14.0-darwin-arm64.tar.gz
Installing node-v22.14.0-darwin-arm64...
[1]  + killed     ./asdf install nodejs 22.14.0  

# list of installations
(base) rishi@Rishis-MacBook-Pro asdf % ls ~/.asdf/installs/nodejs/
22.14.0.incomplete      22.16.0                 24.9.0
(base) rishi@Rishis-MacBook-Pro asdf % ./asdf list nodejs
  22.16.0
  24.9.0
(base) rishi@Rishis-MacBook-Pro asdf % 
```

### `internal/installs/installs.go`

- Added `incompleteSuffix = ".incomplete"` constant.
- **`Installed()`** — now skips any subdirectory whose name ends with
  `.incomplete`, so interrupted installs are never surfaced by `asdf list`.
- **`StagingPath()`** — new function that returns
  `InstallPath(...) + ".incomplete"`. This is the directory the plugin
  install script writes into.
- **`CleanIncomplete()`** — new function that removes the staging
  directory if it exists. Safe to call when no such directory is present.
  Called at the start of every install attempt to clean up stale
  directories left behind by previous crashes.

### `internal/versions/versions.go`

Changes are entirely within `InstallOneVersion`:

- Computes `stagingDir` via `installs.StagingPath()` alongside the
  existing `installDir` and `downloadDir`.
- Calls `installs.CleanIncomplete()` before starting to wipe any stale
  staging directory from a prior interrupted install.
- Passes `stagingDir` (not `installDir`) as `ASDF_INSTALL_PATH` in the
  environment given to the plugin's download and install callbacks.
  Plugin scripts are unaffected — they simply write to `$ASDF_INSTALL_PATH`
  as before.
- Creates `stagingDir` (not `installDir`) with `os.MkdirAll`.
- On install callback failure: removes `stagingDir` (same cleanup
  behaviour as before, now correctly targeting the staging path).
- On install callback success: calls `os.Rename(stagingDir, installDir)`
  to atomically promote the staged directory to the final install path.
  If the rename itself fails (very rare edge case), cleans up the
  staging directory and returns an error.

### `internal/installtest/installtest.go`

Updated the test helper `InstallOneVersion` to mirror the same
staging-and-rename flow as the production code, so tests that rely on
this helper produce installations consistent with the new behaviour.

---

### `internal/installs/installs_test.go` (new test cases)

| Test | What it verifies |
|---|---|
| `TestIsInstalled/returns false for incomplete installation` | An `.incomplete` directory does not make `IsInstalled` return `true` |
| `TestInstalled_FiltersIncomplete/does not list incomplete installations` | `Installed()` excludes `.incomplete` dirs from its results |
| `TestStagingPath/returns install path with .incomplete suffix` | `StagingPath()` produces the correct string |
| `TestCleanIncomplete/removes incomplete staging directory` | `CleanIncomplete()` recursively deletes the staging directory |
| `TestCleanIncomplete/succeeds even when no incomplete directory exists` | `CleanIncomplete()` is safe to call when nothing exists |

### `internal/versions/versions_test.go` (new test cases)

| Test | What it verifies |
|---|---|
| `TestInstallOneVersion/staging directory does not exist after successful install` | After a successful install, `22.14.0.incomplete/` is gone and `22.14.0/` exists with correct contents |
| `TestInstallOneVersion/staging directory is cleaned up when install fails` | When the install script exits non-zero, both the staging dir and final dir are absent |
| `TestInstallOneVersion/cleans up stale incomplete directory from previous failed install` | A pre-existing `.incomplete` dir (simulating a prior crash) is cleaned up and the new install succeeds |

---

## Behaviour by Failure Mode

| Scenario | Result after this PR |
|---|---|
| Install script returns non-zero | Staging dir cleaned up. Final dir never created. |
| Ctrl+C (SIGINT) during install | Go runtime returns error from `RunCallback`. Staging dir cleaned up. Final dir never created. |
| SIGKILL / power outage during install | Staging dir left on disk. **Not visible in `asdf list`**. Cleaned up automatically on next install attempt. |
| Normal success | Staging dir atomically renamed to final path. |

---

## Testing Locally

```bash
# Run all affected unit tests
go test ./internal/installs/ ./internal/versions/ -v -count=1

# Run only the new tests
go test ./internal/installs/ -v -run "TestInstalled_FiltersIncomplete|TestStagingPath|TestCleanIncomplete|TestIsInstalled"
go test ./internal/versions/ -v -run "TestInstallOneVersion/staging|TestInstallOneVersion/cleans_up_stale"

# Run the full internal test suite
go test ./internal/... -count=1
```

---

## Compatibility

- **No breaking changes.** The `ASDF_INSTALL_PATH` variable passed to
  plugin scripts now points to the staging directory during installation
  instead of the final directory. The path format is otherwise identical
  and the variable name is unchanged. Plugins that write to
  `$ASDF_INSTALL_PATH` (which is all of them) continue to work without
  modification.
- Any `.incomplete` directories already sitting on disk from previous
  versions of asdf will be invisible to `asdf list` and will be
  automatically removed on the next `asdf install` attempt for that
  version.

### DIFF
--- a/internal/installs/installs.go
+++ b/internal/installs/installs.go
@@ -7,12 +7,15 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/asdf-vm/asdf/internal/config"
 	"github.com/asdf-vm/asdf/internal/data"
 	"github.com/asdf-vm/asdf/internal/plugins"
 	"github.com/asdf-vm/asdf/internal/toolversions"
 )
+
+const incompleteSuffix = ".incomplete"
 
 // Installed returns a slice of all installed versions for a given plugin
 func Installed(conf config.Config, plugin plugins.Plugin) (versions []string, err error) {
@@ -31,7 +34,13 @@ func Installed(conf config.Config, plugin plugins.Plugin) (versions []string, er
 			continue
 		}
 
-		versions = append(versions, toolversions.VersionStringFromFSFormat(file.Name()))
+		name := file.Name()
+		// Skip incomplete installations left behind by interrupted installs
+		if strings.HasSuffix(name, incompleteSuffix) {
+			continue
+		}
+
+		versions = append(versions, toolversions.VersionStringFromFSFormat(name))
 	}
 
 	return versions, err
@@ -53,6 +62,20 @@ func DownloadPath(conf config.Config, plugin plugins.Plugin, version toolversion
 	}
 
 	return filepath.Join(data.DownloadDirectory(conf.DataDir, plugin.Name), toolversions.FormatForFS(version))
+}
+
+// StagingPath returns the temporary staging path used during installation.
+// Installations are first performed in this directory and then atomically
+// renamed to the final InstallPath on success, so that interrupted installs
+// never appear as valid installed versions.
+func StagingPath(conf config.Config, plugin plugins.Plugin, version toolversions.Version) string {
+	return InstallPath(conf, plugin, version) + incompleteSuffix
+}
+
+// CleanIncomplete removes any stale staging directories for a given plugin and
+// version. It is safe to call even if no incomplete directory exists.
+func CleanIncomplete(conf config.Config, plugin plugins.Plugin, version toolversions.Version) error {
+	return os.RemoveAll(StagingPath(conf, plugin, version))
 }
 
 // IsInstalled checks if a specific version of a tool is installed
--- a/internal/installs/installs_test.go
+++ b/internal/installs/installs_test.go
@@ -77,6 +77,68 @@ func TestIsInstalled(t *testing.T) {
 		version := toolversions.Version{Type: "version", Value: "1.0.0"}
 		assert.True(t, IsInstalled(conf, plugin, version))
 	})
+	t.Run("returns false for incomplete installation", func(t *testing.T) {
+		// Create an .incomplete staging directory (simulating interrupted install)
+		version := toolversions.Version{Type: "version", Value: "2.0.0"}
+		stagingPath := InstallPath(conf, plugin, version) + ".incomplete"
+		err := os.MkdirAll(stagingPath, os.ModePerm)
+		assert.Nil(t, err)
+
+		assert.False(t, IsInstalled(conf, plugin, version))
+	})
+}
+
+func TestInstalled_FiltersIncomplete(t *testing.T) {
+	conf, plugin := generateConfig(t)
+
+	t.Run("does not list incomplete installations", func(t *testing.T) {
+		// Install a real version
+		mockInstall(t, conf, plugin, "1.0.0")
+
+		// Create an .incomplete staging directory
+		incompletePath := filepath.Join(conf.DataDir, "installs", testPluginName, "2.0.0.incomplete")
+		err := os.MkdirAll(incompletePath, os.ModePerm)
+		assert.Nil(t, err)
+
+		installedVersions, err := Installed(conf, plugin)
+		assert.Nil(t, err)
+		assert.Equal(t, []string{"1.0.0"}, installedVersions)
+	})
+}
+
+func TestStagingPath(t *testing.T) {
+	conf, plugin := generateConfig(t)
+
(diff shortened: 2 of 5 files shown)


Stratus3D (commented):
Hi @RishiAhuja,

Thank you for the PR, however this is not the approach we want to take. There has been past discussion of this on this issue: https://github.com/asdf-vm/asdf/issues/2036

Specifically, two approaches were suggested here: https://github.com/asdf-vm/asdf/issues/2036#issuecomment-2751747127 And I suggested option 1 (trap the signal) over option 2 (temp/staged dir, as you have here): https://github.com/asdf-vm/asdf/issues/2036#issuecomment-2752390089. I still think that is the right approach. We should be trapping signals and we currently do not.

Also, please do not use AI to write pull request descriptions or issue comments. I do not have time to read the excessively verbose output.

PR #2274, opened 2026-05-04
docs: clarify what `<command>` means in `asdf env` (#459)

### LINKED ISSUE #459: Clarify usage of asdf env command in help output
### Steps to reproduce

Run `asdf`. In help output a line is printed for the `asdf env` command:

```
asdf env <command> [executable]      Prints or runs an executable under a command environment
```
I would assume `<command>` is an arbitrary command. But from what I can tell needs to be a plugin name or a shim name? I'm also not sure how this differs from executing the shim directly.

FYI @vic

tmchow:
# Summary

Replaces the `<!-- TODO: expand on this with example -->` placeholder under `## Env` in `docs/manage/core.md` with a one-paragraph clarification plus two worked examples. The clarification spells out what the original issue reporter (#459) wasn't sure about: `<command>` is a shimmed binary name (`node`, `python`, `ruby`, etc.), not an arbitrary shell command.

The help text in `help.txt` keeps the same `<command>` placeholder for consistency with the other shim-name commands in that file (`asdf which <command>`, `asdf exec <command>`, `asdf shim-versions <command>`); changing the placeholder in just one of them would create a worse inconsistency. The doc clarification is the right place to spell out the convention.

The two examples are deliberately minimal:

- `asdf env node` prints the environment that `node` runs in.
- `asdf env node which npm` runs `which` inside that same environment so `npm` resolves to the version-specific shim.

Both work against current `asdf` (verified locally that the relevant code path in `internal/cli/cli.go` is unchanged in shape).

Fixes #459

## Other Information

The `## Exec` section just above `## Env` still has its own `TODO: expand on this with example` placeholder. I didn't touch it because that's a different command and the issue here only asks for `env` clarification. Happy to send a follow-up PR for `exec` if useful.

### DIFF
--- a/docs/manage/core.md
+++ b/docs/manage/core.md
@@ -22,7 +22,15 @@ Executes the command shim for the current version.
 asdf env <command> [util]
 ```
 
-<!-- TODO: expand on this with example -->
+`<command>` is the name of a shimmed binary (for example `node`, `python`, `ruby`), not an arbitrary shell command. `asdf env` resolves the plugin and version that would handle that shim and runs `[util]` (defaults to `env`) inside the same environment, so you can inspect or invoke other tools with the plugin's `PATH` and `ASDF_*` variables already set.
+
+```shell
+# Print the environment that `node` runs in.
+$ asdf env node
+
+# Run `which` inside that same environment to confirm where `npm` resolves.
+$ asdf env node which npm
+```
 
 ## Info
 


Stratus3D (approved):
Thanks @tmchow and AI! Yes, we never really settled on how this command is suppose to work, but the changes here do correctly document how it works today. We may revisit this in the future.

PR #2279, opened 2026-05-14
ci: build linux/ppc64le release binaries

runlevel5:
I'd like to express my gratitude to authors for creating this super useful toolings. I've been using asdf (pre-0.16) in enterprise environments for a quite awhile and was setting up an internal build system to compile asdf (post-0.16) from sources. It works just fine but inconvenient as the maintenance poses a bit of work. I would like to propose that the linux ppc64le binary could be shipped as part of the release process as it would greatly benefit many PowerPC users out there.


Stratus3D (approved):
Hi @runlevel5 , thanks for the PR! Yes, I'd be happy to have our release process generate your ppc64le binaries. I've got a couple questions:

* Instead of `include` should we just add `ppc64le` to the matrix and exclude it on MacOS?
* Do you think we should run CI tests on `ppc64le` as well?

runlevel5:
> Do you think we should run CI tests on ppc64le as well?

Ideally you should. Please apply for ppc64le GitHub runner with https://github.com/IBM/actionspz

> Instead of include should we just add ppc64le to the matrix and exclude it on MacOS?

Sure I will update the PR

runlevel5:
@Stratus3D the PR is ready for review

Stratus3D:
Hi @runlevel5 , I'm sorry, but I've changed my mind on this one. I don't think it makes sense to go through the process of requesting a ppc64le GitHub runner. This is a fairly unique unusual architecture so I think it makes sense for you to continue to build asdf for it yourself. Sorry for the change of plans!

runlevel5:
@Stratus3D thanks for the response. Alternatively cross-compilation is also a great solution, in fact it has always been the more popular option in many projects. If you do not object, I could re-submit another Pull Request

Stratus3D:
My concern with publishing ppc64le release binaries is that they will be unsupported. Users might consider a published binary to be an indicator of official support. I don't have currently have any way of testing ppc64le binaries or debugging them on a ppc64le system.

runlevel5:
> My concern with publishing ppc64le release binaries is that they will be unsupported. Users might consider a published binary to be an indicator of official support. I don't have currently have any way of testing ppc64le binaries or debugging them on a ppc64le system.

It is a fair concern though I do not think it would be much of an issue because majority of ppc64le users are very technical savvy so high chance they would help contribute in fixing the issue. Lastly you could use QEMU to run ppc64le emulator, for a small size app like asdf, emulator can handle it easily.

